PR #13, opened 2018-12-10
Accessibility bugs

YuliKl:
Fixing two accessibility bugs pertaining to 200% text scaling and radio button invocation with Narrator.

## Description

- Added text wrapping property.
- Updated NavigationView options' RadioButtons to use more specific Checked events instead of more generic Click events.

## Motivation and Context

- Bug 19597343: Text Scaling Desktop: XAML Controls Gallery : Information present for RichEditBox is clipped in 200% Text Scaling.
- Bug 19440518: Narrator Desktop : XAML Controls Gallery : Unable to change the position of navigation pane using caps+enter 

## How Has This Been Tested?

manual verification

## Types of changes


stmoy:
As a bit more context: The Click handlers need to be replaced with Checked handlers because those are expected to be triggered by automation peers too. 
The Checked/Unchecked events are the abstract ones that are universally called no matter which input is used.
For RadioButtons in general, we need to use the Checked events over the Click events.

PR #22, opened 2019-01-04
Missing UIA names

YuliKl:
Last batch of changes to add in missing automation names.

## Description

Added UIA names and updated sample code when needed. Added a workaround for both description and state of ToggleSwitch to the Settings page. And accidentally fixed the formatting in NavigationViewPage using Ctrl K, Ctrl D.

## Motivation and Context

Internal bugs

## How Has This Been Tested?

Manually verified

## Types of changes

### DIFF
--- a/XamlControlsGallery/SettingsPage.xaml
+++ b/XamlControlsGallery/SettingsPage.xaml
@@ -51,11 +51,7 @@
                     FontWeight="Normal"
                     Text="Sound" />
                 <StackPanel Margin="0,10,0,0">
-                    <ToggleSwitch x:Name="soundToggle" Header="Sound" Toggled="soundToggle_Toggled" OffContent="Off" OnContent="On">
-                        <ToggleSwitch.HeaderTemplate>
-                            <DataTemplate><!-- empty template to support correct visuals and Narrator experience --></DataTemplate>
-                        </ToggleSwitch.HeaderTemplate>
-                    </ToggleSwitch>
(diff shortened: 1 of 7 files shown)


XamlControlsGallery/SettingsPage.xaml
```diff
@@ -51,7 +51,11 @@
                     FontWeight="Normal"
                     Text="Sound" />
                 <StackPanel Margin="0,10,0,0">
-                    <ToggleSwitch x:Name="soundToggle" Toggled="soundToggle_Toggled" OffContent="Off" OnContent="On"/>
+                    <ToggleSwitch x:Name="soundToggle" Header="Sound" Toggled="soundToggle_Toggled" OffContent="Off" OnContent="On">
+                        <ToggleSwitch.HeaderTemplate>
+                            <DataTemplate><!-- empty template to support correct visuals and Narrator experience --></DataTemplate>
```
stmoy: This is strange. Since the ToggleSwitch has a Header set, shouldn't it do the "right thing" by default? I'm concerned that this might be an indication that ToggleSwitch controls don't have the right Narrator experience by default.

YuliKl: It didn't have a header set. The "Sound" text is a distinct TextBlock. I added a header, which got me the correct Narrator behavior but resulted in two "sound" headers. So I'm heading the ToggleSwitch's built-in header to keep the visuals consistent with the current, larger text size, implementation.

PR #27, opened 2019-01-08
Update to CalandarView language UI 

YuliKl:
Replaced free-form TextBox language entry with predefined language list available inside a (non-editable) ComboBox.

## Description

The predefined list of languages is copied from https://docs.microsoft.com/en-us/windows/uwp/publish/supported-languages. I chose to put English first with others in alphabetical order.

## Motivation and Context

Fixed internal bug 19973652 by removing the need for error MessageDialog.

## How Has This Been Tested?

Manually verified

## Types of changes

### DIFF
--- a/XamlControlsGallery/Common/LanguageList.cs
+++ b/XamlControlsGallery/Common/LanguageList.cs
@@ -1,0 +1,124 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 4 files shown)


YuliKl:
@kikisaints, the two or three letter language code is sufficient. Something like "fr-CA" provides both the language and the locale: French, in Canada. But it turns out that CalendarView doesn't actually care about the second half of that compound combination.

kikisaints:
@YuliKl Got it! In that case it all looks good to me 💯

PR #30, opened 2019-01-09
Swipe control text fixed at 225% scale text

kikisaints:
Fixed the swipe control to actually display all info it can when the system text is scaled to 225%.

## Description
Just upped the width number. Didn't completely remove it because when I do, the swipe control expands the entire width of the window which looks super awkward when attempting to swipe it. It will show all the text necessary on the item at max 225% text scaling though with this fix.

## Motivation and Context
Some text was being cut off when the text scaling was increased through the system.

## How Has This Been Tested?
Tested on min and max window widths and at min and max system text scaling options.

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/SwipePage.xaml
+++ b/XamlControlsGallery/ControlPages/SwipePage.xaml
@@ -36,7 +36,7 @@
                                            Invoked="Flag_ItemInvoked"/>
                             </SwipeItems>
                         </Border.Resources>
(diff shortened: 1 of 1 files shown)


stmoy:
Is it possible to go higher than 225%? If so, I'm worried that swapping one hard-coded width for another is not a robust solution.

YuliKl:
225% is currently as high as that accessibility slider allows. The MAS requirement is to support 200%.

PR #32, opened 2019-01-14
Custom focus

YuliKl:
Removing custom focus logic, which I believe to be intended to improve gamepad navigation.

## Description

The primary ScrollViewer was inadvertently trapping focus, prevent users from being able to Tab pas the first element. This also resulted in Tab and Shift+Tab showing different behavior. Removing three private methods (Loaded, GotFocus, KeyDown) responsible for unpredictable keyboard behavior.

Also updated if statement in another private KeyDown method to use MUXC instead of WUXC NavigationViewItem, and added null check.

## Motivation and Context

internal bug 19414657

## How Has This Been Tested?

Manually verified with keyboard and gamepad

## Types of changes

### DIFF
--- a/XamlControlsGallery/ItemsPageBase.cs
+++ b/XamlControlsGallery/ItemsPageBase.cs
@@ -71,7 +71,7 @@ namespace AppUIBasics
             if (e.Key == VirtualKey.Up)
             {
                 var nextElement = FocusManager.FindNextElement(FocusNavigationDirection.Up);
-                if (nextElement.GetType() == typeof(NavigationViewItem))
+                if (nextElement != null && nextElement.GetType() == typeof(Microsoft.UI.Xaml.Controls.NavigationViewItem))
                 {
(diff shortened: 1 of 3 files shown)


XamlControlsGallery/ItemsPageBase.cs
```diff
@@ -71,7 +71,7 @@ protected void OnItemGridViewKeyDown(object sender, KeyRoutedEventArgs e)
             if (e.Key == VirtualKey.Up)
             {
                 var nextElement = FocusManager.FindNextElement(FocusNavigationDirection.Up);
-                if (nextElement.GetType() == typeof(NavigationViewItem))
+                if (nextElement != null && nextElement.GetType() == typeof(Microsoft.UI.Xaml.Controls.NavigationViewItem))
```
Austin-Lamb: Nitpick: could just use "nextElement?.GetType()" if you like the shorter syntax.

YuliKl: Thank you, I didn't know about this syntax.

PR #47, opened 2019-01-31
CommandBarFlyout's show modes

YuliKl:
Added check for whether pointer is over the button target to determine correct mode for CommandBarFlyout.

## Description

When invoked with keyboard, users must be able to navigate with keyboard within the CommandBarFlyout. When invoked with a pointer, users should see the more compact, transient CommandBarFlyout. Unfortunately, the Click and ContextRequested events don't have arguments detailing which input device was used to fire them. So I'm using pointer position (Buton's IsPointerOver property) as a proxy. Unfortunately this approach is not full-proof, nothing prevents users from moving their cursor over the button and pressing the Space key. But this seems better than always showing the expanded Standard flyout mode.

## Motivation and Context

Internal bug 20144165

## How Has This Been Tested?

Manually verified

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/CommandBarFlyoutPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/CommandBarFlyoutPage.xaml.cs
@@ -36,8 +36,9 @@ namespace AppUIBasics.ControlPages
         }
 
         private void MyImageButton_ContextRequested(Windows.UI.Xaml.UIElement sender, ContextRequestedEventArgs args)
-        {            
-            ShowMenu((sender as Button).IsPointerOver);
+        {   
+            // always show a context menue in standard mode
+            ShowMenu(false);
         }
 
         private void MyImageButton_Click(object sender, Windows.UI.Xaml.RoutedEventArgs e)
--- a/XamlControlsGallery/ControlPages/CommandBarFlyoutPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/CommandBarFlyoutPage.xaml.cs
@@ -37,7 +37,7 @@ namespace AppUIBasics.ControlPages
 
         private void MyImageButton_ContextRequested(Windows.UI.Xaml.UIElement sender, ContextRequestedEventArgs args)
         {   
-            // always show a context menue in standard mode
+            // always show a context menu in standard mode
             ShowMenu(false);
         }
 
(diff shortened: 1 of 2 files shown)


XamlControlsGallery/ControlPages/CommandBarFlyoutPage.xaml.cs
```diff
@@ -36,13 +36,13 @@ private void ShowMenu()
         }
 
         private void MyImageButton_ContextRequested(Windows.UI.Xaml.UIElement sender, ContextRequestedEventArgs args)
-        {
-            ShowMenu();
+        {            
+            ShowMenu((sender as Button).IsPointerOver);
```
llongley: ContextRequested is raised when we right-click to explicitly request a context menu, which should show in standard mode - transient mode is only when the flyout is being shown incidentally in a situation where the user may not want it.

XamlControlsGallery/ControlPages/CommandBarFlyoutPage.xaml.cs
```diff
@@ -36,13 +36,14 @@ private void ShowMenu()
         }
 
         private void MyImageButton_ContextRequested(Windows.UI.Xaml.UIElement sender, ContextRequestedEventArgs args)
-        {
-            ShowMenu();
+        {   
+            // always show a context menue in standard mode
```
stmoy:
```suggestion
            // always show a context menu in standard mode
```

XamlControlsGallery/ControlPagesSampleCode/CommandBarFlyout/CommandBarFlyoutSample1_cs.txt
```diff
@@ -1,16 +1,17 @@
...
     FlyoutShowOptions myOption = new FlyoutShowOptions();
-    myOption.ShowMode = FlyoutShowMode.Transient;
+    myOption.ShowMode = isTransient ? FlyoutShowMode.Transient : FlyoutShowMode.Standard;
     CommandBarFlyout1.ShowAt(Image1, myOption);
 }
 
 private void MyImageButton_ContextRequested(Windows.UI.Xaml.UIElement sender, ContextRequestedEventArgs args)
 {
-    ShowMenu();
+    // always show a context menue in standard mode
```
stmoy:
```suggestion
    // always show a context menu in standard mode
```

PR #57, opened 2019-02-22
Add sample for AnimatedVisualPlayer

sohchatt:
## Description
Added a new sample for AnimatedVisualPlayer

## Motivation and Context
self-explanatory

## How Has This Been Tested?
built app locally on 17763 + Insider flights

## Types of changes


XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
```
stmoy: Please remove unnecessary "USING"s.

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -0,0 +1,83 @@
...
+using Windows.Foundation.Collections;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Controls.Primitives;
+using Windows.UI.Xaml.Data;
+using Windows.UI.Xaml.Input;
+using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Navigation;
+
+// The Blank Page item template is documented at https://go.microsoft.com/fwlink/?LinkId=234238
```
stmoy: Please remove boilerplate code and comments.

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
```diff
@@ -0,0 +1,107 @@
...
+                    <TextBlock Width="600" TextWrapping="WrapWholeWords">
+                        This AnimatedVisualPlayer consumes a Lottie animation which was created in 
+                        Adobe AfterEffects and translated into Windows.UI.Composition objects using 
+                        <Hyperlink NavigateUri="https://aka.ms/lottie">
+                           Lottie-Windows
+                        </Hyperlink>.
+                        Since the 
+                        <Hyperlink NavigateUri="https://docs.microsoft.com/uwp/api/windows.ui.composition.compositionshape">
+                            CompositionShapes
+                        </Hyperlink> being used in the animation are supported on Windows 10 version 17763 and above,
```
stmoy: Do we need to do anything specific to make this work down-level to the creator's update (RS2)? Or is this handled by the control? In other words, I'm wondering if we need to do any version adaptive or conditional xaml here.

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
```diff
@@ -0,0 +1,107 @@
...
+                    </TextBlock>
+                    <Border Width="400" Height="400" 
+                            BorderBrush="{ThemeResource ToolTipBorderThemeBrush}"
+                            BorderThickness="5"
+                            Background="LightGray"
+                            Margin="0, 20, 0, 20"
+                            >
+
+                        <!--  AnimatedVisualPlayer  -->
+                        <muxcontrols:AnimatedVisualPlayer x:Name="Player"
```
stmoy: Please include this chunk of markup as the example xaml markup.

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
```diff
@@ -0,0 +1,107 @@
...
+                        </Hyperlink>.
+                        Since the 
+                        <Hyperlink NavigateUri="https://docs.microsoft.com/uwp/api/windows.ui.composition.compositionshape">
+                            CompositionShapes
+                        </Hyperlink> being used in the animation are supported on Windows 10 version 17763 and above, 
+                        the AnimatedVisualPlayer instatiates an Image as FallbackContent when its Source is unavailable.
+                    </TextBlock>
+                    <Border Width="400" Height="400" 
+                            BorderBrush="{ThemeResource ToolTipBorderThemeBrush}"
+                            BorderThickness="5"
```
stmoy: Please remove the BorderBrush/Thickness. It's not necessary for this sample.

XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -1472,6 +1472,29 @@
       "ImagePath": "",
       "Description": "",
       "Items": [
+        {
+          "UniqueId": "AnimatedVisualPlayer",
+          "Title": "AnimatedVisualPlayer",
+          "Subtitle": "The AnimatedVisualPlayer allows playback of animated Windows.UI.Composition.Visual trees.",
```
stmoy: Maybe this should be "You can use AnimatedVisualPlayer to playback animations" or something. The way that it is currently worded is overly technical and doesn't describe the user value.

XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -1472,6 +1472,29 @@
...
+          "ImagePath": "Assets/Animations.png",
+          "Description": "The AnimatedVisualPlayer allows playback of animated Windows.UI.Composition.Visual trees.",
+          "Content": "<p>Look at the <i>AnimatedVisualPlayerPage.xaml</i> and <i>AnimatedVisualPlayerPage.xaml.cs</i> files in Visual Studio to see the full code for this page.</p>",
+          "IsNew": true,
```
stmoy: Is this control still in preview in WinUI, or has it formally shipped? If it's still in preview, we need to mark it differently.

XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -1472,6 +1472,29 @@
...
+          "Docs": [
+            {
+              "Title": "AnimatedVisualPlayer API",
+              "Uri": "https://www.docs.microsoft.com/uwp/api/microsoft.ui.xaml.controls.animatedvisualplayer"
+            },
+            {
+              "Title": "Sample application with walkthrough of scenarios",
```
stmoy:
```suggestion
              "Title": "Full sample",
```

XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -1472,6 +1472,29 @@
...
+              "Uri": "ms-windows-store://pdp/?productid=9N3J5TG8FF7F"
+            },
+            {
+              "Title": "Documentation and Tutorials",
```
stmoy:
```suggestion
              "Title": "Guidelines and Tutorials",
```

stmoy: This link appears under the "Documentation" section on the page so we should be crisp about what kind of documentation it is. I suggested "Guidelines" but I'm open to other ideas. I just don't think "Documentation" under "Documentation" looks right.

XamlControlsGallery/DataModel/ControlInfoData.json
sohchatt: It's in preview today but will be released whenever @jesbis has the next release of WinUI planned. All the code for AVP has been migrated to the release branch.

XamlControlsGallery/DataModel/ControlInfoData.json
sohchatt: except it can't playback all animations, only those created with Windows.UI.Composition.

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
sohchatt: Sort of is, because the animation has a Uniform Stretch. I think the border clarifies that everything within it is the animated visual tree. Thoughts?

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
sohchatt: Unclear what's needed here -- @stmoy can you clarify or cite the relevant pattern I need to follow?

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
sohchatt: The comment in the XAML  below addresses this issue, partially. The AnimatedVisualPlayer relies on AnimationController APIs which are present only on RS4+. However, the AVP has a FallbackContent property that will instantiate its DataTemplate whenever its Source fails. In this example, the LottieVisualSource being used has a dependency on RS5+. To summarize: 

1.  when downlevel is invoked is up to the Source author and which version the IAnimatedVisualSource is able to target
2. the player always uses FallbackContent when it can't use its Source

I didn't want to have to outline this in the code sample. For now, the only Source type being ingested by the Player is provided by Lottie-Windows. As different Sources are create over time, they'll need to be documented accordingly.

XamlControlsGallery/DataModel/ControlInfoData.json
sohchatt: Agreed. Any chance you can remove the "Related Controls" list to the right as well?

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml.cs
sohchatt: I used the Removed Unused Usings feature in VS2017 and these don't get removed.

sohchatt: Removed

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml.cs
sohchatt: Removed

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
sohchatt: seems like you can't have it inside a stackpanel. Instead, I've added the local:ControlExample.Xaml element to make things clear

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
stmoy: Gotcha, thanks for the explanation. To be a bit more specific - we're actively updating the gallery to run down to RS2. How will this sample behave if running on an RS2 machine? Will the app build if the minimum app target is RS2?

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
stmoy: A uniform stretch should be achievable without a border, but I agree that the border helps to clarify the boundary. Given that context, I don't have a problem with the border (and we can always change it later if we need to 😉).

XamlControlsGallery/DataModel/ControlInfoData.json
stmoy: True, but the meta-feedback was that this text was overly technical. For example, the Implicit Transitions page says "Use implicit transitions to automatically animate changes to properties" (even though it only supports a few of the UIElement properties) and the ListView page says "The ListView lets you show a collection of items in a list that scrolls vertically". 

What about "AnimatedVisualPlayer allows you to playback Lottie animations" or something?

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
sohchatt: It will render a static image instead of an animation on RS2/3/4

XamlControlsGallery/DataModel/ControlInfoData.json
sohchatt: Makes sense. I'll change this to be something without tech specifics.

improbabilityphysics: AnimatedVisualPlayer allows you to playback animations such as Lottie?

XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
stmoy: Cool, thanks for the clarification.

XamlControlsGallery/DataModel/ControlInfoData.json
stmoy: Thanks for the input @improbabilityphysics 😁 - @sohchatt updated the text to mention "motion graphics" specifically so I think we're good now.

PR #59, opened 2019-02-22
Added Acrylic Luminosity sample

kikisaints:
New 19H1 feature is the Acrylic luminosity - this PR is to add a small demo of that property to the Acrylic page.

## Description
Added a sections after the last sample on the Acrylic page with a tint opacity and luminosity opacity slider that updates the acrylic surface live to show you how the new luminosity property works in conjunction with the tint opacity property.

## Motivation and Context
The main reason why the luminosity opacity property was introduced was to improve shadow performance by removing the need for the shadow to be "cut out" with the shape of the elevated surface. Luminosity when raise and in balance with the tint opacity can almost completely mask the color black from shapes that are behind the Acrylic surface.

## How Has This Been Tested?
- Multiple screens
- Min and max window sizes
- Min and max tint and luminosity values
- Light + dark theme
- High contrast

## Types of changes

### DIFF
--- a/XamlControlsGallery/XamlControlsGallery.csproj
+++ b/XamlControlsGallery/XamlControlsGallery.csproj
@@ -10,8 +10,8 @@
     <AssemblyName>AppUIBasics</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.17763.0</TargetPlatformVersion>
-    <TargetPlatformMinVersion>10.0.17134.0</TargetPlatformMinVersion>
+    <TargetPlatformVersion>10.0.18843.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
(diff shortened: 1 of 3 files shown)


XamlControlsGallery/XamlControlsGallery.csproj
```diff
@@ -10,8 +10,8 @@
     <AssemblyName>AppUIBasics</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.17763.0</TargetPlatformVersion>
-    <TargetPlatformMinVersion>10.0.17134.0</TargetPlatformMinVersion>
+    <TargetPlatformVersion>10.0.18843.0</TargetPlatformVersion>
```
stmoy: Since this feature requires an insider version of the SDK (or newer), I think we'll need to hold off pulling this into master until the SDK locks.

jevansaks: The property is in WinUI, I'm curious what changes here required the retarget? Maybe it was a quirk thing? We can split the TargetPlatformVersion and the MaxTestedVersion if we need the app to be running against 19H1 quirks.

kikisaints: I think we're (Steven and I) are going to create a separate 19H1 branch for the gallery app and I'll make the PR to that branch. afaik this wasn't a quirk, just a change in the framework to add this luminosity property.

PR #62, opened 2019-02-26
Acrylic Luminosity Sample

kikisaints:
New 19H1 feature is the Acrylic luminosity - this PR is to add a small demo of that property to the Acrylic page.

## Description
Added a sections after the last sample on the Acrylic page with a tint opacity and luminosity opacity slider that updates the acrylic surface live to show you how the new luminosity property works in conjunction with the tint opacity property.

## Motivation and Context
The main reason why the luminosity opacity property was introduced was to improve shadow performance by removing the need for the shadow to be "cut out" with the shape of the elevated surface. Luminosity when raise and in balance with the tint opacity can almost completely mask the color black from shapes that are behind the Acrylic surface.

## How Has This Been Tested?
- Multiple screens
- Min and max window sizes
- Min and max tint and luminosity values
- Light + dark theme
- High contrast

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/AcrylicPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/AcrylicPage.xaml.cs
@@ -24,12 +24,17 @@ namespace AppUIBasics.ControlPages
         {
             ColorSelector.SelectedIndex = ColorSelectorInApp.SelectedIndex = 0;
             FallbackColorSelector.SelectedIndex = FallbackColorSelectorInApp.SelectedIndex = 0;
-            OpacitySlider.Value = OpacitySliderInApp.Value = 0.8;
+            OpacitySlider.Value = OpacitySliderInApp.Value = OpacitySliderLumin.Value = 0.8;
+            LuminositySlider.Value = 0.8;
         }
 
         private void Slider_ValueChanged(object sender, RangeBaseValueChangedEventArgs e)
         {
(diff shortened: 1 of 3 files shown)


XamlControlsGallery/ControlPages/AcrylicPage.xaml.cs
```diff
@@ -44,5 +49,11 @@ private void FallbackColorSelector_SelectionChanged(object sender, SelectionChan
             Rectangle shape = (sender == FallbackColorSelectorInApp) ? CustomAcrylicShapeInApp : CustomAcrylicShape;
             ((AcrylicBrush)shape.Fill).FallbackColor = ((SolidColorBrush)e.AddedItems.First()).Color;
         }
+
+        private void LuminositySlider_ValueChanged(object sender, RangeBaseValueChangedEventArgs e)
+        {
+            Rectangle shape = CustomAcrylicShapeLumin;
+            ((AcrylicBrush)shape.Fill).TintLuminosityOpacity = e.NewValue;
```
stmoy: Do we need to wrap this in an API check so that it doesn't crash down-level? Or does the property still exist down-level, but it degrades gracefully (ie. is ignored or something)?

kikisaints: Just checked! If running on RS5 or lower, Acrylic will use the old recipe even if the new property is explicitly defined.

stmoy: Thanks, sounds good!

PR #64, opened 2019-02-26
Remove setting PlaceholderForeground explicitly in App.xaml.

ddams:
Remove setting PlaceholderForeground explicitly in App.xaml.

## Description
There is a product bug 19027035 which can be partially blamed on the Control Gallery app since it is used by tester to generate that bug.
Part of the bug description: Hint text should be visible in the focused TextBox in dark theme.
Debugging the Control Gallery app, I realized PlaceholderForeground  property is set explicitly in App.xaml for TextBox and ComboBox style, and it is causing trouble for focused visual state. 

## Motivation and Context
Internal bug 19027035 

##How Has This Been Tested?
Manually verified.

This is my first pull request for XAML Control Gallery , please let me know if I am missing anything.

### DIFF
--- a/XamlControlsGallery/App.xaml
+++ b/XamlControlsGallery/App.xaml
@@ -72,14 +72,6 @@
                 <Setter Property="HorizontalContentAlignment" Value="Stretch" />
             </Style>
 
-            <Style TargetType="TextBox">
-                <Setter Property="PlaceholderForeground" Value="{ThemeResource TextControlPlaceholderForeground}" />
-            </Style>
-
-            <Style TargetType="ComboBox">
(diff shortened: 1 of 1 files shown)


ddams:
> These styles aren't doing anything, right? Looks like the controls in question already set these same brushes in generic.xaml, so specifying them in App.xaml should be equivalent to not specifying them.

The default style changes placeholder brush to different visual states, for example: TextControlPlaceholderForegroundDisabled, TextControlPlaceholderForegroundPointerOver and TextControlPlaceholderForegroundFocused.

Those are ignored if TextBox.PlaceholderForeground is set. I think there maybe some historical reason to set PlaceholderForeground property in App.xaml but I believe it is unnecessary and does not fit current style design.

PR #83, opened 2019-04-17
Use latest prerelease and include new ScrollViewer sample

micahl:
Use latest prerelease and include new ScrollViewer sample

## Description

Fix an error where ItemsRepeater was still marked as preview.
Bump the project to use the latest prerelease WinUI
Add a sample page for the new ScrollViewer control highlighting a few things:
 - the ContentOrientation property
 - Basic behavior (similar to the existing ScrollViewer)
 - the IgnoredInputKind property to selectively ignore inputs
 - Customizing the animation used when programatically panning or zooming

## Motivation and Context

Provide a sample of using the control for developers to see how to use some of the new capabilities.

## How Has This Been Tested?

Manually tested

## Types of changes

### DIFF
--- a/XamlControlsGallery/DataModel/ControlInfoData.json
+++ b/XamlControlsGallery/DataModel/ControlInfoData.json
@@ -1178,7 +1178,7 @@
           "Docs": [
             {
               "Title": "ScrollViewer - API",
-              "Uri": "http://docs.microsoft.com/en-us/windows/uwp/api/microsoft.ui.xaml.controls.scrollviewer"
+              "Uri": "https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.scrollviewer"
             },
             {
               "Title": "Guidance",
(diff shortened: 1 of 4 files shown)


XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -1167,6 +1167,33 @@
...
+          "Title": "ScrollViewer (preview)",
+          "Subtitle": "A container control that lets the user pan and zoom its content.",
+          "ImagePath": "ms-appx:///Assets/ScrollViewer.png",
+          "Description": "The new ScrollViewer lets a user scroll, pan, and zoom to see content that's larger than the viewable area. Many content controls, like ListView, have ScrollViewers built into their control templates to provide automatic scrolling.",
+          "Content": "<p>Look at the <i>ScrollViewer2Page.xaml</i> file in Visual Studio to see the full code for this page.</p>",
+          "IsPreview": true,
+          "Docs": [
+            {
+              "Title": "ScrollViewer - API",
+              "Uri": "http://docs.microsoft.com/en-us/windows/uwp/api/microsoft.ui.xaml.controls.scrollviewer"
```
stmoy: Is this the correct URI?

XamlControlsGallery/XamlControlsGallery.csproj
```diff
@@ -10,7 +10,7 @@
     <AssemblyName>AppUIBasics</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.17763.0</TargetPlatformVersion>
+    <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>
```
stmoy: Is this required? If so, we should merge into MASTER_19H1 branch instead of MASTER.

micahl: No, I don't believe that is required.  What is required, however, is that the latest Win 10 SDK be installed to have a fix in the markup compiler.  I'm assuming that means also having the latest OS. 
 I haven't tried using a new SDK on an old OS to see how well it works.  So, it may be that this should still merge into MASTER_19H1?  Without that fix for the markup compiler there would be a runtime error (only in debug) complaining that it was "Unable to cast object of type 'Microsoft.UI.Xaml.Controls.InputKind' to type 'System.Int32'.  See internal issue 20088938.

XamlControlsGallery/DataModel/ControlInfoData.json
micahl: No.  :)  I'll fix it.

PR #116, opened 2019-07-22
Updated state of app title

YuliKl:
Ensured app title remains visible in all NavigationView modes and states.

## Description

Previously, the "XAML Controls Gallery" title of the app would be hidden whenever the primary NavigationView's pane was closed. As a result, when the gallery ran in a small window its app title would not be visible most of the time. This change keeps the title visible but moves it out of the way of the pane sliver when needed. The title moves to avoid overlapping the pane when appropriate, so I added an implicit animation to make the transition more visually appealing.

Also updated the vertical alignment of the title to look more centered relative to caption controls.

## Motivation and Context

As a best practice, apps should have a visible title. When drawing into the window title bar, the burden to have a title falls on the app itself. The gallery was drawing a title, but this textblock was being hidden on smaller widths when NavigationView defaults to Compact or Minimal modes.

## How Has This Been Tested?

Manually verified.

image

image

## Types of changes

### DIFF
--- a/XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
+++ b/XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
@@ -27,6 +27,7 @@ using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Input;
 using Windows.UI.Xaml.Media;
 using Windows.UI.Xaml.Navigation;
+using Windows.Foundation.Metadata;
 
 namespace AppUIBasics
 {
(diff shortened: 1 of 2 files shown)


XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
```diff
@@ -281,19 +282,34 @@ private void OnControlsSearchBoxQuerySubmitted(AutoSuggestBox sender, AutoSugges
...
+        }
+
+        private void NavigationViewControl_DisplayModeChanged(Microsoft.UI.Xaml.Controls.NavigationView sender, Microsoft.UI.Xaml.Controls.NavigationViewDisplayModeChangedEventArgs args)
+        {
+            UpdateAppTitleMargin(sender);
+        }
+
+        private void UpdateAppTitleMargin(Microsoft.UI.Xaml.Controls.NavigationView sender)
+        {
+            if (!(ApiInformation.IsApiContractPresent("Windows.Foundation.UniversalApiContract", 7)))
```
stmoy: Doesn't this say: "if the API Contract 7 is **not** present, do the thing"? Won't this cause crashes where the API is not present since it will try to set the TranslationTransition?

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
```diff
@@ -281,19 +282,34 @@ private void OnControlsSearchBoxQuerySubmitted(AutoSuggestBox sender, AutoSugges
...
+            {
+                AppTitle.TranslationTransition = new Vector3Transition();
+            }
+
+            if (sender.IsPaneOpen == false && (sender.DisplayMode == Microsoft.UI.Xaml.Controls.NavigationViewDisplayMode.Expanded ||
+                sender.DisplayMode == Microsoft.UI.Xaml.Controls.NavigationViewDisplayMode.Compact))
             {
-                AppTitleBar.Margin = new Thickness(40, 0, 0, 0);
+                AppTitle.Translation = new System.Numerics.Vector3(34, 0, 0);
```
stmoy: Similarly, this will cause crashes where the Translation property is not available. I think we'll need two codepaths: One that sets margin pre-RS5 and one that sets Translation post RS5.

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
YuliKl: Fail. This is what I get for copying code and getting careless.

You're absolutely correct, now fixed.

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
YuliKl: Yeah, I was thinking about Vector3 being available for a while but somehow failed to notice that Translation is new. Also fixed now.

Issue #42, opened 2019-01-24
Add Xaml sample to NavigationView's API in Action section


YuliKl:
**Is this a totally new sample, an addition to an existing sample page, or a modification of an existing sample page?**
- [ ] New sample
- [ ] Addition to existing sample
- [x] Modification to existing sample

**If this is an addition/modification to an existing sample page, which one?**
NavigationViewPage.xaml

**Describe the sample**

Add sample markup that dynamically updates to match current state of the control.

PR #125, opened 2019-07-24
Added code to NavigationView API in action sample

YuliKl:
Added interactive code snippet for the last sample on the NavigationView page.

## Description

Used ControlExample.Xaml control to show code sample. The only options check box not wired correctly in the sample is "Autosuggestbox visible" because I couldn't figure out a good property to bind to.

Also updated the namespace title from *muxcontrols* to *muxc*.

## Motivation and Context

#42 

image

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
@@ -237,10 +237,13 @@ namespace AppUIBasics.ControlPages
                 AutoSuggestBox asb = new AutoSuggestBox() { QueryIcon = new SymbolIcon(Symbol.Find) };
                 asb.SetValue(AutomationProperties.NameProperty, "search");
                 nvSample.AutoSuggestBox = asb;
+
+                navViewASB.Value = "\r\n    <muxc:NavigationView.AutoSuggestBox> \r\n        <AutoSuggestBox QueryIcon=\"Find\" AutomationProperties.Name=\"Search\" /> \r\n    <" + "/" + "muxc:NavigationView.AutoSuggestBox> \r\n";
             }
             else
             {
                 nvSample.AutoSuggestBox = null;
+                navViewASB.Value = null;
             }
         }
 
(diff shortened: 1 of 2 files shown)


stmoy:
> The only options check box not wired correctly in the sample is "Autosuggestbox visible" because I couldn't figure out a good property to bind to.

The ItemsRepeater xaml and cs pages show how to create a ControlExampleSubstitution that doesn't use binding (and instead uses Key and Value).

YuliKl:
> > The only options check box not wired correctly in the sample is "Autosuggestbox visible" because I couldn't figure out a good property to bind to.
> 
> The ItemsRepeater xaml and cs pages show how to create a ControlExampleSubstitution that doesn't use binding (and instead uses Key and Value).

Thank you, that was a helpful pointer. Who knew you can set the Value property in code behind ;D

XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
```diff
@@ -237,10 +237,13 @@ private void autoSuggestCheck_Click(object sender, RoutedEventArgs e)
                 AutoSuggestBox asb = new AutoSuggestBox() { QueryIcon = new SymbolIcon(Symbol.Find) };
                 asb.SetValue(AutomationProperties.NameProperty, "search");
                 nvSample.AutoSuggestBox = asb;
+
+                navViewASB.Value = "\r\n    <muxc:NavigationView.AutoSuggestBox> \r\n        <AutoSuggestBox QueryIcon=\"Find\" AutomationProperties.Name=\"Search\" /> \r\n    <" + "/" + "muxc:NavigationView.AutoSuggestBox> \r\n";
```
stmoy: This is set in both here and in the markup above. Are they the same? If so, can they be condensed?

YuliKl: No, they can't be condensed. This code is inside the CheckBox's click event. When the page first loads, this event doesn't fire. So I need to initialize the ControlExampleSubstitution in markup, otherwise the appropriate ASB markup won't appear on page load. From code behind, I'm toggling the ControlExampleSubstitution's value between null and the full markup string. The value needs to change depending on the state of the CheckBox and toggle appropriately on click.

stmoy: Gotcha, thanks. I'm mostly concerned about the maintenance of the duplicated markup. Would it be possible to encapsulate this code into a method that is called both in the CheckBox Loaded event and also in the Checkbox handler?

YuliKl: Fixed.

Issue #41, opened 2019-01-23
Update CommandBar sample to show its properties dynamically


YuliKl:
**Is this a totally new sample, an addition to an existing sample page, or a modification of an existing sample page?**
- [ ] New sample
- [ ] Addition to existing sample
- [x] Modification to existing sample

**If this is an addition/modification to an existing sample page, which one?**
CommandBarPage.xaml

**Describe the sample**

In the screenshot below, the Xaml does not match CommandBar's current state. Would be great to have the Xaml dynamically update.

**Additional context**

![image](https://user-images.githubusercontent.com/13142665/51642478-48100480-1f1e-11e9-9c39-102977e0447c.png)

PR #129, opened 2019-08-07
Update CommandBar to be update sample code change dynamically; Closes…

### LINKED ISSUE #40: Update sample markup to show IsEnabled property for buttons
- [ ] New sample

ButtonPage.xaml
HyperlinkButtonPage.xaml
RepeatButtonPage.xaml
ToggleButtonPage.xaml

**Describe the sample**

All three of these samples have a setting to disable the control. The property that setting is tied to does not appear in the Xaml example. This should update dynamically. 

**Additional context**

image
<br>
image

chingucoding:
… #41;

Updated code sample of CommandBar to reflect the actual component
## Description

Changed the CommandBar page to display changes in state in the code sample
## Motivation and Context

Fixes #40.
## How Has This Been Tested?

Tested manually by running the test application
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/CommandBarPage.xaml
+++ b/XamlControlsGallery/ControlPages/CommandBarPage.xaml
@@ -64,7 +64,7 @@
             <local:ControlExample.Substitutions>
                 <local:ControlExampleSubstitution Key="IsOpen" IsEnabled="True" Value="{x:Bind PrimaryCommandBar.IsOpen, Mode=OneWay}" />
                 <local:ControlExampleSubstitution Key="IsSticky" IsEnabled="{x:Bind PrimaryCommandBar.IsSticky, Mode=OneWay}" Value=" IsSticky=&quot;True&quot; " />
-                <local:ControlExampleSubstitution Key="MultipleButtonsSecondaryCommands" IsEnabled="{x:Bind MultipleButtons ,Mode=OneWay}" Value="{StaticResource MultipleButtonsSecondaryCommands}"/>
+                <local:ControlExampleSubstitution Key="MultipleButtonsSecondaryCommands" IsEnabled="{x:Bind MultipleButtons, Mode=OneWay}" Value="{StaticResource MultipleButtonsSecondaryCommands}"/>
             </local:ControlExample.Substitutions>
         </local:ControlExample>      
     </StackPanel>
--- a/XamlControlsGallery/ControlPages/CommandBarPage.xaml
+++ b/XamlControlsGallery/ControlPages/CommandBarPage.xaml
@@ -18,6 +18,14 @@
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d">
 
+    <Page.Resources>
+        <x:String x:Key="MultipleButtonsSecondaryCommands" xml:space="preserve">
+        &lt;AppBarButton Icon="Add" Label="Button 1"/&gt;
+        &lt;AppBarButton Icon="Delete" Label="Button 2"/&gt;
+        &lt;AppBarSeparator /&gt;
+        &lt;AppBarButton Icon="FontDecrease" Label="Button 3"/&gt;
+        &lt;AppBarButton Icon="FontIncrease" Label="Button 4"/&gt;</x:String>
+    </Page.Resources>
     <StackPanel>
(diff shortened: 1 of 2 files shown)


XamlControlsGallery/ControlPages/CommandBarPage.xaml.cs
```diff
@@ -18,8 +19,31 @@ namespace AppUIBasics.ControlPages
     /// <summary>
     /// An empty page that can be used on its own or navigated to within a Frame.
     /// </summary>
-    public sealed partial class CommandBarPage : Page
+    public sealed partial class CommandBarPage : Page, INotifyPropertyChanged
```
YuliKl: I may be completely wrong, but I didn't think you needed to implement INotifyPropertyChanged to use x:Bind (as opposed to Binding). What happens when you just declare a public MultipleButtons bool, without the getter and setter?

XamlControlsGallery/ControlPages/CommandBarPage.xaml
```diff
@@ -43,16 +51,21 @@
...
-        &lt;AppBarButton Icon="Setting" Label="Settings"/&gt;
+        &lt;AppBarButton Icon=&quot;Setting&quot; Label=&quot;Settings&quot;/&gt;$(MultipleButtonsSecondaryCommands)
     &lt;/CommandBar.SecondaryCommands&gt;
 &lt;/CommandBar&gt;
 </x:String>
             </local:ControlExample.Xaml>
+            <local:ControlExample.Substitutions>
+                <local:ControlExampleSubstitution Key="IsOpen" IsEnabled="True" Value="{x:Bind PrimaryCommandBar.IsOpen, Mode=OneWay}" />
+                <local:ControlExampleSubstitution Key="IsSticky" IsEnabled="{x:Bind PrimaryCommandBar.IsSticky, Mode=OneWay}" Value=" IsSticky=&quot;True&quot; " />
+                <local:ControlExampleSubstitution Key="MultipleButtonsSecondaryCommands" IsEnabled="{x:Bind MultipleButtons ,Mode=OneWay}" Value="{StaticResource MultipleButtonsSecondaryCommands}"/>
```
YuliKl:
```suggestion
                <local:ControlExampleSubstitution Key="MultipleButtonsSecondaryCommands" IsEnabled="{x:Bind MultipleButtons, Mode=OneWay}" Value="{StaticResource MultipleButtonsSecondaryCommands}"/>
```

XamlControlsGallery/ControlPages/CommandBarPage.xaml.cs
```diff
@@ -45,12 +69,14 @@ private void AddSecondaryCommands_Click(object sender, RoutedEventArgs e)
 
             if (PrimaryCommandBar.SecondaryCommands.Count == 1)
             {
+                MultipleButtons = true;
```
YuliKl: This feedback is really minor, but I would love to see this line at the very end of the If statement. Right now, it reads strangely as the very first line of the condition when exactly one button is visible. Would feel more logical after all the buttons were added to the secondary commands.

chingucoding: Oh yes, that makes more sense. I will change that. Thank you for that feedback ! :smile:

XamlControlsGallery/ControlPages/CommandBarPage.xaml.cs
```diff
@@ -101,6 +128,7 @@ private void RemoveSecondaryCommands()
         {
             while (PrimaryCommandBar.SecondaryCommands.Count > 1)
             {
+                MultipleButtons = false;
```
YuliKl: You don't need to set this inside the while loop. Not a big deal, but it would be slightly more efficient if this line executed only once :)

chingucoding: Fixed that in the newest commit, in addition to being slightly more efficient the code (at least in my opinion) gets a bit more understandable. Thanks for bringing that up !

stmoy: This piece needs to be moved outside too, right?

chingucoding: Oh yes totally missed that, I'll fix this right now.

chingucoding: Now everything should be fine.

PR #135, opened 2019-08-11
Change StandardUICommandPage to select the item it is for; Closes #106;

### LINKED ISSUE #106: Context menu in StandardUICommand sample is broken
**Describe the bug**

Right-clicking an item to delete doesn't work. The item gets desected on right-click and choosing the Delete menu option does nothing.

1. Go to StandardUICommand page
2. Right-click on "List item 2"
3. Click Delete in the open context menu

**Expected behavior**

List item 2 is removed from the list of items.

**Actual behavior**

List item 2 loses its hovered/pressed backplate as the context menu opens, so the context menu doesn't appear connected to any list item. The Delete menu option has no effect.

**Screenshots**

image

**Version Info**

1.2.8.0

chingucoding:
Fixed a bug where the context menu in the StandardUICommandPage would not delete the item when it was not selected.
## Description

Made the menu flyout automatically select the component it is being created for. By doing that, the item in the list is being selected and the deletion of the item will work. In addition to that it is clear to which component the context menu belongs.
## Motivation and Context

Closes #106.
## How Has This Been Tested?

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
@@ -98,8 +98,8 @@ namespace AppUIBasics.ControlPages
             ListItemData data = (ListItemData)args.Item;
             MenuFlyoutItem item = new MenuFlyoutItem() { Command = data.Command };
             flyout.Opened += delegate (object element, object e) {
-                var flyOutElement = element as MenuFlyout;
-                var elementToHighlight = flyOutElement.Target as ListViewItem;
+                MenuFlyout flyoutElement = element as MenuFlyout;
+                ListViewItem elementToHighlight = flyoutElement.Target as ListViewItem;
                 elementToHighlight.IsSelected = true;
             };
             flyout.Items.Add(item);
--- a/XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.ObjectModel;
 using System.Windows.Input;
 using Windows.Foundation.Metadata;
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
```diff
@@ -96,7 +96,12 @@ private void ListViewRight_ContainerContentChanging(ListViewBase sender, Contain
         {
             MenuFlyout flyout = new MenuFlyout();
             ListItemData data = (ListItemData)args.Item;
-            MenuFlyoutItem item = new MenuFlyoutItem() { Command = data.Command};
+            MenuFlyoutItem item = new MenuFlyoutItem() { Command = data.Command };
+            flyout.Opened += delegate (object element, object e) {
+                var flyOutElement = element as MenuFlyout;
```
YuliKl:
```suggestion
                MenuFlyout flyoutElement = element as MenuFlyout;
```

You're casting your variable to a MenuFlyout anyway, so might as well declare it as such for clarity. And I would update the capitalization in the name to better align with the control's capitalization (Flyout).

XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
```diff
@@ -96,7 +96,12 @@ private void ListViewRight_ContainerContentChanging(ListViewBase sender, Contain
...
+                var elementToHighlight = flyOutElement.Target as ListViewItem;
```
YuliKl:
```suggestion
                ListViewItem elementToHighlight = flyoutElement.Target as ListViewItem;
```

Similar feedback - declaring the type instead of using var feels more clear.

XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
chingucoding: I agree, that makes the code more readable. Was probably a relic of refactoring.

XamlControlsGallery/ControlPages/StandardUICommandPage.xaml.cs
chingucoding: Again: I agree. Thank you for the feedback! ^^

PR #138, opened 2019-08-13
Focus searchbar onstart

### LINKED ISSUE #109: focus on search upon launch
- [ ] New sample
- [ ] Modification to existing sample
None of the above 

None

**Describe the sample**

N/A

Not a sample request

**Additional context**

chingucoding:
Made app focus the searchbar (AutoSuggestBox) at app start.
## Description

Changed the function OnNewControlsMenuItemLoaded to instead of focusing the "New page" tab to focus the searchbar.
## Motivation and Context

Closes #109.
## How Has This Been Tested?

## Types of changes

### DIFF
--- a/XamlControlsGallery/Navigation/NavigationRootPage.xaml
+++ b/XamlControlsGallery/Navigation/NavigationRootPage.xaml
@@ -71,8 +71,6 @@
             Header=" "
             HeaderTemplate="{StaticResource NavigationViewHeaderTemplate}"
             IsTabStop="False"
-            IsPaneOpen="False"
-            PaneDisplayMode="Auto"
             PaneOpening="NavigationViewControl_PaneOpened"
             PaneClosing="NavigationViewControl_PaneClosing"
             DisplayModeChanged="NavigationViewControl_DisplayModeChanged"
--- a/XamlControlsGallery/Navigation/NavigationRootPage.xaml
+++ b/XamlControlsGallery/Navigation/NavigationRootPage.xaml
@@ -71,6 +71,7 @@
             Header=" "
             HeaderTemplate="{StaticResource NavigationViewHeaderTemplate}"
             IsTabStop="False"
+            IsPaneOpen="False"
             PaneOpening="NavigationViewControl_PaneOpened"
             PaneClosing="NavigationViewControl_PaneClosing"
             DisplayModeChanged="NavigationViewControl_DisplayModeChanged"
--- a/XamlControlsGallery/Navigation/NavigationRootPage.xaml
+++ b/XamlControlsGallery/Navigation/NavigationRootPage.xaml
@@ -71,6 +71,8 @@
             Header=" "
             HeaderTemplate="{StaticResource NavigationViewHeaderTemplate}"
             IsTabStop="False"
+            IsPaneOpen="False"
+            PaneDisplayMode="Auto"
             PaneOpening="NavigationViewControl_PaneOpened"
             PaneClosing="NavigationViewControl_PaneClosing"
             DisplayModeChanged="NavigationViewControl_DisplayModeChanged"
--- a/XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
+++ b/XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
@@ -169,9 +169,8 @@ namespace AppUIBasics
         {
             if (IsFocusSupported)
             {
-                _newControlsMenuItem.Focus(FocusState.Keyboard);
+                controlsSearchBox.Focus(FocusState.Keyboard);
             }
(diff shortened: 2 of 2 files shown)


chingucoding:
Oh yes, this makes sense. I'll commit the fix right now.

chingucoding:
So what i changed:
- The navigation pane now is in display mode auto, so it will open in the corresponding mode for the different screen sizes. In addition to this the NavigationPane.IsOpen value is preset to "False" for it to not be open when the navigation pane is in mode "Minimal" or "Compact".
- The function handling the focusing of the search checks wether the navigation pane is opened.

XamlControlsGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -71,6 +71,8 @@
             Header=" "
             HeaderTemplate="{StaticResource NavigationViewHeaderTemplate}"
             IsTabStop="False"
+            IsPaneOpen="False"
```
YuliKl: You don't want to set this. When NavigationView's DisplayMode=Expanded, the Pane should be open by default. That's how you get this:
image

When IsPaneOpen="False" and DisplayMode=Expanded, you'll see this, which isn't how we want the app to start up:
image

XamlControlsGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -71,6 +71,8 @@
...
+            PaneDisplayMode="Auto"
```
YuliKl: Auto is the default PaneDisplayMode value. So while this line doesn't hurt anything, it's also not necessary.

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
```diff
@@ -167,11 +167,10 @@ private void SetDeviceFamily()
 
         private void OnNewControlsMenuItemLoaded(object sender, RoutedEventArgs e)
         {
-            if (IsFocusSupported)
+            if (IsFocusSupported && NavigationViewControl.IsPaneOpen)
```
YuliKl: Instead of checking for whether or not the Pane is open, I recommend checking whether the DisplayMode is Expanded.

XamlControlsGallery/Navigation/NavigationRootPage.xaml
chingucoding: Cleaned that up in the newest commit.

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
chingucoding: Fixed that in the newest commit.

YuliKl:
I'm confused. For some reason, with your latest iteration, I'm still seeing the root NavigationView open its Pane on app launch. Focus is on the hamburger button and I can see that the NavigationView thinks it's in Minimal display mode...but something is forcing the pane to open when it shouldn't. @chingucoding, do you also see this behavior, or is it just on my machine?

chingucoding:
Strangely enough the behavior is (sometimes) observable on my machine too.
When we set `IsPaneOpen="False"` the apps however behaves as we want to: On smaller sizes the pane is closed, and when the screen is large enough the pane is open and the search has the focus.
At least that is the behavior I could observe for now on my machine.

YuliKl:
> When we set `IsPaneOpen="False"` the apps however behaves as we want to: On smaller sizes the pane is closed, and when the screen is large enough the pane is open and the search has the focus.

Agreed, this results in the correct behavior. Please set IsPaneOpen to false.

I'll just go sit in the corner and contemplate why NavigationView is not behaving the way I expect. 🤷‍♂️

chingucoding:
> I'll just go sit in the corner and contemplate why NavigationView is not behaving the way I expect. 🤷‍♂️

This is truly mind boggling. 

If you want I can look further into this to check if this is a bug on our side or if that is a problem in the underlying API. ^^

YuliKl:
> If you want I can look further into this to check if this is a bug on our side or if that is a problem in the underlying API. ^^

Thank you, but investigating is probably not a good use of your time. You've already figured out how get the right behavior in the app. If your investigation turns up that there's an underlying problem in the NavigationView control, my team is pretty unlikely to fix the control bug - because an easy workaround on the app side already exists. So I think we scratch our heads and move on.

XamlControlsGallery/Navigation/NavigationRootPage.xaml
stmoy: Just double checking since this thread is still open - @YuliKl was your comment addressed?

YuliKl: Yes. I'm happy with where the code changes ended up.

PR #145, opened 2019-08-15
Fix "bug" where OnNavigatedFrom was not called; Closes #74;

### LINKED ISSUE #74: Navigation events are fired when moving to a page but not from.
**Describe the bug**

OnNavigatedFrom and related navigation events do not surface when navigated from any control's sample page.

1. Add OnNavigatedFrom function to any control's sample page codebehind (e.g., TeachingTipPage.xaml.cs) with an attached break point.
2. Navigated to and from this page. 

**Expected behavior**

I expect the break point to be triggered.

**Actual behavior**

The break point is not triggered. 

**Screenshots**

**Version Info**

1.1.1.0

Windows 10 version:
- [X] Insider Build (18865)
- [ ] Creators Update (15063)

Device form factor:
- [X] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

I am attempting to add a function to close open teaching tips when navigating away from the TeachingTip sample page and OnNavigatedFrom is not working.

chingucoding:
Fixed a "bug" where the OnNavigatedFrom event of the ControlPages were not being called.
## Description

I added a call to the OnNavigatedFrom method from the Pages manually in the ItemPage.xaml.cs file. 

This change/fix is a bit more complex since the OnNavigatedFrom method is not being called because the hierarchical structure of the Page/XAML is not layed out in a way that would trigger the method on the ControlPage.
Oversimplified diagram of controls nesting:
```
<App>
    <NavigationRootPage>
        <ItemPage> <---- "Root" page / "source of frame"
            <-ControlSamplePage- /> <--------- Not the root page / "source of frame"
        </ItemPage>
    </NavigationRootPage>
</App>
```

Since the individual control pages are not the source of frame, OnNavigatedFrom will not be called (at least according to [this](https://stackoverflow.com/questions/53109162/uwp-page-onnavigatedfrom-not-called-when-app-is-terminated) )
However since the ItemPage(.xaml) is the source of frame its OnNavigatedFrom will be called. 
In the ItemPages OnNavigatedFrom method the nested controls page  OnNavigatedFrom method will now be called. This is done using reflection since the methode Page.OnNavigatedFrom is protected and can not be accessed otherwise. 

## Motivation and Context

Fixes #74.
## How Has This Been Tested?

Tested manually by starting application and checking execution of PullToRefreshPage.xaml.cs using debugger.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ItemPage.xaml.cs
+++ b/XamlControlsGallery/ItemPage.xaml.cs
@@ -203,6 +203,8 @@ namespace AppUIBasics
                 ConnectedAnimationService.GetForCurrentView().PrepareToAnimate("controlAnimation", target);
             }
 
+            // We use reflection to call the OnNavigatedFrom function the user leaves this page
+            // See this PR for more information: https://github.com/microsoft/Xaml-Controls-Gallery/pull/145
             Frame contentFrameAsFrame = contentFrame as Frame;
             Page innerPage = contentFrameAsFrame.Content as Page;
             MethodInfo dynMethod = innerPage.GetType().GetMethod("OnNavigatedFrom",
--- a/XamlControlsGallery/ItemPage.xaml.cs
+++ b/XamlControlsGallery/ItemPage.xaml.cs
@@ -1,4 +1,4 @@
-﻿//*********************************************************
+//*********************************************************
 //
 // Copyright (c) Microsoft. All rights reserved.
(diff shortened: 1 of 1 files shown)


chingucoding:
Yes , totally missed that. I will add a comment ! :)

PR #157, opened 2019-08-25
Fix bug where losing focus removes color; Closes #144;

### LINKED ISSUE #144: Color of text missing after RichEditBox loses focus in dark theme
**Describe the bug**

**This only applies to dark theme**
When the text has been colored using the SplitButton, and the user leaves the RichEditBox (by clicking outside) the text color changes to white (since we are using dark theme). After clicking the box again the color does not change back to the original one (the color chosen using the SplitButton) but instead the text is color black.

1. Open SplitButton page (while using dark mode for the whole app or the sample)
2. Select text and color it in any color except black
3. Click outside of RichEditBox
4. Click inside the RichEditBox

**Expected behavior**

The text returns to the previously applied colors

**Actual behavior**

The text is colored black.

**Screenshots**

**Version Info**

1.2.8.0

Windows 10 version:
- [x] Insider Build (18956)
- [ ] May 2019 Update (18362)
- [ ] Creators Update (15063)

Device form factor:
- [x] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

This applies to the SplitButton page and the RichEditBox page.

### LINKED ISSUE #146: Changing color in RichEditBox editor sample removes previously applied color
**Describe the bug**

When changing the color in the RichEditBox - Editor sample, the color of text written before changes to black instead of staying in that color.

1. Go to RichEditBox page to the "custom editor" sample
2. Change color to any color except black
3. Write some text
4. Change color

**Expected behavior**

The newly written text will be in the new color while the old text stays in the color it was written in.

**Actual behavior**

The text changes it color to black

**Screenshots**

**Version Info**

1.2.8.0
Windows 10 version:
- [x] Insider Build (18956)
- [ ] May 2019 Update (18362)
- [ ] Creators Update (15063)

Device form factor:
- [x] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

chingucoding:
Fixed a bug where losing focus and reentering the RichEditBox with dark theme enabled would remove colors previously applied.
## Description

Saved the formatted RTF to a string and reapplied it when the user enters the RichEditBox again.
## Motivation and Context

Fixes #144.
## How Has This Been Tested?

Tested manually by entering text with different colors, clicking outside of the RichEditBox and reentering it again.
## Types of changes


chingucoding:
> any chance you know what might be causing this 
[This line](https://github.com/microsoft/Xaml-Controls-Gallery/pull/157/commits/89f33e7ec78c11880097a902ed38b1cbc83c2ea4#diff-721ab38a7675d91c721f449d73d04947R178) is responsible for the shifting of the caret position.
I assume setting the text "confuses" the RichEditBox since the caret might not be able to be at the same position so instead of trying to restore it, it gets set to the start.
> how to fix it?
What we could try is to restore the caret position to where it was before the color was changed.

chingucoding:
Everything should now fix as exspected.

chingucoding:
As it seems now reentering the rich edit box does not restore the colors that were applied before the color was lost :/

Regarding the behavior of the text search box: It seems like setting the background explicetly messes with the restoration of the string, however I have absolutely no idea why.

Having a RichEditBox with text colorization and text search highlight seems way more complicated than anticipated when the app is running using dark theme ...

I will look further into this tomorrow since it is already 2 am here 😅

chingucoding:
Everything should work now and the code sample are up to date.
@YuliKl While working on this I wondered if this behavior is actually wanted and makes sense to maintain for component. Since the component essentially resets the style on dark theme everytime it gains focus, maybe the internal behaviour should change. Should we create an issue over at WinUI?

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -9,6 +9,7 @@
 //*********************************************************
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
stmoy: What is this include used for?

chingucoding: Thanks for pointing that out! It is not necessary and I removed it.

PR #198, opened 2019-09-05
App title bar

### LINKED ISSUE #141: Fix appearance of the top of the Gallery
**Describe the bug**

There's an unwanted stripe at the top of the app above the hero image. When the fix for [Ability to turn off "TopPadding" in NavigationView](https://github.com/microsoft/microsoft-ui-xaml/issues/47) becomes available in the NuGet package used by the gallery, we should remove this extra whitespace.

**Screenshots**

image

YuliKl:
The goal of this change was to validate and take advantage of NavigationView's new IsTitleBarAutoPaddingEnabled API. To make the app title bar look seamless, I ended up rewriting most of the What's New page, primarily to avoid having the app title overlap with the page content.

## Description

I would particularly appreciate some feedback about NewControlsPage, especially its code behind.

- Added a gradient over the hero image on All and New controls pages to keep app title legible. 
- Tightened up margins and fixed some visual alignment issues
- Deleted useless state groups from PageHeader

I'm still not happy with the appearance of the app in the most narrow widths, will try to tackle that in a later PR.

## Motivation and Context

Fixes #141 

## How Has This Been Tested?

Manually verified on desktop at different app widths and themes

## Types of changes


YuliKl:
@chingucoding, do you by any chance have time to provide feedback?

XamlControlsGallery/XamlControlsGallery.csproj
```diff
@@ -1082,7 +1082,7 @@
       <Version>6.1.9</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.UI.Xaml">
-      <Version>2.2.190611001-prerelease</Version>
+      <Version>2.2.190822001-prerelease</Version>
```
stmoy: Is this correct? 2.2 is released now, right?

YuliKl: It is released, and doesn't include any pre-release features. So when I updated to release version, the ScrollViewer (preview) page loudly complained.

chingucoding:
> @chingucoding, do you by any chance have time to provide feedback?

Yes, sure.

XamlControlsGallery/PageHeader.xaml
```diff
@@ -117,9 +82,9 @@
                         </AppBarButton.Icon>
                         <AppBarButton.Resources>
                             <controls:TeachingTip x:Name="ToggleThemeTeachingTip1"                   
-                            Target="{x:Bind ThemeButton}"
-                            Title="Change themes without hassle"                                        
-                            Subtitle="It's easier than ever to see control samples in both light and dark theme!">
+                                Target="{x:Bind ThemeButton}"
```
chingucoding: For some reason, now the "Toggle Theme" button interlaces with the apps caption controls:

image

XamlControlsGallery/NewControlsPage.xaml.cs
```diff
@@ -26,35 +27,74 @@ public NewControlsPage()
...
-            menuItem.IsSelected = true;
-            NavigationRootPage.Current.NavigationView.Header = menuItem.Content;
-            Items = ControlInfoDataSource.Instance.Groups.SelectMany(g => g.Items.Where(i => i.IsNew)).OrderBy(i => i.Title).ToList();
-            UpdatedItems = ControlInfoDataSource.Instance.Groups.SelectMany(g => g.Items.Where(i => i.IsUpdated)).OrderBy(i => i.Title).ToList();
-            PreviewItems = ControlInfoDataSource.Instance.Groups.SelectMany(g => g.Items.Where(i => i.IsPreview)).OrderBy(i => i.Title).ToList();
         }
+        public object Key { get; set; }
 
-        protected override bool GetIsNarrowLayoutState()
+        private string _title;
```
chingucoding: This can be replaced with an auto property, but it does not make much of a difference in the end. 😅

XamlControlsGallery/NewControlsPage.xaml.cs
```diff
@@ -15,7 +15,8 @@
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Navigation;
 using System.Collections.Generic;
-
+using System.Collections.ObjectModel;
+using System.Threading.Tasks;
```
chingucoding: I think this using directive is unecessary ^^

XamlControlsGallery/App.xaml
```diff
@@ -40,6 +40,12 @@
                 </ResourceDictionary>
             </ResourceDictionary.ThemeDictionaries>
 
+            <LinearGradientBrush x:Key="HeroImageGradientBrush" StartPoint="0.5,0" EndPoint="0.5,1">
+                <GradientStop Offset="0" Color="{ThemeResource SystemAltMediumColor}"/>
+                <GradientStop Offset="0.3" Color="{ThemeResource SystemAltMediumColor}"/>
```
chingucoding: Maybe we could change the 0.3 to 0.5 so the "dark top border" of the hero image is high enough that it sort of resembles the standard border apps have. This way, in my opinion, it would blend in a bit more natural as the bottom border of that "top border" would line up nicely with the caption controls and the app name on the left. 

Current behaviour:
image image

With 0.5 as GradientStop:
image  image

Of course this however boils down to personal preference. Either way this is a really nice touch to the hero image!

XamlControlsGallery/AllControlsPage.xaml.cs
```diff
@@ -31,7 +31,8 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
         {
             var menuItem = NavigationRootPage.Current.NavigationView.MenuItems.Cast<Microsoft.UI.Xaml.Controls.NavigationViewItem>().ElementAt(1);
             menuItem.IsSelected = true;
-            NavigationRootPage.Current.NavigationView.Header = menuItem.Content;
+            NavigationRootPage.Current.NavigationView.Header = string.Empty;
```
chingucoding: Just out of curiosity: Do we need this line? And if so what exactly does it do?  
I have commented it out and I saw no difference in the AllControlsPage. 😅

YuliKl: I think whether or not this line does anything depends on whether or not the Header text is already set. The NavigationView Header only shows up on narrow widths, but I think that without this line we run the risk of showing the wrong header depending on the path the user took to get to this page.

YuliKl:
Thank you @chingucoding, I really appreciate your time. Your feedback is addressed with the latest commit.

XamlControlsGallery/AllControlsPage.xaml.cs
chingucoding: Oh okay, I see. We should leave it that way than just to be save.

chingucoding:
> Thank you @chingucoding, I really appreciate your time. Your feedback is addressed with the latest commit.

Thanky you @YuliKl for asking me for feedback 😄

YuliKl:
@stmoy I've sync'ed to master, including the change you made to the nuget package version. Can you merge this PR if you approve of the changes?

stmoy:
The scrollbar appears below the header rectangle on the All Controls page, meaning the scroll thumb can't be interacted with by default since the rectangle eats the input. Is this new with this change or an existing bug?

image

XamlControlsGallery/App.xaml
```diff
@@ -40,6 +40,12 @@
                 </ResourceDictionary>
             </ResourceDictionary.ThemeDictionaries>
 
+            <LinearGradientBrush x:Key="HeroImageGradientBrush" StartPoint="0.5,0" EndPoint="0.5,1">
+                <GradientStop Offset="0" Color="{ThemeResource SystemAltMediumColor}"/>
```
stmoy: Nit: since the colors at offset 0 and offset 0.5 are the same, I don't think you need the one at offset 0. I tried it out and I _think_ it looks the same, but I might be missing something. Not important, just calling it out for FYI.

XamlControlsGallery/App.xaml
```diff
@@ -50,24 +56,41 @@
...
                 <Setter Property="Margin" Value="8,8,0,0" />
             </Style>
-
+           
+            <Style x:Name="NarrowHeaderTextBlockStyle" BasedOn="{StaticResource HeaderTextBlockStyle}" TargetType="TextBlock">
+                <Setter Property="Foreground" Value="{StaticResource SystemControlForegroundChromeWhiteBrush}"/>
+                <Setter Property="VerticalAlignment" Value="Top"/>
+                <Setter Property="HorizontalAlignment" Value="Left"/>
+                <Setter Property="Margin" Value="36,12,0,0"/>
+            </Style>
```
stmoy: Can we align with the caption button heights? My guess is "probably not" based on the height of the hamburger and back buttons, but this looks so strange to me: 

image

XamlControlsGallery/PageHeader.xaml
```diff
@@ -98,7 +63,7 @@
                     TextTrimming="CharacterEllipsis"
                     TextWrapping="NoWrap" />
             </Grid>
-            <Border x:Name="commandBarBorder" Grid.Column="1">
+            <Border x:Name="commandBarBorder" Grid.Column="1" VerticalAlignment="Bottom">
```
stmoy: It looks like the header text is very indented. Expected? 

image

chingucoding: The problem originates in the NavigationView which creates a left padding which is as wide as the back button and the toggle pane button so the header does not overlap with the two buttons.

See [this line in NavigationView.cpp](https://github.com/chingucoding/microsoft-ui-xaml/blob/b7984c3f76ad3598493e5cf9e9cad23ce71bacd8/dev/NavigationView/NavigationView.cpp#L3188).

stmoy: Hmm! I think this well and good when the content is supposed to be inline with the buttons (ie. where the "XAML Controls Gallery" text is) but not so great when the title is below (ie. the "Acrylic" text). 

What if we used @YuliKl's favorite tool: negative margins? 😜

chingucoding: Haha, a negative margin would indeed fix this. But it would create a new problem:
The left padding is only present in minimal mode. So we would have to listen for that or otherwise the text starts clipping into the navigation pane of the NavigationView.

Setting the correct negative margin and not clearing it for non minimal displaymodes would result in this behaviour:
navview-header-alignbug

So it would be a bit more work to fix that behavior, but with the beloved negative margin it would be possible 😅

YuliKl:
> The scrollbar appears below the header rectangle on the Whats New & All Controls pages, meaning the scroll thumb can't be interacted with by default since the rectangle eats the input. Is this new with this change or an existing bug?

Good find. This was actually a problem in the All Controls page before, but not in the New Controls page. Now fixed on both (when I push out a new iteration)

XamlControlsGallery/App.xaml
YuliKl: Unfortunately no, because the caption controls are 32px tall and the back/hamburger buttons are 40px tall. But I do want to take another pass over Minimal mode in general because there are several (most) things I don't like about it. I mostly want to get this massive change in first so I can take some time to play with just the narrow app state of things.

XamlControlsGallery/PageHeader.xaml
YuliKl: You're highlighting the thing I dislike most about the Minimal mode layout right now. I need more time to work on it, and would prefer not to block this PR on that change.
So: yes, expected; no, not pretty.

PR #202, opened 2019-09-10
Fix crashing with acrylic page, Closes #107

### LINKED ISSUE #107: Styles > Acrylic page still crashes in 1.2.9.0
**Describe the bug**

Since updating the Xaml Controls Gallery app in the store to version 1.2.8 (which includes 19H1 and WinUI2.2 preview), the app has become unstable per reports in the dev center.

One known issue is that the Acrylic Luminosity Opacity slider causes a crash. Is that the only issue?

**Version Info**

1.2.8

**Additional context**

chingucoding:
Fix the problem with the acrylic sample page crashing. Reenabled the TintLuminosityOpacity sample for Windows 10 Build 18362 users.
## Description

The problem for the crashing was that we casted an Microsoft.UI.Xaml.Media.AcrylicBrush to Windows.UI.Xaml.Media.AcrylicBrush which failed and crashed the app.
## Motivation and Context

Closes #107 
## How Has This Been Tested?

Tested manually by starting application and navigating to Acrylic page.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/AcrylicPage.xaml
+++ b/XamlControlsGallery/ControlPages/AcrylicPage.xaml
@@ -4,9 +4,7 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:AppUIBasics"
     xmlns:media="using:Microsoft.UI.Xaml.Media"
-    xmlns:preview="using:Microsoft.UI.Xaml.Controls"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-    xmlns:contract8Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 8)"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d">
     <Page.Resources>
@@ -200,8 +198,7 @@
             </local:ControlExample.Substitutions>
         </local:ControlExample>
 
-        <!-- Disable this sample until we find the root cause of the Luminosity slider crash. -->
-        <local:ControlExample x:Name="Example5" HeaderText="Luminosity with in-app Acrylic." MinimumUniversalAPIContract="8">
+        <local:ControlExample x:Name="Example5" HeaderText="Luminosity with in-app Acrylic.">
             <local:ControlExample.Example>
                 <Grid x:Name="Example5Grid" MinWidth="650" MinHeight="200">
                     <Grid.ColumnDefinitions>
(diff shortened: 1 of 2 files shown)


XamlControlsGallery/ControlPages/AcrylicPage.xaml
```diff
@@ -198,10 +199,9 @@
                 <local:ControlExampleSubstitution Key="FallbackColor" Value="{x:Bind FallbackColorSelector.SelectedItem, Mode=OneWay}"/>
             </local:ControlExample.Substitutions>
         </local:ControlExample>
-        
+
         <!-- Disable this sample until we find the root cause of the Luminosity slider crash. -->
-        <local:ControlExample x:Name="Example5" HeaderText="Luminosity with in-app Acrylic." 
-                              Visibility="Collapsed">
+        <local:ControlExample x:Name="Example5" HeaderText="Luminosity with in-app Acrylic." MinimumUniversalAPIContract="8">
```
stmoy: Is `MinimumUniversalAPIContract="8"` required if we're using Microsoft.UI.Xaml.Acrylic? I suspect not. If it is, then I'm guessing we need more version checks inside the sample as well?

@kikisaints

chingucoding: According to Visual Studio, the property "TintLuminosityOpacity" is not available on Windows build 17134. However using the quickfix, it adds the UniversalApiContract 8 value (which again fails the build). So to prevent crashes in apps that are not running on devices with UniversalApiContract 8, I added this MinimumUniversalApiContract.

stmoy: Overall, @chingucoding is changing to use MUX acrylic not WUX acrylic. Is this a case where something in WinUI only works on certain uplevel versions of the OS? In general, I thought WinUI controls worked downlevel, hence my question about doing an OS check here. 

My question is: is this OS check necessary since we're using MUX acrylic? If this OS check is required, do we also need to do API checks in the codebehind that set the luminosity?

YuliKl: What I'm seeing is that setting the ControlExample's `MinimumUniversalAPIContract="8"` does not appear to be necessary. I do get a compiler error when trying to set `TintLuminosityOpacity` in markup (e.g. as one of the parameters of CustomAcrylicInAppLuminosity):
>The 'Nullable type assignment' feature is not supported in Min Target Platform Version '10.0.17134.0'. Update your Min Target Platform Version to '10.0.17763.0' or above.

I suspect this is the VS compiler being overly helpful and not accounting for the NuGet package. I may be wrong.

I'm not seeing any errors or warnings around this line of code
```c#
((Microsoft.UI.Xaml.Media.AcrylicBrush)shape.Fill).TintLuminosityOpacity = e.NewValue;
```
And I definitely saw some really bad behavior when we had inadvertently mixed **M**UXC.NavigationView and **W**UXC.NavigationView calls in RootNavigationView. So I believe @chingucoding fixed the crashing bug by explicitly calling MUXM.AcrylicBrush.

My recommendation: remove `MinimumUniversalAPIContract="8"` and `xmlns:contract8Present`, then merge this PR.

PR #203, opened 2019-09-10
Fix bug with changing focus and apply changes from #157 to SplitButton page, Closes #200

### LINKED ISSUE #200: Split Button: Not able to perform tab/shift-tab navigation after performing Bold/Italic/Underline operation in edit box
**Describe the bug**

1. Steps to reproduce the behavior:
1. Activate Split button from All controls page.
1. Take the keyboard focus to the rich edit box, press ctrl+a to select the text, then press ctrl+b/i/u.
1. Now press tab/shift-tab and observe.

**Expected behavior**

User should be able to navigate in the page after performing Bold/Italic/Underline context menu operation.

**Actual behavior**

- Keyboard focus gets struck.
- The application doesn't respond to tab/shift-tab.
- Not able to regain focus with alt+tab too.

**Screenshots**

**Version Info**

1.2.9.0

Windows 10 version:
- [ ] Insider Build (xxxxx)
- [ ] May 2019 Update (18362)
- [ ] Creators Update (15063)

Device form factor:
- [ ] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

Jevan mentions: It works, but after doing this then tab/shift+tab is very slow. I suspect a perf issue in the app. Maybe a runaway callback?

chingucoding:
Fix a bug where hitting control+b and similar commands will result in navigating the app with focus becoming nearly impossible
## Description

Changed the event that was used since the previous event for text colorization would create an endless loop.

I also added the code that was introduced to the RichEditBox in #157 so that the same behavior does not occur in the SplitButton sample.
## Motivation and Context

Closes #200
## How Has This Been Tested?

Tested manually by starting application, visiting the RichEditBox on the SplitButton page and hitting control+b. After that I was able to successfully navigate to other commands using tab navigation.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/SplitButtonPage.xaml
+++ b/XamlControlsGallery/ControlPages/SplitButtonPage.xaml
@@ -88,7 +88,11 @@
                     </muxcontrols:SplitButton.Flyout>
                 </muxcontrols:SplitButton>
 
-                <RichEditBox x:Name="myRichEditBox" Grid.Column="1" MinWidth="240" MinHeight="96" PlaceholderText="Type something here" TextChanged="MyRichEditBox_TextChanged"/>
+                <RichEditBox x:Name="myRichEditBox" Grid.Column="1" MinWidth="240" MinHeight="96" 
+                             PlaceholderText="Type something here" 
+                             GotFocus="MyRichEditBox_GotFocus"
+                             LosingFocus="MyRichEditBox_LosingFocus"
+                             TextChanging="MyRichEditBox_TextChanging"/>
             </Grid>
         </local:ControlExample>
     </StackPanel>
--- a/XamlControlsGallery/ControlPages/SplitButtonPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/SplitButtonPage.xaml.cs
@@ -1,4 +1,5 @@
 ﻿using Windows.UI;
+using Windows.UI.Text;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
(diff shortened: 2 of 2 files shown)


stmoy:
Philosophical question for you both (@YuliKl @chingucoding) - this change introduces a bunch of codebehind to the sample, but this code isn't included in the sample code shown on the page. Should it be? Or is this just a bunch of "glue" to help make the sample work as expected?

YuliKl:
> Philosophical question for you both (@YuliKl @chingucoding) - this change introduces a bunch of codebehind to the sample, but this code isn't included in the sample code shown on the page. Should it be? Or is this just a bunch of "glue" to help make the sample work as expected?

I definitely think of the code as "glue". It has nothing to do with SplitButton per se, just makes the coloring scenario work better.

chingucoding:
I agree with @YuliKl, most of the code is just "glue" to deal with the text changing its color. 

One thing we might do is create an empty event listener function. In the sample, the SplitButton "uses" the myColorButton_Click function. Maybe we could add that function with a comment inside saying something like "Handling button click". 
However I don't really think that it would be that helpful to just have a C# code block with an empty function. What do you think about that @stmoy and @YuliKl ?

stmoy:
> However I don't really think that it would be that helpful to just have a C# code block with an empty function. What do you think about that @stmoy and @YuliKl ?

Agreed, I don't think it's super useful. I also agree with the overall direction as Yulia described. Net result: no changes needed.

PR #213, opened 2019-09-26
Fix bug where app would crash when opening MediaPlayerElement page

### LINKED ISSUE #211: MediaPlayerElement page crashes
**Describe the bug**

Choosing the MediaPlayerElement page crashes

1. Launch the Gallery
1. Go to Media category
1. Click on MediaPlayerElement

**Expected behavior**

Crash :(

**Actual behavior**

No crash :)

**Screenshots**

**Version Info**

Windows 10 version:
- [x] Insider Build (xxxxx)
- [ ] May 2019 Update (18362)
- [ ] Creators Update (15063)

Device form factor:
- [x] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

It looks like the crash occurs in the ImageLoader class. 

Why is ImageLoader looking for the animations.png in the first place? 
image

chingucoding:
## Description

Fixed the image link for the AnimatedVisualPlayer page, which was also needed the MediaPlayerElement page.
## Motivation and Context

Fixes #211 
## How Has This Been Tested?

Tested manually by starting application and navigating to MediaPlayeElement page.
## Types of changes

### DIFF
--- a/XamlControlsGallery/DataModel/ControlInfoData.json
+++ b/XamlControlsGallery/DataModel/ControlInfoData.json
@@ -1527,7 +1527,7 @@
           "UniqueId": "AnimatedVisualPlayer",
           "Title": "AnimatedVisualPlayer",
           "Subtitle": "An element to render and control playback of motion graphics.",
-          "ImagePath": "Assets/Animations.png",
+          "ImagePath": "ms-appx:///Assets/Animations.png",
           "Description": "An element to render and control playback of motion graphics.",
           "Content": "<p>Look at the <i>AnimatedVisualPlayerPage.xaml</i> and <i>AnimatedVisualPlayerPage.xaml.cs</i> files in Visual Studio to see the full code for this page.</p>",
           "IsNew": true,


chingucoding:
> The good news is that this change fixes the crash. The bad news is that I have no idea why...

The thing is that the image was needed as the AnimatedVisualPlayer is listed in the "related controls" section. And since that URI was faulty, it crashed the app. At least that is my guess. Maybe we should check if other pages that have that dependency also crash without that fix.

YuliKl:
> The thing is that the image was needed as the AnimatedVisualPlayer is listed in the "related controls" section. And since that URI was faulty, it crashed the app. At least that is my guess. Maybe we should check if other pages that have that dependency also crash without that fix.

Oh, I see. It's a little odd that our app is loading anything from "related controls". But taking that behavior as a given, the technique to fix the problem makes perfect sense.

chingucoding:
> > The thing is that the image was needed as the AnimatedVisualPlayer is listed in the "related controls" section. And since that URI was faulty, it crashed the app. At least that is my guess. Maybe we should check if other pages that have that dependency also crash without that fix.
> 
> Oh, I see. It's a little odd that our app is loading anything from "related controls". But taking that behavior as a given, the technique to fix the problem makes perfect sense.

It sort of makes sense, since the icon sort of is part of the controls page. Everywhere we display the controls (All Controls page in XCG, ToolBox in Visual Studio), they appear with their respective icon so it sort of makes sense we do the same in the "related controls" section. 

But most importantly the bug is fixed!

PR #208, opened 2019-09-20
Add RevealStyle sample for Split- and DropDownbutton

chingucoding:
## Description

Add new samples for the RevealStyle applied to SplitButton and DropDownButton
## Motivation and Context

Closes #155 
## How Has This Been Tested?

Tested manually by starting application.
## Screenshots (if appropriate):
image
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #216, opened 2019-10-03
Updated references to MUXC

YuliKl:
Updated all instances of using:Microsoft.UI.Xaml.Controls to use mux prefix

## Description

Searched for all instances of using:Microsoft.UI.Xaml.Controls, ensured they all use the same xmlns:muxc prefix. Updated sample code to also use "muxc".

I anticipated merge conflicts with #208 but they haven't materialized. Perhaps I just haven't waited long enough for the PR to propagate...? 

## How Has This Been Tested?

manually verified

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/XamlUICommandPage.xaml
+++ b/XamlControlsGallery/ControlPages/XamlUICommandPage.xaml
@@ -2,10 +2,8 @@
       xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
       xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" 
       xmlns:local="using:AppUIBasics"
-      xmlns:local2="using:AppUIBasics.ControlPages"
       xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
       xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-      xmlns:muxcontrols="using:Microsoft.UI.Xaml.Controls"
       mc:Ignorable="d">
     <Page.Resources>
         <Style x:Key="HorizontalSwipe" TargetType="ListViewItem" BasedOn="{StaticResource ListViewItemRevealStyle}">
--- a/XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
+++ b/XamlControlsGallery/ControlPages/AnimatedVisualPlayerPage.xaml
@@ -6,7 +6,7 @@
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:animatedvisuals="using:AnimatedVisuals"
     xmlns:local="using:AppUIBasics"
-    xmlns:muxcontrols="using:Microsoft.UI.Xaml.Controls"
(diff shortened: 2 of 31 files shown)


YuliKl:
Found and resolved the merge conflict. Should be all set now for review and merge.

XamlControlsGallery/ControlPages/XamlUICommandPage.xaml
```diff
@@ -2,10 +2,8 @@
       xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
       xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" 
       xmlns:local="using:AppUIBasics"
-      xmlns:local2="using:AppUIBasics.ControlPages"
       xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
       xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-      xmlns:muxcontrols="using:Microsoft.UI.Xaml.Controls"
```
stmoy: This isn't needed? Are the XamlUICommands below coming from WUX or MUX?

stmoy:
Philosophical question: should we preface these with "muxc" or "winui"? There doesn't need to be a 1:1 mapping between the alias and the namespace. In fact, the alias is usually a bit more semantic (like "local").

XamlControlsGallery/ControlPages/XamlUICommandPage.xaml
YuliKl: Apparently from WUX, which surprised me. I first renamed it, than noticed that the reference was grayed out. Navigated to the page a couple of times after removing to make sure I didn't break anything :)

YuliKl:
> Philosophical question: should we preface these with "muxc" or "winui"? There doesn't need to be a 1:1 mapping between the alias and the namespace. In fact, the alias is usually a bit more semantic (like "local").

I thought we had settled on muxc for our docs. But that may have been wishful thinking on my part because I like the one-to-one symmetry of simply abbreviating Microsoft.UI.Xaml.Controls.

I'm definitely open to feedback on this question.

chingucoding:
> > Philosophical question: should we preface these with "muxc" or "winui"? There doesn't need to be a 1:1 mapping between the alias and the namespace. In fact, the alias is usually a bit more semantic (like "local").
> 
> I thought we had settled on muxc for our docs. But that may have been wishful thinking on my part because I like the one-to-one symmetry of simply abbreviating Microsoft.UI.Xaml.Controls.
> 
> I'm definitely open to feedback on this question.

I agree with @stmoy on this. Usually one would choose those alias's to be semantic (e.g. contrct7present). Having said that, both "muxc" (which not only stands for the package but partially also for the github url) and "winui" get the point across in my opinion. 

In my opinion, from a developer standpoint "muxc" makes more sense however, since "winui" is the product name, but not the technical name (which is Microsoft.UI.Xaml.Controls). So if we think from a developer standpoint, I think it makes a bit more sense to refer to the controls as "muxc" to make a bit more clearer what the origin of those components is. But on the other hand, "muxc" is a bit ambiguous, if one does not know what it stands for, "winui" is pretty self explanatory.

PR #222, opened 2019-10-14
Updated GridView Samples

anawishnoff:
## Description
For the GridView page, I added two samples:
1. Basic Example that shows extensive source code, including code-behind
2. Layout example that shows how to increase space between columns/rows as well as setting a max number of items to layout before wrapping.

I also edited the existing sample to improve its functionality (make drag and drop work, eliminate swipe), and edited the sample code to make it more detailed and helpful.

## Motivation and Context

This update makes GridView more accessible to new users through the basic sample, as it provides a starting point that users can copy/paste and test on their own. The other new sample provides helpful layout information that may not have been easily found. The overall changes to the code samples will improve users' understanding of how the control works.
## How Has This Been Tested?

Manually tested by me, with corner cases for all of the input fields (negative numbers, non-numbers, etc).
Basic sample:
CGGV1

Layout sample:
cggv2

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)


XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -14,8 +14,9 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:AppUIBasics"
-    xmlns:local1="using:AppUIBasics.ControlPages">
-    
+    xmlns:local1="using:AppUIBasics.ControlPages"
+    xmlns:sys="using:System">
```
chingucoding: I think `sys` is unused in that page.

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -66,26 +67,200 @@
...
+                            <RowDefinition Height="Auto"></RowDefinition>
+                        </Grid.RowDefinitions>
+
+                        <TextBlock Grid.Row="0" Grid.ColumnSpan="3" Text="Space Between Columns" Width="150" TextWrapping="Wrap" Margin="0,0,0,5" HorizontalAlignment="Left"/>
+                        <TextBox Grid.Row="1" Grid.Column="0" Width="100" x:Name="ColSpace" PlaceholderText="{x:Bind ActualColSpace, Mode=TwoWay}"  Margin="0,0,5,10" 
+                                TextChanged="StyledGrid_ChangeCol"/>
+                        <Button x:Name="ColSpacesDec" Grid.Row="1" Grid.Column="1" Content="-" FontSize="15" Width="30" Click="StyledGrid_DecreaseColSpace" 
+                                Margin="0,0,5,10" VerticalAlignment="Bottom" HorizontalAlignment="Left"/>
+                        <Button x:Name="ColSpacesInc" Grid.Row="1" Grid.Column="2" Content="+" FontSize="15" Width="30" Click="StyledGrid_IncreaseColSpace" 
+                                Margin="0,0,0,10" VerticalAlignment="Bottom" HorizontalAlignment="Left"/>
```
chingucoding: I think it would be better to disable the "-" button if we have reached 0 as value.

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -66,26 +67,200 @@
...
+                            <RowDefinition Height="Auto"></RowDefinition>
+                        </Grid.RowDefinitions>
+
+                        <TextBlock Grid.Row="0" Grid.ColumnSpan="3" Text="Space Between Rows" Width="150" TextWrapping="Wrap" Margin="0,0,0,5" HorizontalAlignment="Left"/>
+                        <TextBox Grid.Row="1" Grid.Column="0" Width="100" x:Name="RowSpace" PlaceholderText="{x:Bind ActualRowSpace, Mode=TwoWay}"  Margin="0,0,5,10" 
+                                 TextChanged="StyledGrid_ChangeRow"/>
+                        <Button x:Name="RowSpacesDec" Grid.Row="1" Grid.Column="1" Content="-" FontSize="15" Width="30" Click="StyledGrid_DecreaseRowSpace" 
+                                Margin="0,0,5,10" VerticalAlignment="Bottom" HorizontalAlignment="Left"/>
+                        <Button x:Name="RowSpacesInc" Grid.Row="1" Grid.Column="2" Content="+" FontSize="15" Width="30" Click="StyledGrid_IncreaseRowSpace" 
+                                Margin="0,0,0,10" VerticalAlignment="Bottom" HorizontalAlignment="Left"/>
```
chingucoding: Same here, disabling the button would be better in my opinion.

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -66,26 +67,200 @@
...
+                            <RowDefinition Height="Auto"></RowDefinition>
+                        </Grid.RowDefinitions>
+
+                        <TextBlock Grid.Row="0" Grid.ColumnSpan="3" Text="Max # of Items Before Wrapping" Width="150" TextWrapping="Wrap" Margin="0,0,0,5" HorizontalAlignment="Left"/>
+                        <TextBox Grid.Row="1" Grid.Column="0" Width="100" x:Name="MaxItems" PlaceholderText="{x:Bind ActualMaxItems, Mode=TwoWay}"  
+                                 Margin="0,0,5,10" TextChanged="StyledGrid_ChangeMaxItems" HorizontalAlignment="Left"/>
+                        <Button x:Name="MaxItemsDec" Grid.Row="1" Grid.Column="1" Content="-" FontSize="15" Width="30" Click="StyledGrid_DecreaseMaxItems" 
+                                Margin="0,0,5,10" VerticalAlignment="Bottom" HorizontalAlignment="Left"/>
+                        <Button x:Name="MaxItemsInc" Grid.Row="1" Grid.Column="2" Content="+" FontSize="15" Width="30" Click="StyledGrid_IncreaseMaxItems" 
+                                Margin="0,0,0,10" VerticalAlignment="Bottom" HorizontalAlignment="Left"/>
```
chingucoding: Disabling the "-" button when the value is 1 would be better, especially as sometimes one can see the TextBox changing to 0 and back to 1.

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -114,9 +289,11 @@
                             Tag="TextTemplate" />
                         <ToggleButton Margin="0,8,0,0" Click="FlowDirectionCheckBox_Click" Content="Reverse FlowDirection" />
                         <TextBlock Margin="0,18,0,10" Text="GridView Properties" />
+                        <TextBlock TextWrapping="Wrap" FontSize="13" MaxWidth="150"> In order to drag, drop, and reorder items within the GridView, make sure all boxes are checked below.</TextBlock>
```
chingucoding: Interestingly, I was able to reorder the items without having to check the "IsItemClickEnabled" checkbox. This was also possible when selection mode was set to none. So this text is a bit misleading I think.

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -9,94 +9,504 @@
 //*********************************************************
 using AppUIBasics.Common;
 using AppUIBasics.Data;
+using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Diagnostics;
 using System.Linq;
+using System.Reflection.Metadata;
+using System.Security.AccessControl;
```
chingucoding: I don't think we need those 😅

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -9,94 +9,504 @@
...
+        private void StyledGrid_InitWrapGrid(object sender, RoutedEventArgs e)
+        {
+            // Update ItemsWrapGrid object created on page load by assigning it to StyledGrid's ItemWrapGrid
+            StyledGridIWG = sender as ItemsWrapGrid;
+
+            // Now we can change StyledGrid's MaximumRowsorColumns property within its ItemsPanel>ItemsPanelTemplate>ItemsWrapGrid.
+            StyledGridIWG.MaximumRowsOrColumns = ActualMaxItems;
+        }
+
+        private async void sendMessage()
```
chingucoding: nit: rename to `SendMessage`

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -9,94 +9,504 @@
...
+                    item.Margin = NewMargin;
+                }
+            }
+
+            // If new value is negative, reset text box to 0 and display an error message.
+            else
+            {
+                ColSpace.Text = "";
+                ColSpace.PlaceholderText = "0";
+                sendMessage();
```
chingucoding: The MessageDialog will say "Value cannot be less than 1." which is not true for the column and row spacing. So instead, we should state that the value can not be negative (or prevent the scenario altogether).

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -9,94 +9,504 @@
...
+                    item.Margin = NewMargin;
+                }
+            }
+
+            // If new value is negative, reset text box to 0 and display an error message.
+            else
+            {
+                RowSpace.Text = "";
+                RowSpace.PlaceholderText = "0";
+                sendMessage();
```
chingucoding: Same as line 264.

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -9,94 +9,504 @@
...
+            // Update ItemsWrapGrid object created on page load by assigning it to StyledGrid's ItemWrapGrid
+            StyledGridIWG = sender as ItemsWrapGrid;
+
+            // Now we can change StyledGrid's MaximumRowsorColumns property within its ItemsPanel>ItemsPanelTemplate>ItemsWrapGrid.
+            StyledGridIWG.MaximumRowsOrColumns = ActualMaxItems;
+        }
+
+        private async void sendMessage()
+        {
+            MessageDialog messageDialog = new MessageDialog("Value cannot be less than 1.");
```
chingucoding: The message dialog does not respect the theme of the app correctly.

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -9,94 +9,504 @@
...
+                item.Margin = NewMargin;
+            }
+
+            NotifyPropertyChanged();
+        }
+
+        private void StyledGrid_DecreaseColSpace(object sender, RoutedEventArgs e)
+        {
+            // Make sure new value is positive
+            if (ActualColSpace - 1 >= 0)
```
chingucoding: We can simply check if ActualColSpace is larger than 0, which is a bit easier to read in my opinion.

XamlControlsGallery/ControlPages/GridViewPage.xaml
anawishnoff: Tried to clarify a bit in my upcoming commit. Thanks for pointing this out!

XamlControlsGallery/ControlPages/GridViewPage.xaml
anawishnoff: Switched all the buttons to disable when a negative/below 1 value is entered, and removed code surrounding message dialog/sendMessage() function. Thanks!

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -114,9 +288,12 @@
                             Tag="TextTemplate" />
                         <ToggleButton Margin="0,8,0,0" Click="FlowDirectionCheckBox_Click" Content="Reverse FlowDirection" />
```
chingucoding: I think it would be nice to also reflect the FlowDirection in the XAML sample. 😅

anawishnoff: Nice catch! Fixed in upcoming commit.

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -66,26 +66,201 @@
...
+                            <ColumnDefinition Width="Auto"></ColumnDefinition>
+                            <ColumnDefinition Width="Auto"></ColumnDefinition>
+                        </Grid.ColumnDefinitions>
+                        <Grid.RowDefinitions>
+                            <RowDefinition Height="Auto"></RowDefinition>
+                            <RowDefinition Height="Auto"></RowDefinition>
+                        </Grid.RowDefinitions>
+
+                        <TextBlock Grid.Row="0" Grid.ColumnSpan="3" Text="Space Between Columns" Width="150" TextWrapping="Wrap" Margin="0,0,0,5" HorizontalAlignment="Left"/>
+                        <TextBox Grid.Row="1" Grid.Column="0" Width="100" x:Name="ColSpace" PlaceholderText="{x:Bind ActualColSpace, Mode=TwoWay}"  Margin="0,0,5,10"
```
chingucoding: If one enters text, then presses the "clear button" of the TextBox and then hits the increase/decrease button, the app crashes as we are unable to convert "" to int, and thus fail to increase the value. Maybe setting a default value if the TextBox loses focus could fix this.

anawishnoff: Only saw that issue on my end for changing the MaxItems, but I'm about to push a fix where everything works correctly over here.

YuliKl:
In dark theme, the item text isn't legible:
image

XamlControlsGallery/ControlPagesSampleCode/GridView/GridViewSample1_xaml.txt
```diff
@@ -0,0 +1,25 @@
...
+    ItemTemplate="{StaticResource ImageTemplate}"
+    IsItemClickEnabled="True"
+    ItemClick="BasicGridView_ItemClick"
+    SelectionMode="Single"/>
+                    
+<DataTemplate x:Key="ImageTemplate" x:DataType="local1:CustomDataObject">
+     <Image Stretch="UniformToFill" Source="{x:Bind ImageLocation}" Width="190" Height="130">
+</DataTemplate>
+                    
+<!-- In this example, the ListView's ItemTemplate property is bound to a data template (shown above)
```
YuliKl:
```suggestion
<!-- In this example, the GridView's ItemTemplate property is bound to a data template (shown above)
```

XamlControlsGallery/ControlPagesSampleCode/GridView/GridViewSample1_xaml.txt
```diff
@@ -0,0 +1,25 @@
...
+called ImageTemplate, defined in the Page.Resources section of the XAML file. 
+                    
+The data template is defined to display an object of type CustomDataObject, which is created from a class
+that's defined in the code-behind (see code below).
+                    
+The ItemsSource for the ListView is an ObservableCollection<CustomDataObject> called Items and
```
YuliKl:
```suggestion
The ItemsSource for the GridView is an ObservableCollection<CustomDataObject> called Items and
```

XamlControlsGallery/ControlPagesSampleCode/GridView/GridViewSample1_xaml.txt
```diff
@@ -0,0 +1,25 @@
...
+is defined in the C# code-behind. It is bound to the ListView's ItemsSource property in the code-behind.
```
YuliKl:
```suggestion
is defined in the C# code-behind. It is bound to the GridView's ItemsSource property in the code-behind.
```

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -66,26 +66,201 @@
...
+                        </Grid.ColumnDefinitions>
+                        <Grid.RowDefinitions>
+                            <RowDefinition Height="Auto"></RowDefinition>
+                            <RowDefinition Height="Auto"></RowDefinition>
+                        </Grid.RowDefinitions>
+
+                        <TextBlock Grid.Row="0" Grid.ColumnSpan="3" Text="Space Between Columns" Width="150" TextWrapping="Wrap" Margin="0,0,0,5" HorizontalAlignment="Left"/>
+                        <TextBox Grid.Row="1" Grid.Column="0" Width="100" x:Name="ColSpace" PlaceholderText="{x:Bind ActualColSpace, Mode=TwoWay}"  Margin="0,0,5,10" 
+                                TextChanged="StyledGrid_ChangeCol"/>
+                        <Button x:Name="ColSpacesDec" Grid.Row="1" Grid.Column="1" Content="-" FontSize="15" Width="30" Click="StyledGrid_DecreaseColSpace"
```
YuliKl: These should be RepeatButtons so users can just press down to increment a bunch (instead of making discreet clicks).

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -66,26 +66,201 @@
...
+
+                </StackPanel>
+            </local:ControlExample.Options>
+            <local:ControlExample.Xaml>
+                <x:String xml:space="preserve">
+&lt;!-- The GridView used for this example is shown below. Setter properties are used to customize 
+some parts of the GridViewItems (i.e. the margins). --&gt;
+&lt;GridView
+    x:Name="StyledGrid"
+    ItemTemplate="{StaticResource StyledGridTemplate}"&gt;
```
YuliKl: I think the actual code is using ImageOverlayDataTemplate

anawishnoff:
> In dark theme, the item text isn't legible:
> image

Changed to use BackgroundBaseMediumBrush to better support the shift between light and dark mode.

stmoy:
@anawishnoff - it looks like recent pulls into master has caused merge conflicts. Can you please resolve these conflicts?

anawishnoff:
@stmoy done!

PR #224, opened 2019-10-15
Update Media.CS UI Tests

hassanuz:
Fixes UI Test failure for Media.CS.

### DIFF
--- a/UITests/Media.cs
+++ b/UITests/Media.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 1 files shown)


UITests/Media.cs
```diff
@@ -24,11 +24,7 @@ public static void ClassInitialize(TestContext context)
             var button = session.FindElementByName("MediaPlayerElement");
             button.Click();
             var mediaElements = session.FindElementsByClassName("MediaPlayerElement");
-            Assert.IsTrue(mediaElements.Count >= 2);
-            mediaElement1 = mediaElements[0];
```
stmoy: Should the mediaElement1 and mediaElement2 variables defined on lines 15 and 16 also be deleted?

hassanuz: Removed.

PR #242, opened 2019-11-01
changed MinHeight of headerRoot to 68

salmanmkc:
66 doesn't scale well, 66 * 1.25 = 82.5, therefore it rounds. 68 doesn't need to round when scaled, since 68*1.25 = 85

## Motivation and Context
Improve scaling

## How Has This Been Tested?
Building
## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/XamlControlsGallery/PageHeader.xaml
+++ b/XamlControlsGallery/PageHeader.xaml
@@ -46,7 +46,7 @@
 
         <Rectangle x:Name="WideBackground" Fill="{Binding ElementName=headerControl, Path=Background}" Opacity="{Binding ElementName=headerControl, Path=BackgroundColorOpacity}" />
         
-        <Grid x:Name="headerRoot" Padding="{Binding ElementName=headerControl, Path=Padding}" VerticalAlignment="Top" MinHeight="66">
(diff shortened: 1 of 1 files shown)


YuliKl:
I'm hesitant to take this change because I have spent **a lot** of time fiddling with PageHeader as it seems unpredictable and temperamental. Making this height better for rounding seems like a good change, I'm just not sure that 68 is the correct number. Maybe defaulting to slightly shorter would be better. 

I'll keep this in mind for the next round of PageHeader tweaks that I'll inevitably end up making for one reason or anther :)

salmanmkc:
> I'm hesitant to take this change because I have spent **a lot** of time fiddling with PageHeader as it seems unpredictable and temperamental. Making this height better for rounding seems like a good change, I'm just not sure that 68 is the correct number. Maybe defaulting to slightly shorter would be better.
> 
> I'll keep this in mind for the next round of PageHeader tweaks that I'll inevitably end up making for one reason or anther :)

Reduced to 64. 64*1.25 = 80, whole number, works nicely.

PR #243, opened 2019-11-01
added vertical alignments for boarderpage radio buttons

salmanmkc:
Alignments can be off due to scaling

## Description
https://github.com/microsoft/microsoft-ui-xaml/issues/1520

## Motivation and Context
Should stop any misalignment with radio buttons on this page

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/XamlControlsGallery/ControlPages/BorderPage.xaml
+++ b/XamlControlsGallery/ControlPages/BorderPage.xaml
@@ -1,4 +1,4 @@
-﻿<!--
+<!--
     //*********************************************************
     //
     // Copyright (c) Microsoft. All rights reserved.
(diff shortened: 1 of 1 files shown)


YuliKl:
Changing the vertical alignment of the overall control won't fix the scaling problems. The issue is in the control template and tracked by https://github.com/microsoft/microsoft-ui-xaml/issues/1520.

I recommend abandoning this PR.

salmanmkc:
> Changing the vertical alignment of the overall control won't fix the scaling problems. The issue is in the control template and tracked by [microsoft/microsoft-ui-xaml#1520](https://github.com/microsoft/microsoft-ui-xaml/issues/1520).
> 
> I recommend abandoning this PR.

Yep, that is the issue that I referenced when making the PR
image

salmanmkc:
and I thought the problem they're talking about is multi-line now?

YuliKl:
> and I thought the problem they're talking about is multi-line now?

The glyph drawing above and visually disconnected from the text is a problem for both single-line and multi-line radioButtons/checkBoxes. We can debate which case looks more confusing, but the problem is the same.

Have you tried validating your fix? Do you have an after screenshot showing that changing the alignment fixes this? (I'm happy to be proven wrong, I'm just fairly certain that the fix needs to be in the ControlTemplate rather than each app that uses these controls.)

salmanmkc:
> > and I thought the problem they're talking about is multi-line now?
> 
> The glyph drawing above and visually disconnected from the text is a problem for both single-line and multi-line radioButtons/checkBoxes. We can debate which case looks more confusing, but the problem is the same.
> 
> Have you tried validating your fix? Do you have an after screenshot showing that changing the alignment fixes this? (I'm happy to be proven wrong, I'm just fairly certain that the fix needs to be in the ControlTemplate rather than each app that uses these controls.)

I couldn't even replicate it before, I tried changing scaling, resizing the Window, etc. couldn't actually reproduce the misalignment.

YuliKl:
> I couldn't even replicate it before, I tried changing scaling, resizing the Window, etc. couldn't actually reproduce the misalignment.

To repro, open the Windows Settings app. Go to Ease of Access, in Display page drag the Make Text Bigger Slider to its maximum and click Apply.
image

PR #244, opened 2019-11-01
Alignments for radiocontrols and checkboxes on NavigationViewPage

salmanmkc:
Added alignments for all the Radiocontrols and checkboxes on each page in case they don't align when scaling fails
## Description
Just set the VerticalAlignment property for each to center

## Motivation and Context
To have consistent scaling.
https://github.com/microsoft/microsoft-ui-xaml/issues/1520

## How Has This Been Tested?
Building

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)


YuliKl:
Same comment as for #243. This change won't fix the problem introduced by scaling.

PR #262, opened 2019-11-12
Width of code snippet expanded to remove horizontal scrolling

salmanmkc:
Changed the width of the content frame so that it is wide enough for wide content code snippets

## Description
Scrolling vertically can be interrupted by a horizontal scroll in the code snippet when the users mouse enters the vicinity of the code snippet area. I have changed the width for WideFrameContent ever so slightly so that there is no longer a need to scroll horizontally. At least as tested on the Tab View page. This also removes the difficulty of reading code that barely doesn't fit, and requires the user to scroll horizontally to see the content on the Tab View page, and other pages hopefully as well.

## Motivation and Context
Code snippets can sometimes be a bit to wide to fit on the page, so to see all of it you have to scroll horizontally, I wanted to remove the need for that, so it's easier to read by the user. When vertically scrolling and entering an area which has horizontal scrolling, stops the vertical scroll and starts scrolling horizontally, therefore if someones mouse is in the area of the code snippet whilst vertically scrolling, it will start scrolling horizontally in the code snippet. This removes that, at least when tested on 1080p, which most people use.

## How Has This Been Tested?
Building before and after, this property is set whenever you load a page based on the width of the content.

Before
image

After
image

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/XamlControlsGallery/ItemPage.xaml
+++ b/XamlControlsGallery/ItemPage.xaml
@@ -51,7 +51,7 @@
                 <VisualState x:Name="NormalFrameContent" />
                 <VisualState x:Name="WideFrameContent">
                     <VisualState.Setters>
-                        <Setter Target="contentFrame.Width" Value="1000" />
+                        <Setter Target="contentFrame.Width" Value="1028" />
                         <Setter Target="contentFrame.HorizontalAlignment" Value="Left" />
                     </VisualState.Setters>
                 </VisualState>


salmanmkc:
1028 is also divisible by 4, therefore it scales well

stmoy:
What happens when the window is horizontally shrunk to become less wide? My guess is that it'll still horizontally scroll?

salmanmkc:
> What happens when the window is horizontally shrunk to become less wide? My guess is that it'll still horizontally scroll?

Yes, but only if you shrink it a lot, for example here you can see it still shows the whole code snippet whilst having the horizontal width reduced some substantial amount.

image

stmoy:
I'm still a little confused. If the window is small (like the below screenshot), horizontal scrolling is still necessary. It seems like the fix in this PR doesn't address the underlying issue - it just hides the issue at large screen sizes? 

image

salmanmkc:
> I'm still a little confused. If the window is small (like the below screenshot), horizontal scrolling is still necessary. It seems like the fix in this PR doesn't address the underlying issue - it just hides the issue at large screen sizes?
> 
> image

The issue to me is that there was a need to scroll horizontally when there was extra screen space that wasn't being used that could have been utlised so that horizontal scrolling is not required.

YuliKl:
I don't think this change makes much of a difference. I have not found any down-sides, so am fine with making this change.

PR #263, opened 2019-11-12
You clicked messages added to MenuBar page

salmanmkc:
Adding to MenuBar page as requested in issue: [258](https://github.com/microsoft/Xaml-Controls-Gallery/issues/258)
## Description
Adds messages below saying which MenuFlyoutItem/RadioMenuFlyoutItem was clicked by the user. Also removed some "..."s for consistent formatting.

## Motivation and Context
Issue highlighted [here](https://github.com/microsoft/Xaml-Controls-Gallery/issues/258)

## How Has This Been Tested?
Building and testing different types of buttons: MenuFlyoutItem and RadioMenuFlyoutItem.

Nothing shows when you click a MenuFlyoutItem/RadioMenuFlyoutItem
image.

After:
Text below each control which shows which MenuFlyoutItem/RadioMenuFlyoutItem was clicked last
image

image

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/XamlControlsGallery/ControlExample.xaml
+++ b/XamlControlsGallery/ControlExample.xaml
@@ -64,7 +64,6 @@
         </Grid.RowDefinitions>
 
         <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,15" Text="{x:Bind HeaderText}" />
-        
         <RichTextBlock x:Name="ErrorTextBlock" Visibility="Collapsed" Grid.Row="1" Margin="0,0,0,12" Foreground="Red" IsTextSelectionEnabled="True">
             <Paragraph>
                 This sample requires a later version of Windows to be fully functional. Learn more about version adaptive apps:
--- a/XamlControlsGallery/ControlExample.xaml
+++ b/XamlControlsGallery/ControlExample.xaml
@@ -63,7 +63,7 @@
             <RowDefinition />
         </Grid.RowDefinitions>
 
-        <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,16" Text="{x:Bind HeaderText}" />
+        <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,15" Text="{x:Bind HeaderText}" />
         <RichTextBlock x:Name="ErrorTextBlock" Visibility="Collapsed" Grid.Row="1" Margin="0,0,0,12" Foreground="Red" IsTextSelectionEnabled="True">
             <Paragraph>
                 This sample requires a later version of Windows to be fully functional. Learn more about version adaptive apps:
--- a/XamlControlsGallery/ControlExample.xaml
+++ b/XamlControlsGallery/ControlExample.xaml
@@ -64,8 +64,6 @@
         </Grid.RowDefinitions>
 
         <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,16" Text="{x:Bind HeaderText}" />
-        <TextBlock Margin="0,40,0,4" Text="{x:Bind onClickedText, Mode=TwoWay}"/>
-        
         <RichTextBlock x:Name="ErrorTextBlock" Visibility="Collapsed" Grid.Row="1" Margin="0,0,0,12" Foreground="Red" IsTextSelectionEnabled="True">
             <Paragraph>
                 This sample requires a later version of Windows to be fully functional. Learn more about version adaptive apps:
--- a/XamlControlsGallery/ControlExample.xaml
+++ b/XamlControlsGallery/ControlExample.xaml
@@ -64,6 +64,7 @@
         </Grid.RowDefinitions>
 
         <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,15" Text="{x:Bind HeaderText}" />
+        <TextBlock Margin="0,40" Text="{x:Bind onClickedText}"/>
 
         <RichTextBlock x:Name="ErrorTextBlock" Visibility="Collapsed" Grid.Row="1" Margin="0,0,0,12" Foreground="Red" IsTextSelectionEnabled="True">
             <Paragraph>
--- a/XamlControlsGallery/ControlExample.xaml
+++ b/XamlControlsGallery/ControlExample.xaml
@@ -64,6 +64,7 @@
         </Grid.RowDefinitions>
 
         <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,15" Text="{x:Bind HeaderText}" />
+
         <RichTextBlock x:Name="ErrorTextBlock" Visibility="Collapsed" Grid.Row="1" Margin="0,0,0,12" Foreground="Red" IsTextSelectionEnabled="True">
             <Paragraph>
                 This sample requires a later version of Windows to be fully functional. Learn more about version adaptive apps:
--- a/XamlControlsGallery/ControlExample.xaml
+++ b/XamlControlsGallery/ControlExample.xaml
@@ -63,9 +63,9 @@
             <RowDefinition />
         </Grid.RowDefinitions>
 
-        <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,15" Text="{x:Bind HeaderText}" />
-        <TextBlock Margin="0,40" Text="{x:Bind onClickedText, Mode=TwoWay}"/>
-
+        <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,16" Text="{x:Bind HeaderText}" />
+        <TextBlock Margin="0,40,0,4" Text="{x:Bind onClickedText, Mode=TwoWay}"/>
+        
         <RichTextBlock x:Name="ErrorTextBlock" Visibility="Collapsed" Grid.Row="1" Margin="0,0,0,12" Foreground="Red" IsTextSelectionEnabled="True">
             <Paragraph>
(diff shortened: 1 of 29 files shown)


MikeHillberg:
This is great! Now you can actually see the keyboard accelerators working.

Would it be possible to move the "you clicked" text to be above the sample markup? Some of the samples are long enough that the message is scrolled out of view.

MikeHillberg:
In the AppBarButtonPage case, it puts the text inside the <ControlExample> content to make it show up alongside the buttons. Putting it on the side of a MenuBar would probably look odd though, since a MenuBar typically takes up the whole width. And putting it under the MenuBar but above the sample would create an ugly gap.

So I think the way you have it might be best.

salmanmkc:
> In the AppBarButtonPage case, it puts the text inside the content to make it show up alongside the buttons. Putting it on the side of a MenuBar would probably look odd though, since a MenuBar typically takes up the whole width. And putting it under the MenuBar but above the sample would create an ugly gap.
> 
> So I think the way you have it might be best.

How does this look?
image

MikeHillberg:
I like your first way better.

salmanmkc:
> I like your first way better.

the issue with that way as you mentioned is that you can't see the text unless you scroll down, this removes the need to scroll down, I've modified the ControlsExample, so this can be applied to other pages as well, I will see how it looks on them.

salmanmkc:
> I like your first way better.

The issue with that way, as you mentioned, is that you can't see the text unless you scroll down, this removes the need to scroll down, I've modified the ControlsExample, so this can be applied to other pages as well, I will see how it looks on them.

salmanmkc:
Here, this looks a lot cleaner now, tidied up the margins, removed the old gaps from before
image

MikeHillberg:
That looks great!

salmanmkc:
made the changes on the CommandBarFlyout page, which was just a property change and the removal of the TextBlock since I changed the ControlExample itself, not sure if I should have waited for the PR of this first?

XamlControlsGallery/ControlExample.xaml.cs
```diff
@@ -90,6 +90,13 @@ public string HeaderText
             set { SetValue(HeaderTextProperty, value); }
         }
 
+        public static DependencyProperty onClickedTextProperty = DependencyProperty.Register("onClickedText", typeof(string), typeof(ControlExample), new PropertyMetadata(null));
```
stmoy: Nit: Please capitalize these variables.

XamlControlsGallery/ControlPages/MenuBarPage.xaml.cs
```diff
@@ -26,5 +26,19 @@ public MenuBarPage()
         {
             this.InitializeComponent();
         }
+
+        private void OnElementClicked(object sender, Windows.UI.Xaml.RoutedEventArgs e)
```
stmoy: All three of these events have the same logic. Why can't they be combined into a single OnElementClicked method?

XamlControlsGallery/ControlExample.xaml
```diff
@@ -63,8 +63,9 @@
             <RowDefinition />
         </Grid.RowDefinitions>
 
-        <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,15" Text="{x:Bind HeaderText}" />
-
+        <TextBlock Style="{ThemeResource SubtitleTextBlockStyle}" Margin="0,16" Text="{x:Bind HeaderText}" />
+        <TextBlock Margin="0,40,0,4" Text="{x:Bind onClickedText, Mode=TwoWay}"/>
```
stmoy: This logic adds the "OnClickedText" textblock to every control example, even if it wasn't needed. This adds unnecessary elements to every single control example. 

This change should localize the TextBlock specifically to the pages that need it, similar to the CommandBarFlyout page.

XamlControlsGallery/ControlPages/MenuBarPage.xaml.cs
salmanmkc: yes, I'm not sure how to get the parent element, I've tired doing 
```
var example = ((sender as MenuFlyoutItem).Parent as ControlExample);
            example.onClickedText = "You clicked: " + (sender as MenuFlyoutItem).Text;
```

XamlControlsGallery/ControlExample.xaml
salmanmkc: I did another PR which utilises the same control, I was thinking of doing more for basically all the pages that have have user controls. The change I had at the beginning, showing the textblock below the code snippet was local to the page, however in order to get it above, and below the header I needed to modify the whole ControlExample, unless there's another way to do it?

stmoy: This logic is specifically applicable only to particular examples. My recommendation would be to put the TextBlock within the ControlExample. For example, consider something like:

```
        <local:ControlExample HeaderText="A simple MenuBar" 
                              XamlSource="MenuBar\MenuBarSample1.txt">
            <StackPanel>
                <TextBlock Text="You clicked..." />
                <muxc:MenuBar>
                    ...
                </muxc:MenuBar>
            </StackPanel>
        </local:ControlExample>
```

salmanmkc: I can find numerous examples which would utilise this change

salmanmkc: I can find numerous examples which would utilise this change but I will do as you request

stmoy: Generally the APIs on ControlExample should be widely applicable to most/all examples. I agree that there are several places that could use this sort of text, but I don't think that most/all examples need the click handler text. Hence the suggestion above.

Additionally, the inclusion of another TextBlock is another UIElement in the tree whether the specific example required the element or not. This causes an unnecessary performance hit for most controls. 

The benefit to including this text in the ControlExample would be consistency and ease of use (which are good benefits!) but since the feature isn't generally required, I believe we should encapsulate the logic to the specific examples that need them.

XamlControlsGallery/ControlPages/MenuBarPage.xaml.cs
salmanmkc: just for learning purposes, would you know how to successfully get the parent object of the ControlExample in this case? Thanks

stmoy: Walking up the parent tree is generally not advised, but there is some sample code in the following issue that shows how to do it. [microsoft-ui-xaml #1331](https://github.com/microsoft/microsoft-ui-xaml/issues/1331)

Doing so involves using `currParent = VisualTreeHelper.GetParent(child);` and recursing up the tree.

XamlControlsGallery/ControlExample.xaml
salmanmkc: Agreed, it will add unnecessary overhead for the Controls that don't utilise it.

salmanmkc:
Okay, this is the current state:
image

salmanmkc:
My method of identifying the buttons may not be the most correct XAML way, do let me know how I can improve upon it, but it is more efficient thang going up the tree.

salmanmkc:
My method of identifying which example the MenuFlyoutButtons belong to may not be the most correct XAML way, do let me know how I can improve upon it, but it is more efficient thang going up the tree.

salmanmkc:
My method of identifying which example the MenuFlyoutButtons belongs to may not be the most correct XAML way, do let me know how I can improve upon it, but it is more efficient than going up the tree.

XamlControlsGallery/ControlPages/MenuBarPage.xaml.cs
```diff
@@ -26,5 +26,23 @@ public MenuBarPage()
...
+        {
+            var selectedFlyoutItem = sender as MenuFlyoutItem;
+            string exampleNumber = selectedFlyoutItem.Name.Substring(0, 1);
```
YuliKl: An alternative approach to using the first letter in the Name could be to use the Tag property to pass information about which TxtBlock to update.

XamlControlsGallery/ControlPages/MenuBarPage.xaml
```diff
@@ -12,124 +12,133 @@
...
-            <muxc:MenuBar>
-                <muxc:MenuBarItem Title="File">
-                    <MenuFlyoutItem Text="New"/>
-                    <MenuFlyoutItem Text="Open..."/>
-                    <MenuFlyoutItem Text="Save"/>
-                    <MenuFlyoutItem Text="Exit"/>
-                </muxc:MenuBarItem>
+            <StackPanel>
+                <TextBlock x:Name="SelectedOptionText" Text="" />
+                    <muxc:MenuBar>
```
YuliKl: nit: the amount by which this line is indented is incorrect.

XamlControlsGallery/ControlPages/MenuBarPage.xaml.cs
michael-hawker: 🦙 [Or using the Toolkit](https://docs.microsoft.com/en-us/windows/communitytoolkit/extensions/visualtree) 😋

XamlControlsGallery/ControlPages/MenuBarPage.xaml.cs
michael-hawker: Or set the Tag of the parent MenuBar and bind the TextBlock's text to that Tag. Then we wouldn't need to name everything.

MikeHillberg:
Thanks @salmanmkc!

salmanmkc:
> Thanks @salmanmkc!

My pleasure Mike! 😇

PR #264, opened 2019-11-12
updated image, since the subtitles are fixed

salmanmkc:
I can't upload the file, so just linked to it on Imgur

## Motivation and Context
The image that is shown has the descriptions showing for each control, rather than the subtitles

## How Has This Been Tested?
Making sure the link is accessible

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)


README.md
```diff
@@ -13,7 +13,7 @@ description: Demonstrates the usage of all XAML/WinUI controls in an interactive
 
 # XAML Controls Gallery
 
-![Xaml Gallery Screenshot](README_Images/XamlGalleryLanding.PNG)
+![Xaml Gallery Screenshot](https://i.imgur.com/W5dmkLl.png)
```
chingucoding: I think it would be better to replace the existing image with the new image if we want to change it instead of relying on imgur for this image. You should be able to overwrite the file in the folder and commit the new file together with this change.

stmoy: +1. It should be possible to update the image in the GitHub repo. We don't want to rely on imgur links, especially because we use the Gallery in situations where an internet connection isn't always guaranteed.

salmanmkc: ah okay, done that, thanks!

PR #271, opened 2019-11-19
Wording and format changes 

salmanmkc:
Just some small wording/format changes

## Description
- "This sample" -> "The XAML Controls Gallery"
- Made the note a blockquote
- Added a rule above the note, since the sections should be separated in my opinion
- Made the "The XAML Controls Gallery shows how to" line h2, I have tried h3, h4 and h5 as well, but h2 seems to be the most effective since this is an important part of the read me that should stand out

## Motivation and Context
Improve ReadMe

## How Has This Been Tested?
Checked preview with multiple possible changes

Before:
image

After (final):
image

## Types of changes


README.md
```diff
@@ -19,15 +19,16 @@ description: Demonstrates the usage of all XAML/WinUI controls in an interactive
...
 - **Basic layout:** This sample will show all of the possible layout options for your app and allow you to interact with the panels to show how to achieve any layout you are looking for.
 - **Adaptive UI:** In addition to showing how each control responds to different form factors, the app itself is responsive and shows various methods for achieving adaptive UI.
 - **Version adaptive code:** This sample shows how to write version adaptive code so that the app can run on previous versions of Windows while also using the latest capabilities on the most recent verison of Windows.
 
-**Note** The Xaml Controls Gallery requires Visual Studio [2017](http://go.microsoft.com/fwlink/?LinkID=532422) or [2019](https://visualstudio.microsoft.com/vs/) to build and Windows 10 to execute.
+---
+>**Note**: The Xaml Controls Gallery requires Visual Studio [2017](http://go.microsoft.com/fwlink/?LinkID=532422) or [2019](https://visualstudio.microsoft.com/vs/) to build and Windows 10 to execute.
```
chingucoding: In this case, I think the vertical line is enough to create a visual distinction between the sections. 😅

chingucoding:
We probably should also update this line from "this sample" to "this app":
https://github.com/microsoft/Xaml-Controls-Gallery/blob/24d343347920fcc300130f0ec1cdd6d31bf7a58b/README.md#L43

README.md
salmanmkc: The reason I use a blockquote is that the styling for notes that I've seen in other documentation seems to use block quotes as well (see https://github.com/MicrosoftDocs/WindowsCommunityToolkitDocs/blob/master/docs/controls/wpf-winforms/WebView.md). Is that what you're referring to, or are you suggesting a thinner line?

salmanmkc: I should probably update it to be `!NOTE` though instead of `Note:`

chingucoding: > I should probably update it to be `!NOTE` though instead of `Note:`

I think the !NOTE syntax is based on css classes, but on the link you shared, they don't seem to work correctly.



> The reason I use a blockquote is that the styling for notes that I've seen in other documentation seems to use block quotes as well (see https://github.com/MicrosoftDocs/WindowsCommunityToolkitDocs/blob/master/docs/controls/wpf-winforms/WebView.md). Is that what you're referring to, or are you suggesting a thinner line?

Oh yes, maybe a thinner line, but if other projects do that, than maybe we can leave it that way.

salmanmkc: > I think the !NOTE syntax is based on css classes, but on the link you shared, they don't seem to work correctly.
I have reached out to question why they have exclamation marks for each blockquote, will get back to this when I get a response back.

> Oh yes, maybe a thinner line, but if other projects do that, than maybe we can leave it that way.
I have added a commit for having another H2 instead, to have consistent styling, perhaps that looks better since the thinnest line seems to be quite thick anyway.

salmanmkc: > I think the !NOTE syntax is based on css classes, but on the link you shared, they don't seem to work correctly.

I have reached out to question why they have exclamation marks for each blockquote, will get back to this when I get a response back.

> Oh yes, maybe a thinner line, but if other projects do that, than maybe we can leave it that way.

I have added a commit for having another H2 instead, to have consistent styling, perhaps that looks better since the thinnest line seems to be quite thick anyway.

chingucoding: Looks great, but maybe the note should be at the top of the section, since it is very important.

salmanmkc: wasn't sure if it looked a bit off with the note at the top, after looking at it with fresh eyes, I think you're right, it's more important to have it at the top, and it does look okay there, even though the looks aren't the most important thing in this case

PR #273, opened 2019-11-21
ProgressBar update

YuliKl:
Updated the existing progress bar sample to use the new WinUI version of the control.

## Description

- Moved the app to latest WinUI preview version
- Changed the namespace used in progress bar sample
- Marked progress bar page as "preview" until we switch to a release version of WinUI

## Motivation and Context

New-to-WinUI control

## How Has This Been Tested?

manually verified

image

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ProgressBarPage.xaml
+++ b/XamlControlsGallery/ControlPages/ProgressBarPage.xaml
@@ -11,13 +11,16 @@
 -->
 <Page x:Class="AppUIBasics.ControlPages.ProgressBarPage"
         xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:local="using:AppUIBasics"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" 
+        xmlns:local="using:AppUIBasics"
+        xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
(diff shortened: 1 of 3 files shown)


chingucoding:
When this PR gets through, should we (or I) take a quick look if we can clean up some workarounds for issues that are now fixed with the prerelease of WinUI 2.3? 😅

stmoy:
Note that the ProgressBar test is now failing (in an expected way, I think). Will follow up with @hassanuz

stmoy:
@chingucoding

> When this PR gets through, should we (or I) take a quick look if we can clean up some workarounds for issues that are now fixed with the prerelease of WinUI 2.3? 😅

You know we won't say "no" to help 😍

chingucoding:
It seems like we won't be able to get rid of some bug fixes. Even the fix for untargeted TeachingTips does not work in XCG for some reason. Maybe we need to look into this or is this related to the fact that the TeachingTip is in the header?

 Only thing I would do is replace the number input TextBoxes on the "ImplicitTransition" and "GridView" page with the new NumberInput control, if thats fine.

mdtauk:
You could also remove the "Hold and Change Value" button

chingucoding:
> You could also remove the "Hold and Change Value" button

What exactly do you mean? 😅

mdtauk:
> > You could also remove the "Hold and Change Value" button
> 
> What exactly do you mean? 😅

image

If you are using a NumberBox, that has up and down Spin Buttons that allow changing the value

chingucoding:
> If you are using a NumberBox, that has up and down Spin Buttons that allow changing the value

Oh I see. My aim was to clean up some code in XCG, but since we use the same button in the ProgressBar sample, I will change that too.

PR #274, opened 2019-11-24
Switch to NumberBox for number input in pages

chingucoding:
## Description

Switched from using TextBox with eventlisteners and buttons to NumberBox for cleaner implementation of samples.
## Motivation and Context

Improves code readability and code complexity.
## How Has This Been Tested?

Tested manually by navigating through pages and testing all behaviors such as hitting enter, entering values, use included splitbuttons for changing of values and trying to enter wrong values.
## GridView page
image

## Implicit Transition page
image

## ProgressBar page
image

## Types of changes


XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -14,7 +14,9 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:AppUIBasics"
-    xmlns:local1="using:AppUIBasics.ControlPages">
+    xmlns:local1="using:AppUIBasics.ControlPages"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+    xmlns:contract7Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,7)">
```
YuliKl: I don't think you're using this anywhere, are you?

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -120,66 +122,31 @@
...
-                        <Grid.ColumnDefinitions>
-                            <ColumnDefinition Width="Auto"></ColumnDefinition>
-                            <ColumnDefinition Width="Auto"></ColumnDefinition>
-                            <ColumnDefinition Width="Auto"></ColumnDefinition>
-                        </Grid.ColumnDefinitions>
-                        <Grid.RowDefinitions>
-                            <RowDefinition Height="Auto"></RowDefinition>
-                            <RowDefinition Height="Auto"></RowDefinition>
-                        </Grid.RowDefinitions>
+                    <muxc:NumberBox x:Name="ColumnSpace" Minimum="0" Maximum="100" Value="5"
```
YuliKl: The NumberBoxes lost automation names (these had been set using the LabeledBy property), so now Narrator won't announce what they change. Can you add an AutomationProperties.Name to each one?

XamlControlsGallery/ControlPages/GridViewPage.xaml
```diff
@@ -120,66 +122,31 @@
...
+                                    Header="Space between columns"
+                                    Style="{ThemeResource DefaultNumberBoxStyle}" ValueChanged="NumberBox_ValueChanged">
+                    </muxc:NumberBox>
```
YuliKl: nit: Can you make the closing bracket part of the same line rather than separate? So `<muxc:NumberBox x:Name="ColumnSpace" ... ValueChanged="NumberBox_ValueChanged" />`

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -52,10 +38,7 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
             ContentGridView.ItemsSource = Items;
             StyledGrid.ItemsSource = Items;
 
-            ActualColSpace = 5;
-            ActualRowSpace = 5;
-            ActualMaxItems = 3;
-            MaxItems.Text = ActualMaxItems.ToString();
+            //MaxItems.Text = ActualMaxItems.ToString();
```
YuliKl: Is this no longer needed?

XamlControlsGallery/ControlPages/GridViewPage.xaml
YuliKl: Maybe that's not the right fix. Looking at TextBox, seems like Narrator would automatically get the UIA name from the Header property. This doesn't seem to be happening with NumberBox. @SavoySchuler, is this a known bug in the control?

XamlControlsGallery/ControlPages/ProgressBarPage.xaml
```diff
@@ -44,9 +44,10 @@
 
         <local:ControlExample HeaderText="A determinate progress bar.">
             <StackPanel x:Name="Control2" Orientation="Horizontal">
-                <muxc:ProgressBar Width="130" x:Name="ProgressBar2" />
+                <muxc:ProgressBar Width="130" x:Name="ProgressBar2" Value="{x:Bind ProgressValue.Value,Mode=OneWay}"/>
```
YuliKl:
```suggestion
                <muxc:ProgressBar Width="130" x:Name="ProgressBar2" Value="{x:Bind ProgressValue.Value, Mode=OneWay}"/>
```

XamlControlsGallery/ControlPages/ProgressBarPage.xaml
```diff
@@ -44,9 +44,10 @@
...
                 <TextBlock x:Name="Control2Output" Style="{ThemeResource OutputTextBlockStyle}" Width="60" TextAlignment="Center" />
-                <RepeatButton Content="Click and hold" Click="RepeatButton_Click" />
+                <TextBlock Text="Progress" VerticalAlignment="Center" Margin="0,0,10,0"/>
+                <muxc:NumberBox x:Name="ProgressValue" Minimum="0" Maximum="100" SpinButtonPlacementMode="Inline"/>
```
YuliKl: This does need the LabeledBy automation property

XamlControlsGallery/ControlPages/GridViewPage.xaml
chingucoding: Oh missed that 😅 
Your suggestion definitely looks cleaner!

XamlControlsGallery/ControlPages/GridViewPage.xaml
chingucoding: > The NumberBoxes lost automation names (these had been set using the LabeledBy property), so now Narrator won't announce what they change. Can you add an AutomationProperties.Name to each one?

Fixed with latest commit.

XamlControlsGallery/ControlPages/GridViewPage.xaml
chingucoding: Yes, removed that line. Thanks!

XamlControlsGallery/ControlPages/GridViewPage.xaml.cs
```diff
@@ -201,357 +182,41 @@ private void SelectionModeComboBox_SelectionChanged(object sender, SelectionChan
...
-            
-        }
-
-        private void StyledGrid_InitWrapGrid(object sender, RoutedEventArgs e)
-        {
-            // Update ItemsWrapGrid object created on page load by assigning it to StyledGrid's ItemWrapGrid
-            StyledGridIWG = sender as ItemsWrapGrid;
-
-            // Now we can change StyledGrid's MaximumRowsorColumns property within its ItemsPanel>ItemsPanelTemplate>ItemsWrapGrid.
-            StyledGridIWG.MaximumRowsOrColumns = ActualMaxItems;
```
stmoy: This is a ton of deleted code. Double checking that we don't need it. Why don't we need it?

stmoy: Reading a bit deeper: is it because a lot of the deleted code is just checking that the value that the user inputs is valid?

chingucoding: Yes, a lot of the deleted code was needed to parse numbers, ignore text and keep the number in a certain range. With the NumberBox control, all of that is not needed anymore.

stmoy:
@SavoySchuler & @chingucoding - it looks like several PRs are updating the version of the WinUI nuget package to different numbers, creating merge conflicts found here and in #279. Which version of WinUI should we use? I think we just need to pick one 😅 and I don't have a strong opinion either way.

In the future, we should ensure that updating the WinUI package dependency is its own change, and not bundled with feature PRs. Doing so will help avoid merge conflicts and help keep the branch history clean.

chingucoding:
> @SavoySchuler & @chingucoding - it looks like several PRs are updating the version of the WinUI nuget package to different numbers, creating merge conflicts found here and in #279. Which version of WinUI should we use? I think we just need to pick one 😅 and I don't have a strong opinion either way.

I think the best is to settle for the latest release (2.3.191129002), since it is the latest and it is not a prerelease. What are your thoughts @SavoySchuler ? 😃

> In the future, we should ensure that updating the WinUI package dependency is its own change, and not bundled with feature PRs. Doing so will help avoid merge conflicts and help keep the branch history clean.

Yes, that is a great idea! That would definitely prevent such merge conflicts from happening in the future.

chingucoding:
@stmoy Thank you for finalizing this PR and getting it up to date with master (and of course merging it in master)!

PR #276, opened 2019-11-25
Enable Debug builds to run side-by-side with store version

### LINKED ISSUE #275: Xaml Controls Gallery should use a different package name for Debug builds
**Describe the bug**
When making changes to XCG and pressing F5, VS complains because the store-deployed version of Xaml Controls Gallery is already installed and you have to manually uninstall it.

**Expected behavior**
It would be really nice if you could run the store-deployed and VS-deployed versions side-by side.

I'm thinking we maintain another appxmanifest where the package name and app title are like "Xaml Controls Gallery (Debug)" for the debug flavors of the app. Then store-deployed versions are still the same as before but we can run the debug one side-by-side.

jevansaks:
Fixes #275 

Added Debug.appxmanifest where the app's display name and package name are different ("XAML Controls Gallery (Debug)", and changed the code so that the title is pulled from the package configuration instead of hard-coded in XAML.

Also fixed an error I kept seeing where some TreeView page elements were using Name instead of x:Name and the compiler complained.

### DIFF
--- a/UITests/base.cs
+++ b/UITests/base.cs
@@ -25,7 +25,11 @@ namespace UITests
     public class Test_Base
     {
         private const string WindowsApplicationDriverUrl = "http://127.0.0.1:4723";
+#if DEBUG
+        private const string AppUIBasicAppId = "Microsoft.XAMLControlsGallery.Debug_8wekyb3d8bbwe!App";
+#else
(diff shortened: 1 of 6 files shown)


stmoy:
Thank you for doing this @jevansaks!

PR #288, opened 2019-12-08
Add AutomationProperties page

### LINKED ISSUE #199: Add Accessibility sample page
- [x] New sample
- [ ] Modification to existing sample

N.A.

**Describe the sample**

Since we have samples for various controls and also non controls e.g. RevealStyle which is not a "real" control, one might expect to be able to use the XCG to find out about ways to write accessible UIs. However in XCG there are no samples regarding accessibility.

Creating accessible Apps is not easy, and there are a few samples regarding this. However since this is one of the easiest (and IMO one of the best) resources for WinUI components, having no samples that include accessibility features could prevent one from thinking about accessibility. Having samples for the most common accessibility propererties such as AutomationProperties.Name or AutomationProperties.AccessibilityView would make creating accessible components easier.

**Additional context**

This is something also requested by the community (well sort of):
https://github.com/microsoft/microsoft-ui-xaml/issues/1164

chingucoding:
## Description

This PR will add a page containing samples around AutomationProperties, which are useful (and needed) to write accessible XAML.

## Motivation and Context

Closes #199 
## How Has This Been Tested?

image

## Types of changes

### DIFF
--- a/XamlControlsGallery/DataModel/ControlInfoData.json
+++ b/XamlControlsGallery/DataModel/ControlInfoData.json
@@ -626,6 +626,29 @@
       "Description": "",
       "Items": [
         {
+          "UniqueId": "AutomationProperties",
+          "Title": "AutomationProperties",
+          "Subtitle": "Attributes that help with writing accessible XAML.",
+          "ImagePath": "ms-appx:///Assets/Button.png",
+          "Description": "The AutomationProperties attributes provide functionality around accessible components.",
+          "Content": "<p>The aim of <b>AutomationProperties</b> is to enable components that are accessible to screen readers.",
+          "IsNew": true,
+          "Docs": [
+            {
+              "Title": "AutomationProperties - API",
+              "Uri": "https://docs.microsoft.com/uwp/api/windows.ui.xaml.automation.automationproperties"
+            },
+            {
+              "Title": "Accessibility Guidelines",
+              "Uri": "https://docs.microsoft.com/windows/uwp/design/accessibility/accessibility"
+            }
+          ],
+          "RelatedControls": [
+            "TextBlock",
+            "Image"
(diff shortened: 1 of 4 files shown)


chingucoding:
> The only thing we may want to change is the category where AutomationProperties show up. Right now this page is in Basic Input, which doesn't seem quite right. Maybe Status and Info or Styles would make a better home?

This was something I though about too. My though was that accessible UI should be a basic duty for developers and nothing that should be seen as "special" or optional. But since it is not an input, I think "Status and Info" or "Styles" are better. I would think, "Status and Info" might be the best choice. I can update the PR with the best fitting category if you want.

On a different thought, could we even create a new category "Accessibility" for more sample pages to come?

YuliKl:
> On a different thought, could we even create a new category "Accessibility" for more sample pages to come?

Yes, that's also a good option that would further elevate Accessibility and hopefully draw more attention to the topic. Having only one page in a category isn't great, but maybe that will motivate us to add more pages 😊

chingucoding:
For now I move it into the "Status and Info" for now. Having a category with only one element in it feels a bit weird, but if you think we should do this with this PR I would be happy to do that :)

XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -993,6 +993,29 @@
       "ImagePath": "",
       "Description": "",
       "Items": [
+        {
+          "UniqueId": "AutomationProperties",
+          "Title": "AutomationProperties",
+          "Subtitle": "Attributes that help with writing accessible XAML.",
+          "ImagePath": "ms-appx:///Assets/Button.png",
```
stmoy: Looking forward: we should probably get a more appropriate icon here.

PR #292, opened 2019-12-10
Disable AppWindow part of TabView sample on old versions of Win10

### LINKED ISSUE #282: Disable TabView tearout sample on Win10 versions < 18362
In the tab view sample with a second window, dragging and dropping a tab onto the title bar area crashes the entire app. Also, dragging and dropping a tab outside the window crashes the entire app. 

1. Launch the tab view sample
2. Scroll to the bottom and launch the full sample.
3. Try dragging a tab onto the title bar area or try to create a new window

**Expected behavior**

The app shouldn't crash; a new window should be created.

**Actual behavior**

The app crashes.

**Screenshots**

XxWHO2dmEW

**Version Info**
Gallery version (found on Settings page): 1.2.11.0

Windows 10 version:
- [x] October 2018 Update (17763)

Device form factor:
- [x] Desktop

stmoy:
## Description

Fixes #282 

## Motivation and Context

The complete TabView sample includes a complete sample that shows how to make TabView go into the titlebar, rearrange tabs, and pull tabs into their own window. The underlying tech for the last sample is AppWindow, which is only available on 18362+.

This change adds a warning to the sample and disables the TabView tear out behavior. 

## How Has This Been Tested?

Manual testing on a 17763 VM

image

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/TabViewPage.xaml
+++ b/XamlControlsGallery/ControlPages/TabViewPage.xaml
@@ -300,7 +300,7 @@
             </local:ControlExample.Xaml>
         </local:ControlExample>
 
-        <local:ControlExample HeaderText="Complete TabView windowing sample">
+        <local:ControlExample HeaderText="Complete TabView windowing sample" MinimumUniversalAPIContract="8">
             <local:ControlExample.Example>
                 <Button Content="Click here to launch the sample" Click="TabViewWindowingButton_Click"/>
             </local:ControlExample.Example>
--- a/XamlControlsGallery/DataModel/ControlInfoData.json
+++ b/XamlControlsGallery/DataModel/ControlInfoData.json
@@ -1515,6 +1515,10 @@
             {
               "Title": "Guidance",
               "Uri": "https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/tab-view"
+            },
+            {
+              "Title": "Show multiple views for an app",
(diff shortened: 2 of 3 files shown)


XamlControlsGallery/ControlPages/TabViewPage.xaml
```diff
@@ -300,7 +300,7 @@
             </local:ControlExample.Xaml>
         </local:ControlExample>
 
-        <local:ControlExample HeaderText="Complete TabView windowing sample">
+        <local:ControlExample HeaderText="Complete TabView windowing sample" MinimumUniversalAPIContract="8">
```
YuliKl: I'm confused. Doesn't specifying minimum contract version for the ControlExample hide the entire example for contracts  < 8? That's certainly not what your screenshot is showing, and I haven't tried this on an older build. So maybe my understanding of how MinimumUniversalAPIContract works is wrong.

But where is the error text shown in your screenshot?

stmoy: Specifying the minimum contract version doesn't hide/disable the control, no. For example, the editable ComboBox sample just displays as a regular ol' ComboBox (that isn't editable) on lower versions of the SDK. We deliberately decided not to fully hide samples even on lower versions of the SDK if there is reasonable fallback behavior. 

The error text in the screenshot above is the red text above the button which is provided by the MinumumUniversalAPIContract API. There is also a comment in the code.

PR #312, opened 2020-01-10
Acrylic Page: Changed the step amounts for sliders

salmanmkc:
## Description
The sliders show 3 decimal places when you drag the slider, but it increments in steps of 0.1. I have modified all the sliders on the page to increase in steps of 0.0001 instead and allowed this to reflect with the keyboard changes as well.

## Motivation and Context
I thought a more gradual change between the tint looked nicer than having such massive steps between each level.

## How Has This Been Tested?
Building

Before:
image

After (one example):
image

## Types of changes


salmanmkc:
> I think it is better to only display 3 digits and change the stepfrequency to 0.001.
> 
> Having 0.0001 as step frequency makes using them with keyboard unpleasant as it takes around 30 seconds to go from 0 to 1.

Good idea!

YuliKl:
@salmanmkc are you still interested in making the change to use increments of .001? Looks like the code in the PR is still showing 4 decimal places instead of the proposed 3.

salmanmkc:
> @salmanmkc are you still interested in making the change to use increments of .001? Looks like the code in the PR is still showing 4 decimal places instead of the proposed 3.

Hopefully that works now

PR #317, opened 2020-01-10
Patch - ListView Page: Changed color of message box for visibility

anawishnoff:
## Description
The message boxes in the Inverted ListView for Messaging or Data Logging sample were hard to read in both light and dark mode, so I changed the color to make them more distinct and readable. This was done after feedback from @pag3 that his screen made it really difficult to read the text in the message boxes with the current colors and contrast.

## How Has This Been Tested?

Tested locally on my machine and different monitors in both light and dark mode.
image
image

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ListViewPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/ListViewPage.xaml.cs
@@ -337,7 +337,7 @@ namespace AppUIBasics.ControlPages
             // If sent message, use light gray
             else if (MsgAlignment == HorizontalAlignment.Right)
             {
-                BgColor = (SolidColorBrush)Application.Current.Resources["SystemControlBackgroundChromeMediumBrush"];
+                BgColor = (SolidColorBrush)Application.Current.Resources["SystemControlErrorTextForegroundBrush"];
             }
         }
 


anawishnoff:
> Looks great! I am just a bit confused why in your screenshot the error color for dark mode is red and for me it is yellow:
> 
> image

@chingucoding 
Hmm... I'm not sure why this is happening either. I'm using the ErrorTextForegroundBrush which should be standard unless you happened to change it as you changed your accent color?

chingucoding:
> @chingucoding
Hmm... I'm not sure why this is happening either. I'm using the ErrorTextForegroundBrush which should be standard unless you happened to change it as you changed your accent color?

As far as I am aware, I did not change it. Is there even a way to change the value of the SystemErrorTextColor brush in Windows?

chingucoding:
Using my app, I have tested the color on a different PC than mine and it also reported a yellow color for the dark theme `SystemErrorTextColor` . Maybe this is something which changed with different Windows versions? Should I open an issue for this in the WinUI repo or the UWP Docs repo?

No matter what color is correct, I think it should not block this PR since either way the contrast is better with the SystemErrorTextColor.

PR #329, opened 2020-01-17
Fix crash when clearing Numberbox

### LINKED ISSUE #324: Application crashes when we click on any UI when AutomationProperties' "Font Size text box" is empty.
**Describe the bug**

2. Activate Automation Properties from home page.
3. Navigate to text box present beside Font Size ,With backspace clear the value and then press tab or click on any other UI.

**Expected behavior**

Should be able to move the focus out of the text box by tab (or) Should be able to activate any other UI by clicking.

**Actual behavior**

Application gets crashed.

**Screenshots**

image

**Version Info**

1.2.12.0

**Additional context**

Copied from [internal bug](https://dev.azure.com/microsoft/OS/_workitems/edit/24649879)

chingucoding:
## Description

Removed binding and handle value change manually
## Motivation and Context

Fixes #324 
## How Has This Been Tested?

Tested by reproducing steps in #324
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/AutomationPropertiesPage.xaml
+++ b/XamlControlsGallery/ControlPages/AutomationPropertiesPage.xaml
@@ -43,12 +43,12 @@
         <local:ControlExample HeaderText="Providing a label for another control">
             <local:ControlExample.Example>
                 <StackPanel>
-                    <TextBlock FontSize="{x:Bind FontSizeNumberBox.Value,Mode=OneWay}" TextWrapping="WrapWholeWords" Margin="0,8,0,8">Cat ipsum dolor sit amet, incidunt. Ratione. 
+                    <TextBlock x:Name="FontSizeChangingTextBlock" TextWrapping="WrapWholeWords" Margin="0,8,0,8">Cat ipsum dolor sit amet, incidunt. Ratione. 
                         Ipsa omnis dolor rem or vel. Ipsa elit quaerat sequi, nequeporro nisi qui. Vel autem. 
                         Laborum pariatur nemo for veritatis yet explicabo eaque or esse. Ipsum ullam omnis so sed nequeporro but magna.</TextBlock>
                     <StackPanel Orientation="Horizontal">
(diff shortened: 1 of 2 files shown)


stmoy:
@chingucoding - is this change to workaround a bug in the platform, or is this a permanent fix? If it's the former, will we need to revert this code once the platform is fixed?

chingucoding:
> @chingucoding - is this change to workaround a bug in the platform, or is this a permanent fix? If it's the former, will we need to revert this code once the platform is fixed?

I think whether this fix temporary or not depends on how NumberBox will change in the future. The problem is that currently the default value is not valid in our use case. Depending on how the behavior will change this may be permanent. If we are able to e.g. specify a default value, this fix would become obsolete.

There is also an issue over at WinUI regarding fallback value of NumberBox:
https://github.com/microsoft/microsoft-ui-xaml/issues/1708

stmoy:
Thanks for the context. 

> There is also an issue over at WinUI regarding fallback value of NumberBox:
microsoft/microsoft-ui-xaml#1708

I think I'm missing some context - what is the relationship between the fallback of NumberBox and this change? I'm specifically confused why you needed to write the `FontSizeNumberBox_ValueChanged` method -- I'd expect the control to handle this case and prevent invalid values?

chingucoding:
> I think I'm missing some context - what is the relationship between the fallback of NumberBox and this change? I'm specifically confused why you needed to write the FontSizeNumberBox_ValueChanged method -- I'd expect the control to handle this case and prevent invalid values?

When the value of the NumberBox gets cleared, and the NumberBox loses focus, it reports its value. HOWEVER since the value was cleared, for some reason, NumberBox does not enforce that the value is between minimum and maximum but instead falls back to 0. And since 0 is not a valid value for `FontSize` the app crashed.

> I'd expect the control to handle this case and prevent invalid values?

Thats what I thought too, but clearing seems to be a corner case for this.

stmoy:
Ah hah! I see. Thank you for explaining @chingucoding - this makes sense to me. Merging now!

chingucoding:
> Ah hah! I see. Thank you for explaining @chingucoding - this makes sense to me. Merging now!

Always happy to help  \\(@^0^@)/

Issue #333, opened 2020-01-23
Application crashes when Pencil tip is changed on InkToolbar page from circle to rectangle or vice versa.


YuliKl:
**Describe the bug**

**Steps to reproduce the bug**

1. Launch XAML Controls Gallery.
2. Activate InkToolbar from All Controls Page.
3. Click On Pencil button and draw something in drawing area.
4. Click on Rectangle radio button.

**Expected behavior**

Selection should change to respective choice.

**Actual behavior**

Application crashes.

**Screenshots**

**Version Info**

Gallery version (found on Settings page): 

1.2.12.0

**Additional context**

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/24662725)

chingucoding:
Forgot to write "Fixes #333" in #337 . This issue was fixed with #337

PR #337, opened 2020-01-24
Fix app crash when trying to set pencil to rectangle on inktoolbar page

chingucoding:
## Description

Apparently "Rectangle" is not a valid value for pencil's, thus simply changing the PenTip value, when the user selected pencil through the InkToolBar.
## Motivation and Context

#333
## How Has This Been Tested?

Tested by reproducing steps in issue.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/InkToolbarPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/InkToolbarPage.xaml.cs
@@ -1,4 +1,5 @@
-﻿using Windows.UI.Core;
+﻿using System;
+using Windows.UI.Core;
 using Windows.UI.Input.Inking;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/InkToolbarPage.xaml.cs
```diff
@@ -26,9 +26,12 @@ private void UpdatePen()
             {
                 var defaultAttributes = _inkPresenter.CopyDefaultDrawingAttributes();
 
-                defaultAttributes.PenTip = (bool)penTipShape.IsChecked ? PenTipShape.Circle : PenTipShape.Rectangle;
-
-                _inkPresenter.UpdateDefaultDrawingAttributes(defaultAttributes);
+                // If we are using a pencil, changing pentip is not allowed!
+                if(defaultAttributes.Kind != InkDrawingAttributesKind.Pencil)
```
stmoy: Should this instead be:
```suggestion
                if(defaultAttributes.Kind == InkDrawingAttributesKind.Default)
```

?

I think the enum only has two types so these are effectively the same, but the philosophical difference is about specificity. What is here says "Rectangle should work on everything but Pencil", whereas the above suggestion says "The only place Rectangle is guaranteed to work is on Default".

chingucoding: Oh yes, that makes more sense! Great catch!

Issue #331, opened 2020-01-22
Discussion: Squiggles, Rosyln Analyzers, and EditorConfig files


kmgallahan:
I'd like to open a discussion about squiggles & suggestions in the codebase, use of Roslyn analyzers, and EditorConfig files.

I'll admit, I'm a stickler for stamping out squiggles (and suggestions) in my own codebases. I even go further and use multiple analyzer packages to add more rules on top of the basic VS ones.

The WinUI repo has [contribution guidelines](https://github.com/microsoft/microsoft-ui-xaml/blob/master/CONTRIBUTING.md), which I'd presume also apply here. Those guidelines have a [code style and conventions section](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/code_style_and_conventions.md) which reference the [.NET Core team's C# coding style rules](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md), and the [.editorconfig file used for that codebase](https://github.com/microsoft/microsoft-ui-xaml/blob/master/.editorconfig).

This repo does not have a top level .editorconfig file, and I think that should be the first step in any process to get a wrangle on VS squiggles & suggestions.

Although the microsoft-ui-xaml / WinUI .editorconfig file could just be adopted, the [.NET Core version](https://github.com/dotnet/runtime/blob/master/.editorconfig) is also an option.

---

Including the EditorConfig file would just be the first step. I'd then like to slowly start working through some rules to resolve inconsistences in the codebase (without stepping on any toes of course), while also tweaking the config where necessary.

Later, I'd potentially like to use this process with other MS WinUI-related open source codebases, such as:

* Windows Community Toolkit
* Windows Composition Samples
* Fluent XAML Theme Editor

I wanted to start here though because I feel the XAML Controls Gallery is a great example of a semi-complex app with a codebase many newcomers likely dip their toes in to see how to get things done. It is also less sensitive to PR merging than the WinUI repo.

Fewer squiggles and less inconsistencies in an official codebase that gets a lot of eyes would then help encourage proper use of analyzers and EditorConfig files elsewhere.

Let me know your thoughts.

chingucoding:
While I agree in your opinion that we should clear the projects from warnings and squiggles, I do not think that this will be a fun job.

[...]

PR #343, opened 2020-02-05
Adds a minimal EditorConfig and resolves a rule violation.

kmgallahan:
Adds a minimal EditorConfig file to the repo and resolves a rule violation. 

## Description
Adds the minimal EditorConfig file defined [here](https://github.com/kmgallahan/Style-as-Code/blob/master/editorconfig/.editorconfig_minimal).

Resolves 1 violation of rule IDE0016 where the recommendation is to:

Use throw expressions to simplify null checks when possible.

See the rationale for this rule here:

https://styleascode.net/ID/IDE0016

## Motivation and Context
This is the first step in taking action on #331.

See [this comment](https://github.com/microsoft/Xaml-Controls-Gallery/issues/331#issuecomment-582246759) for additional information and feedback.

For [added clarity](https://github.com/microsoft/Xaml-Controls-Gallery/issues/331#issuecomment-582528848), the plan would be to:

1. Submit the minimal EditorConfig
2. Agree on starting point and resolve some violations
3. Move to non-minimal EditorConfig (very few differences for now)
4. Begin adding additional rules 1-2 at a time with proper justification (minimal config is no longer touched at this point)

## How Has This Been Tested?
UI tests not run.

## Types of changes
- [X] Code style

### DIFF
--- a/XamlControlsGallery/.editorconfig
+++ b/XamlControlsGallery/.editorconfig
@@ -10,7 +10,7 @@ indent_style = space
 
 # Code files
 [*.{cs,csx,xaml,xml}]
-charset = utf-8-bom
+charset = utf-8
 indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
--- a/XamlControlsGallery/.editorconfig
+++ b/XamlControlsGallery/.editorconfig
@@ -1,0 +1,109 @@
+﻿# EditorConfig  https://EditorConfig.org
+# Style as Code Minimal EditorConfig v.2020.02.04.0 https://styleascode.net
+
+# top-most EditorConfig file
+root = true
+
+# Uses spaces for indentation.
+[*]
+indent_style = space
+
+# Code files
+[*.{cs,csx}]
+indent_size = 4
+insert_final_newline = true
+
+# C# files
+[*.cs]
+
+########################
+# Formatting
+########################
+
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = true
+csharp_indent_switch_labels = true
+
+# New line preferences
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# Space preferences
+csharp_space_after_cast = false
+csharp_space_after_colon_in_inheritance_clause = true
+csharp_space_after_comma = true
+csharp_space_after_dot = false
+csharp_space_after_keywords_in_control_flow_statements = true
+csharp_space_after_semicolon_in_for_statement = true
+csharp_space_around_binary_operators = before_and_after
+csharp_space_around_declaration_statements = do_not_ignore
+csharp_space_before_colon_in_inheritance_clause = true
+csharp_space_before_comma = false
+csharp_space_before_dot = false
+csharp_space_before_open_square_brackets = false
+csharp_space_before_semicolon_in_for_statement = false
+csharp_space_between_empty_square_brackets = false
+csharp_space_between_method_call_empty_parameter_list_parentheses = false
+csharp_space_between_method_call_name_and_opening_parenthesis = false
+csharp_space_between_method_call_parameter_list_parentheses = false
+csharp_space_between_method_declaration_empty_parameter_list_parentheses = false
+csharp_space_between_method_declaration_name_and_open_parenthesis = false
+csharp_space_between_method_declaration_parameter_list_parentheses = false
+csharp_space_between_parentheses = false
+csharp_space_between_square_brackets = false
+
+# Sort using and Import directives with System.* appearing first
+dotnet_sort_system_directives_first = true
+
+########################
+# Naming
+########################
+
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+dotnet_naming_symbols.static_fields.required_modifiers = static
+
+########################
+# Language Features
+########################
+
+# IDE1005: Delegate invocation can be simplified
+csharp_style_conditional_delegate_call = true:suggestion
+
+# IDE0018: Variable declaration can be inlined
+csharp_style_inlined_variable_declaration = true:suggestion
+
+# IDE0019: Use pattern matching
+csharp_style_pattern_matching_over_as_with_null_check = true:suggestion
+
+# 
+csharp_style_pattern_matching_over_is_with_cast_check = true:suggestion
+
+# IDE0016: Null check can be simplified
+csharp_style_throw_expression = true:suggestion
+
+# 
+dotnet_style_coalesce_expression = true:suggestion
+
+# IDE0028: Collection initialization can be simplified
+dotnet_style_collection_initializer = true:suggestion
+
+#
+dotnet_style_explicit_tuple_names = true:suggestion
+
+# 
+dotnet_style_null_propagation = true:suggestion
+
+# IDE0017: Object initialization can be simplified
+dotnet_style_object_initializer = true:suggestion
+
+# IDE0049: Name can be simplified
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
--- a/XamlControlsGallery/.editorconfig
+++ b/XamlControlsGallery/.editorconfig
@@ -1,17 +1,19 @@
-﻿# EditorConfig  https://EditorConfig.org
-# Style as Code Minimal EditorConfig v.2020.02.04.0 https://styleascode.net
+# EditorConfig  https://EditorConfig.org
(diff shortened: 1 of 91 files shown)


XamlControlsGallery/XamlControlsGallery.sln
```diff
@@ -1,10 +1,15 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2027
+# Visual Studio Version 16
```
chingucoding: Why are we excluding Visual Studio 2017 here?

XamlControlsGallery/.editorconfig
```diff
@@ -0,0 +1,109 @@
...
+
+# top-most EditorConfig file
+root = true
+
+# Uses spaces for indentation.
+[*]
+indent_style = space
+
+# Code files
+[*.{cs,csx}]
```
chingucoding: What about xaml files? 😅

XamlControlsGallery/.editorconfig
```diff
@@ -0,0 +1,109 @@
+﻿# EditorConfig  https://EditorConfig.org
```
chingucoding: Is there a editor config property regarding point 4 of [this document](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md):

> We avoid `this.` unless absolutely necessary.

XamlControlsGallery/.editorconfig
kmgallahan: XAML files can't be analyzed via Roslyn Analyzers, so they don't have any rules that can appear in EditorConfig files.

There is a XAML Analysis package, but it is in beta, can't be configured, and has some issues from when I tried it out.

chingucoding: What about indent size and such?

XamlControlsGallery/.editorconfig
kmgallahan: Yes, however for right now I just wanted to rely on the minimal EditorConfig which includes only the rules that both the Roslyn compiler and .NET Core runtime teams agree on.

They don't agree on how to enforce the use of `.this`, so it isn't in the config for now. I added that part to the best practices before I did the study comparing both configs and making the minimal config.

I plan to work through the rules in this config for now, and add additional ones over time - such as addressing this. But I feel commits adding new (potentially controversial) rules should have their own PRs.

See [this](https://styleascode.net/EditorConfigs/selection.html) regarding the config options I went though, and [this](https://styleascode.net/EditorConfigs/style_as_code.html) for the configs I plan to make available over time.

XamlControlsGallery/XamlControlsGallery.sln
kmgallahan: I don't believe that excludes earlier versions, as that is set by `MinimumVisualStudioVersion = 10.0.40219.1`.

My VS just modified those lines automatically when the `.editorconfig` was added to the solution.

XamlControlsGallery/.editorconfig
kmgallahan: Those can be handled, but I was really just trying to tackle C# rules for now in the theme of the minimal editorconfig. Example from microsoft-ui-xaml:

```
[**.{xaml,xml}]
indent_style = space
indent_size = 4
charset = utf-8-bom
insert_final_newline = true
trim_trailing_whitespace = true
end_of_line = crlf
```

chingucoding: Since you already have a small snippet for .cs files, adding a few lines for XAML wouldn't hurt ^^

XamlControlsGallery/.editorconfig
chingucoding: Hmm I see. I guess in that case, it is best to leave the `this.` controversy out (though I think in WinUI this is rarely used, if used at all).

XamlControlsGallery/XamlControlsGallery.sln
chingucoding: I guess if they don't mean anything (from what I could find online), we can leave it that way.

XamlControlsGallery/.editorconfig
```diff
@@ -0,0 +1,109 @@
...
+# Style as Code Minimal EditorConfig v.2020.02.04.0 https://styleascode.net
+
+# top-most EditorConfig file
+root = true
+
+# Uses spaces for indentation.
+[*]
+indent_style = space
```
chingucoding: We should use tabs since they are more accessible than spaces for indentation.

kmgallahan: Compiler and runtime teams agree that 4 spaces should be used.

chingucoding: Oh okay, didn't knew that. Sorry about my comment then.

kmgallahan: As mentioned, the rules in this config are only those already found both the EditorConfig [used by the runtime](https://github.com/dotnet/runtime/blob/master/.editorconfig), and that [used by the compiler](https://github.com/dotnet/roslyn/blob/master/.editorconfig), and that match exactly in how they are enforced.

None of these rules involve my own opinions.

XamlControlsGallery/XamlControlsGallery.sln
kmgallahan: I actually tried manually changing them back, but then VS fought against me and won by flipping them again before the commit was made...

XamlControlsGallery/.editorconfig
kmgallahan: Bumped the config version with a few changes to include XAML & XML, character encoding, whitespace trimming, and removal of a rule requires another to be present (it isn't).

kmgallahan:
@ranjeshj Have you taken a look at this PR yet?

ranjeshj:
> @ranjeshj Have you taken a look at this PR yet?

Thanks for pinging. I somehow missed this notification. Looks like a good change to have.

XamlControlsGallery/Common/ThemeHelper.cs
```diff
@@ -13,7 +13,7 @@ public static class ThemeHelper
     {
         private const string SelectedAppThemeKey = "SelectedAppTheme";
         private static Window CurrentApplicationWindow;
-        // Keep reference so it does not optimized/gc'ed away
+        // Keep reference so it does not get optimized/garbage collected
```
chingucoding: nit²: I think "optimized" may be a bit confusing, and "optimized away" may be a bit more descriptive. However either way it's clear so I guess I am a bit nit picky here.

XamlControlsGallery/.editorconfig
```diff
@@ -0,0 +1,110 @@
...
+# top-most EditorConfig file
+root = true
+
+# Use spaces for indentation.
+[*]
+indent_style = space
+
+# Code files
+[*.{cs,csx,xaml,xml}]
+charset = utf-8-bom
```
chingucoding: Is there a specific reason for using BOM with UTF-8? The Unicode Standard is discouraging the use of BOM with UTF-8 so why not stick with simple UTF-8?

kmgallahan:
I didn't realize this PR was still open and my changes were still going into it... oops. Was experimenting with rewriting history too.

XamlControlsGallery/Common/ThemeHelper.cs
kmgallahan: Only made this change as my spellchecker picked up "gc'ed". Not particularly important.

XamlControlsGallery/.editorconfig
kmgallahan: I included -bom because it is present in the microsoft-ui-xaml repo's EditorConfig.

https://github.com/microsoft/microsoft-ui-xaml/blob/master/.editorconfig#L123

chingucoding: I think changing it on the WinUI repro isn't worth the effort, but if you/we are creating a new file here, I don't think we should just inherit discouraged behaviors from other repos. 

But well that's just my opinion.

kmgallahan: I've removed it, seems like the better option.

XamlControlsGallery/ControlPages/FlipViewPage.xaml
```diff
@@ -32,7 +32,7 @@
 &lt;FlipView MaxWidth="400" Height="270"&gt;
     &lt;Image Source="ms-appx:///Assets/SampleMedia/cliff.jpg" AutomationProperties.Name="Cliff"/&gt;
     &lt;Image Source="ms-appx:///Assets/SampleMedia/grapes.jpg" AutomationProperties.Name="Grapes"/&gt;
-    &lt;Image Source="ms-appx:///Assets/SampleMedia/rainer.jpg" AutomationProperties.Name="Rainier"/&gt;
+    &lt;Image Source="ms-appx:///Assets/SampleMedia/rainier.jpg" AutomationProperties.Name="Rainier"/&gt;
```
stmoy: Doesn't this break the FlipView page?

kmgallahan: The change was to the sample code. The code that gets executed had the correct spelling already.

kmgallahan:
@stmoy I replied to your comment about the FlipView. Anything else I can do to get this PR merged and closed?

Again, apologies for commits 4-7, didn't realize my force-pushes were going into the remote I had configured earlier.

stmoy:
Thanks for the ping - love this work!

PR #359, opened 2020-02-21
Resolves 13 code style rule violations.

kmgallahan:
This PR resolves all violations of the following rules:

- IDE0018: Variable declaration can be inlined
- IDE1005: Delegate invocation can be simplified

### IDE0018: Variable declaration can be inlined

Recommendation:

> `out` variables should be declared inline when possible.

Rationale: https://styleascode.net/ID/IDE0018

Instances resolved: 6

---

### IDE1005: Delegate invocation can be simplified

Recommendation:

> The null-conditional operator should be used to simplify delegate invocation when possible.

Rationale: https://styleascode.net/ID/IDE1005

Instances resolved: 7

## Motivation and Context
See #331.

## How Has This Been Tested?
Tests were not run. Solution does compile & run without errors.

### DIFF
--- a/XamlControlsGallery/.editorconfig
+++ b/XamlControlsGallery/.editorconfig
@@ -1,5 +1,5 @@
 # EditorConfig  https://EditorConfig.org
-# Style as Code Minimal EditorConfig v.2020.02.06.0 https://styleascode.net
+# Style as Code Minimal EditorConfig v.2020.02.16.0 https://styleascode.net
 
 # top-most EditorConfig file
 root = true
--- a/XamlControlsGallery/Common/NavigationHelper.cs
+++ b/XamlControlsGallery/Common/NavigationHelper.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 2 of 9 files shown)


kmgallahan:
So I'll make this the first PR dedicated to resolving code style rule violations.

@stmoy @chingucoding 

- How's the formatting/content of the PR text? Suggestions?
- As I wrote about [here](https://styleascode.net/Resolving_Violations/recommended_process.html), I'm trying to limit PRs to cover only 1-3 rule violations, depending on how many changes were made / files were touched. Any input on what these limits should be? 
- Preference for larger or smaller PRs to reduce time spent reviewing & merging them?

stmoy:
Thank you again for taking the initiative on this project - I love where we're headed.

> How's the formatting/content of the PR text? Suggestions?

I like the formatting that includes a justification for the rules and a link. The only small feedback is that the different sections inside the Description would be clearer if the specific rule was called out as a header rather than only in the body.

Something like this:

> **IDE0018**
>> Resolves 6 violations of rule IDE0018 where the recommendation is:
>> out variables should be declared inline when possible.
>> See the rationale for this rule here:
>> https://styleascode.net/ID/IDE0018

Thoughts?

> As I wrote about here, I'm trying to limit PRs to cover only 1-3 rule violations, depending on how many changes were made / files were touched. Any input on what these limits should be?

+1 to limiting PRs to smaller chunks. The size in this PR is perfect, imo.

> Preference for larger or smaller PRs to reduce time spent reviewing & merging them?

Smaller > larger.

chingucoding:
I totally agree with @stmoy . Thanks for doing this @kmgallahan , this helps a lot!

Just out of curiosity, how many PR's do you think would it take to clean everything up?

kmgallahan:
@stmoy 

> the different sections inside the Description would be clearer if the specific rule was called out as a header rather than only in the body.

Ok, I switched the text around a bit.

@chingucoding 

> Just out of curiosity, how many PR's do you think would it take to clean everything up?

That depends on how forward-looking you want to be.

### Short-term

For reference, the "minimal" EditorConfig (where there is general agreement on how the rules should be configured) includes:

- **11** language feature rules (already worked through 3)
- **1** naming rule
- **38** formatting rules

I'm not quite sure what the best method will be for working through the formatting rules, as they are focused on simple things like spaces, new lines, and indents. They could just be handled with 1 PR for each of those 3 subjects, but I'm not really sure how many changed LOC / PR that would result in.

That leaves 9 remaining rules, with 1 rule per commit, and 1-3 commits per PR.

### Medium-term

After the minimal EditorConfig has been worked through there are **99** more rules that appear in the Roslyn & .NET Core runtime configs (haven't divided those into formatting/etc. yet).

I don't know how many, if any, other built-in IDExxxx rules there are outside of those.

### Long-term

Beyond the built-in rules, there are other analyzers packages that can be considered down the road as well (like the [MS recommended FxCop Analyzers](https://github.com/dotnet/roslyn-analyzers)).

Issue #363, opened 2020-02-25
Work Tracking: Solution-wide code style application


kmgallahan:
This issue exists to track work items related to solution-wide code style application.

## 1. Adopt a minimal EditorConfig

- [x] Make the decision to start implementing solution-wide code style rules
- [x] Agree on rules to include in the minimal EditorConfig
- [x] Add the minimal EditorConfig to the solution

### Formatting

- [ ] Decide on best approach for resolving the 38 formatting rules (commit/PR structure)

### Naming

- [ ] dotnet_naming_symbols.static_fields.required_modifiers

### Language Features

- [x] IDE0016: Use throw expressions to simplify null checks
- [x] IDE0017: Use object initializers
- [x] IDE0018: Variable declaration can be inlined
- [x] IDE0019: Use pattern matching over as with null check
- [x] IDE0028: Use collection initializers
- [x] IDE0049: Use type keywords
- [x] IDE1005: Delegate invocation can be simplified
- [x] IDExxxx: dotnet_style_explicit_tuple_names
- [x] IDExxxx: dotnet_style_null_propagation
- [x] IDExxxx: dotnet_style_coalesce_expression
- [x] IDExxxx: csharp_style_pattern_matching_over_is_with_cast_check

## 3. Switch to an EditorConfig that will eventually address all built-in rules

- [ ] Agree on rules to include in the initial version of this EditorConfig (note: this is not meant to be the final/complete ruleset, rules will rather be added over time as decisions are made on how to configure them)
- [ ] Swap the old EditorConfig with the new version

## 4. Resolve all built-in rule violations

chingucoding:
@kmgallahan I assume you are not looking to continue working on this issue right? Would you be fine if this issue was closed than?

kmgallahan:
@chingucoding

I started this effort because:

* I love using code analyzers and editorconfig files to standardize and improve my code
* I hate glancing at VS Error List windows and seeing dozens to hundreds of ignored warnings and messages

The idea was to create a somewhat standardized editorconfig that could be used for a few public-facing MS projects, and perhaps community projects as well.

However, the large number of analyzers combined with the friction of implementing changes made this painful, and that was just with the WinUI 2 Gallery 😅.

What are your thoughts on the subject?

chingucoding:
I totally agree how great this can be, especially being able to see the list of errors, warnings and notifications going done. 

[...]

PR #366, opened 2020-02-29
Resolves 25 code style rule violations.

kmgallahan:
This PR resolves all violations of the following rules:

- IDE0017: Use object initializers
- IDE0028: Use collection initializers

Two new violations of rules IDE0018 & IDE1005 were resolved.

Two spelling mistake regressions were also reapplied.

### IDE0017: Use object initializers

Recommendation:

> Both object and collection initializers should be used, regardless of how many elements are being set.

Rationale: https://styleascode.net/ID/IDE0017

Instances resolved: 20

### IDE0028: Use collection initializers

Recommendation:

> Both object and collection initializers should be used, regardless of how many elements are being set.

Rationale: https://styleascode.net/ID/IDE0028

Instances resolved: 3

## Motivation and Context
See #331 & #363.

## How Has This Been Tested?
Tests were not run. Solution does compile & run without errors.

## Types of changes
- [X] Code quality

### DIFF
--- a/XamlControlsGallery/Common/VariedImageSizeLayout.cs
+++ b/XamlControlsGallery/Common/VariedImageSizeLayout.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Collections.Generic;
@@ -108,8 +108,7 @@ namespace AppUIBasics.Common
 
             for (int index = 0; index < m_cachedBounds.Count; index++)
             {
-                double nextOffset = 0.0;
-                int columnIndex = GetIndexOfLowestColumn(m_columnOffsets, out nextOffset);
+                int columnIndex = GetIndexOfLowestColumn(m_columnOffsets, out var nextOffset);
                 var oldHeight = m_cachedBounds[index].Height;
                 m_cachedBounds[index] = new Rect(columnIndex * Width, nextOffset, Width, oldHeight);
                 m_columnOffsets[columnIndex] += oldHeight;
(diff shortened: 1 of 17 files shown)


kmgallahan:
Although the value of initializers goes down when dealing with 1-2 elements, the only options are:

A. Ignore the rule
B. Use suppressions
C. Change the rule configuration to `none`
D. Suggest the analyzers be modified to take a minimum element count & wait
E. Apply the rule as-is

A - goes against the goal of getting to zero warnings and suggestions. 

B - is undesirable as in-code suppressions get messy. 

C - goes against the Roslyn and .NET Core runtime teams code style guidelines

D - I'm not sure if the built-in analyzers are even open for changes, let alone the type that would allow this level of customization

So, we're left with E. Not a terrible thing as initializers do have positives, even for only 1-2 elements.

---

These are the first rules where resolving all violations didn't feel perfect. There will surely be many more down the road.

However, all rules outside of those in the minimal EditorConfig are more open to interpretation, as there is less agreement on their usage.

kmgallahan:
@chincucoding

> To be honest, I don't really understand why the value should go down with increasing number of properties. 

I said the value goes down when there are only 1-2. Comparing:

```cs
List<Cat> cats = new List<Cat>();
cats.Add(new Cat{ Name = "Sylvester", Age=8 });
```
with
```cs
List<Cat> cats = new List<Cat>
{
    new Cat{ Name = "Sylvester", Age=8 }
}
```

it's less compact - instead of taking 2 lines you're using 4, plus indentation is introduced. So readability goes down a touch IMO.

However, you are still getting the other [benefits](https://styleascode.net/Best_Practices/Language_Features/object_and_collection_initializers.html#arguments): 

✔ Prevents objects and collections from being accessed before all necessary elements added / set.

✔ Helps to ensure lines aren’t misplaced during refactoring.

✔ Eliminates repetitive identifier usage (e.g. cat. and cats.Add).

chingucoding:
Oh, misread your comment. Yeah, for 1-2 items this is increasing the number of lines for no real reason. I guess if we want to be consistent, we have to accept this.

PR #382, opened 2020-03-13
Rename buttons to be more accessible for people using narrator

### LINKED ISSUE #380: Button name and control type should not be same.
**Describe the bug**

1. Launch XAML Controls Gallery and turn on Narrator
2. Activate "Button" tab item from home page.
3. Navigate to "button" under "a simple button...." heading and "button" under "Reveal style....." heading.
4. Observe narrator's description for the same.

**Expected behavior**

Button name and control type should not be same.

**Actual behavior**

Narrator only announces "button" in both the cases.

**Screenshots**

image

image

**Version Info**

1.2.13.0

**Additional context**

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/25441100)

chingucoding:
## Description

Changed content of button samples so that they are not announced as "Button, Button".
## Motivation and Context

Closes #380 
## How Has This Been Tested?

Tested using narrator.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ButtonPage.xaml
+++ b/XamlControlsGallery/ControlPages/ButtonPage.xaml
@@ -19,7 +19,7 @@
         <local:ControlExample x:Name="Example1" HeaderText="A simple Button with text content.">
             <local:ControlExample.Example>
                 <RelativePanel>
-                    <Button x:Name="Button1" Content="Normal button" Click="Button_Click"
+                    <Button x:Name="Button1" Content="Standard XAML button" Click="Button_Click"
                             IsEnabled="{x:Bind DisableButton1.IsChecked.Value.Equals(x:False), Mode=OneWay}" />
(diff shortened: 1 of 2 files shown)


XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -19,7 +19,7 @@
         <local:ControlExample x:Name="Example1" HeaderText="A simple Button with text content.">
             <local:ControlExample.Example>
                 <RelativePanel>
-                    <Button x:Name="Button1" Content="Button" Click="Button_Click"
+                    <Button x:Name="Button1" Content="Normal button" Click="Button_Click"
```
stmoy: What about "Standard Xaml Button", "Plain Button", or "Example Button" instead? "Normal" doesn't seem quite right in the context of a UI control. 

CC: @YuliKl

chingucoding: Good idea, changed the names a bit so the meaning is clearer.

PR #392, opened 2020-03-28
Add source code links to item pages

### LINKED ISSUE #391: Sample pages should include link to sample code
The sample pages have markup, and have a link to the documentation. But some of the the samples include code that's not shown on the page nor linked to. It would be useful if the page also had a link to the sample page implementation in this repo.

For example the AutoSuggestBox page has a SuggestionChosen event handler that's not shown, but the page could have a link directly to the [markup](https://github.com/microsoft/Xaml-Controls-Gallery/blob/98e09620cd4182ef5e865c091788fb6ea3ba85d0/XamlControlsGallery/ControlPages/AppBarToggleButtonPage.xaml) and [code](https://github.com/microsoft/Xaml-Controls-Gallery/blob/98e09620cd4182ef5e865c091788fb6ea3ba85d0/XamlControlsGallery/ControlPages/AutoSuggestBoxPage.xaml.cs).

chingucoding:
## Description

Using the page type name, we generate the GitHub source code link and display a link to the C# code and the XAML code.
## Motivation and Context

Closes #391 
## How Has This Been Tested?

Tested by viewing links of some pages (all worked)
image

## Types of changes

@MikeHillberg I hope that is what you had in mind. Otherwise please let me know what can be better.

### DIFF
--- a/XamlControlsGallery/ItemPage.xaml
+++ b/XamlControlsGallery/ItemPage.xaml
@@ -71,7 +71,9 @@
                         <Setter Target="seeAlsoPanel.(Grid.Row)" Value="3" />
                         <Setter Target="seeAlsoPanel.(Grid.Column)" Value="0" />
                         <Setter Target="seeAlsoPanel.(Grid.ColumnSpan)" Value="3" />
-                        
+
+                        <Setter Target="SourcePanel.(Grid.ColumnSpan)" Value="1" />
+
                         <Setter Target="DocumentationPanel.(Grid.ColumnSpan)" Value="1" />
 
                         <Setter Target="RelatedControlsPanel.(Grid.Row)" Value="0" />
@@ -80,7 +82,7 @@
                         <Setter Target="RelatedControlsPanel.(Grid.ColumnSpan)" Value="1" />
 
                         <Setter Target="FeedbackPanel.(Grid.Row)" Value="1" />
-                        <Setter Target="FeedbackPanel.(Grid.Column)" Value="0" />
+                        <Setter Target="FeedbackPanel.(Grid.Column)" Value="2" />
                         <Setter Target="FeedbackPanel.(Grid.ColumnSpan)" Value="1" />
                     </VisualState.Setters>
                 </VisualState>
@@ -94,6 +96,8 @@
                         <Setter Target="seeAlsoPanel.(Grid.Column)" Value="0" />
                         <Setter Target="seeAlsoPanel.(Grid.ColumnSpan)" Value="3" />
 
+                        <Setter Target="SourcePanel.(Grid.ColumnSpan)" Value="1" />
+
                         <Setter Target="DocumentationPanel.(Grid.ColumnSpan)" Value="1" />
 
                         <Setter Target="RelatedControlsPanel.(Grid.Row)" Value="0" />
@@ -102,7 +106,7 @@
                         <Setter Target="RelatedControlsPanel.(Grid.ColumnSpan)" Value="1" />
 
                         <Setter Target="FeedbackPanel.(Grid.Row)" Value="1" />
-                        <Setter Target="FeedbackPanel.(Grid.Column)" Value="0" />
+                        <Setter Target="FeedbackPanel.(Grid.Column)" Value="2" />
                         <Setter Target="FeedbackPanel.(Grid.ColumnSpan)" Value="1" />
 
                         <Setter Target="rightMargin.Width" Value="0" />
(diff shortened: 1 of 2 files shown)


MikeHillberg:
This looks great!

Two small things
* I think the app generally uses the term "sample", so I would make the title "Sample source code"
* In the prototype picture here, it looks like Feedback and Related Controls are rendering to the same place

Also, does every sample have both Xaml and C#?

chingucoding:
Thanks for your feedback, I've updated the header and layout.

Almost every sample has C# code, and every sample has XAML.

stmoy:
Thanks for building this @chingucoding! 

> Also, does every sample have both Xaml and C#?
>> Almost every sample has C# code, and every sample has XAML.

By "sample", are you referring to the sample page itself or individual code samples? ie. there is one TabView sample page with _n_ different samples for each feature. 

If the former, then there is always both a xaml and c# page (TabView.xaml/.cs). If the latter, the statement that every sample has XAML isn't strictly true. For example, the Connected Animation pages only have C#. The ColorPaletteResources page doesn't have xaml OR c#. 

It looks like this code adds links to the sample pages themselves, though, so I think this change is doing the right thing regardless.

XamlControlsGallery/ItemPage.xaml
```diff
@@ -94,6 +96,8 @@
                         <Setter Target="seeAlsoPanel.(Grid.Column)" Value="0" />
                         <Setter Target="seeAlsoPanel.(Grid.ColumnSpan)" Value="3" />
 
+                        <Setter Target="SourcePanel.(Grid.ColumnSpan)" Value="1" />
```
stmoy: Is this not the default? Why does this need to be set?

XamlControlsGallery/ItemPage.xaml
```diff
@@ -151,14 +155,35 @@
...
+                            Style="{ThemeResource SubtitleTextBlockStyle}"
+                            Text="Sample source code"/>
+                        <HyperlinkButton
+                            x:Name="PageCodeGitHubLink"
+                            ToolTipService.ToolTip="{x:Bind PageCodeGitHubLink.NavigateUri, Mode=OneWay}"
+                            Margin="0,12,0,0">
+                            <TextBlock Text="C# code on GitHub"/>
+                        </HyperlinkButton>
+
+                        <HyperlinkButton
```
stmoy: I think the Xaml samples are generally more common/interesting - can you please move this above the C# link?

chingucoding:
Oh right completely forgot that those existed too. I think by linking to the sample page, we indeed end up solving the problem at hand in the best way.

XamlControlsGallery/ItemPage.xaml
```diff
@@ -151,14 +155,35 @@
...
                         <ColumnDefinition />
                     </Grid.ColumnDefinitions>
 
-                    <StackPanel x:Name="DocumentationPanel" Grid.ColumnSpan="3" Margin="0,0,0,40">
+                    <StackPanel x:Name="SourcePanel" Grid.Row="0" Grid.ColumnSpan="3" Margin="0,0,0,40">
+                        <TextBlock
+                            x:Name="SourceHeader"
+                            x:Uid="SourceHeader"
+                            Style="{ThemeResource SubtitleTextBlockStyle}"
+                            Text="Sample source code"/>
```
stmoy: Per my other comment, I think there is some ambiguity in the word "sample". This links out to the page's source code itself, which _isn't_ the sample. It's the page that holds the samples. 

Brainstorming: thoughts on changing this text to "View this page on GitHub"? Or maybe "View on GitHub" and the links would say "C# source" and "Xaml source" or something?

XamlControlsGallery/ItemPage.xaml
chingucoding: We define the SourcePanel with `Grid.ColumnSpan=3`. In narrow layouts we don't want it to stretch over other content, so we set ColumnSpan=1 in that case.

XamlControlsGallery/ItemPage.xaml
chingucoding: How about "View page on GitHub" and then have links: "C#/XAML source code" ? That way it is clear where those link directs to.

XamlControlsGallery/ItemPage.xaml
```diff
@@ -151,14 +155,36 @@
...
+                            x:Uid="SourceHeader"
+                            Style="{ThemeResource SubtitleTextBlockStyle}"
+                            Text="View page code on GitHub"/>
+                        <HyperlinkButton
+                            x:Name="PageMarkupGitHubLink"
+                            ToolTipService.ToolTip="{x:Bind PageMarkupGitHubLink.NavigateUri, Mode=OneWay}">
+                            <TextBlock Text="XAML source code"/>
+                        </HyperlinkButton>
+                        <HyperlinkButton
+
```
stmoy: Nit: please remove this extra whitespace here and on 184.

chingucoding: Fixed

Issue #405, opened 2020-04-17
Hierarchical NavigationView


stmoy:
**Is this a totally new sample, an addition to an existing sample page, or a modification of an existing sample page?**
- [ ] New sample
- [x] Addition to existing sample
- [ ] Modification to existing sample

**If this is an addition/modification to an existing sample page, which one?**
NavigationView

**Describe the sample**

PR #402 added hierarchical navigation to the XCG, but didn't include a sample of hierarchical navigation view on the NavigationView page. This issue tracks adding a sample specifically for this feature. 

**Is your sample request related to a problem? Please describe it**

**Additional context**

anawishnoff:
I'll be trying to put together a sample before the 2.4 release.

PR #410, opened 2020-04-20
Adding Hierarchical NavigationView sample

anawishnoff:
## Description
I created a new sample for the NavigationView page that shows off hierarchy, which is a new feature being released in WinUI 2.4. 

I also added a few more sample pages into the directory so that the Hierarchical NavigationView could show off a variety of pages in its content frame. 

## Motivation and Context

Fixes issue #405. We need a Hierarchical NavigationView sample to go alongside the 2.4 release.

## How Has This Been Tested?

Tested on desktop with light/dark mode.

Tested keyboarding behavior - there are slight issues using the keyboard in this sample, but these issues are a part of NavigationView and are being tracked/fixed here: https://github.com/microsoft/microsoft-ui-xaml/pull/2271

image

## Types of changes


XamlControlsGallery/ControlPages/NavigationViewPage.xaml
```diff
@@ -375,5 +379,56 @@
                 <local:ControlExampleSubstitution x:Name="navViewASB" Key="NavViewASB"/>
             </local:ControlExample.Substitutions>
         </local:ControlExample>
+
+        <local:ControlExample Grid.Row="10"
```
stmoy, line 383: Please move this above the API-in-action sample.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
```diff
@@ -375,5 +379,56 @@
                 <local:ControlExampleSubstitution x:Name="navViewASB" Key="NavViewASB"/>
             </local:ControlExample.Substitutions>
         </local:ControlExample>
+
+        <local:ControlExample Grid.Row="10" 
+                              HeaderText="Hierarchical NavigationView"
+                              VerticalAlignment="Top" 
+                              HorizontalContentAlignment="Stretch" 
+                              XamlSource="NavigationView\NavigationViewSample8_xaml.txt"
+                              WebViewHeight="400">
```
stmoy, line 388: What is "WebViewHeight"? Please remove.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
```diff
@@ -375,5 +379,56 @@
...
+                           Grid.Row="0"
+                           Text="NavigationView supports hierarchy in Left, LeftCompact, and Top display modes. Note that Left or Top mode works best if your parent items should navigate to specific pages, rather than just serve as headings. To expand the sub-menu without selecting the parent item, click the chevron. Choose between the three modes to see how each works." />
+
+                <muxc:NavigationView x:Name="nvSample8" Grid.Row="1" Height="460"
+                                        PaneDisplayMode="Left" 
+                                        IsTabStop="False" 
+                                        SelectionChanged="NavigationView_SelectionChanged8">
+                    <muxc:NavigationView.MenuItems>
+                        <muxc:NavigationViewItem Content="Home" Icon="Home" ToolTipService.ToolTip="Home" Tag="SamplePage1"/>
+                        <muxc:NavigationViewItem Content="Parent Item 1" Icon="Save" ToolTipService.ToolTip="Parent Item 1" Tag="SamplePage2">
```
stmoy, line 413: I know the icons are mostly just placeholders, but are there other icons that help emphasize the hierarchical layout? Maybe like an icon that is a group of people as the root, and then each sub-item is just a single person? Not all the icons need to be different.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
```diff
@@ -375,5 +379,56 @@
...
+                            </muxc:NavigationViewItem.MenuItems>
+                        </muxc:NavigationViewItem>
+                        <muxc:NavigationViewItem Content="Parent Item 2" Icon="Calendar" ToolTipService.ToolTip="Parent Item 2" Tag="SamplePage7">
+                            <muxc:NavigationViewItem.MenuItems>
+                                <muxc:NavigationViewItem Content="Child C" Icon="Page" ToolTipService.ToolTip="Child 1" Tag="SamplePage5"/>
+                                <muxc:NavigationViewItem Content="Child D" Icon="Mail" ToolTipService.ToolTip="Child 2" Tag="SamplePage6"/>
+                            </muxc:NavigationViewItem.MenuItems>
+                        </muxc:NavigationViewItem>
+                    </muxc:NavigationView.MenuItems>
+                    <Frame Margin="0,0,0,0" x:Name="contentFrame8" />
```
stmoy, line 426: Is the 0 margin necessary? If not, please remove.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
```diff
@@ -278,22 +296,50 @@ private void paneFooterCheck_Click(object sender, RoutedEventArgs e)
 
         private void panePositionLeft_Checked(object sender, RoutedEventArgs e)
         {
-            if ((sender as RadioButton).IsChecked == true && nvSample != null)
+            if ((sender as RadioButton).IsChecked == true)
             {
-                nvSample.PaneDisplayMode = Microsoft.UI.Xaml.Controls.NavigationViewPaneDisplayMode.Left;
-                nvSample.IsPaneOpen = true;
-                FooterStackPanel.Orientation = Orientation.Vertical;
+                if ((sender as RadioButton).Name == "nvSampleLeft" && nvSample != null)
```
stmoy, line 301: This if statement seems to be mostly the same code copy-pasted. Can it be refactored into something cleaner?

XamlControlsGallery/SamplePages/SamplePage4.xaml
```diff
@@ -8,35 +8,53 @@
...
+                    <ColumnDefinition Width="2*" />
+                    <ColumnDefinition Width="*" />
+                    <ColumnDefinition Width="*" />
+                </Grid.ColumnDefinitions>
+                <Grid.RowDefinitions>
+                    <RowDefinition Height="Auto" />
+                </Grid.RowDefinitions>
 
-            <!--<Grid MinHeight="100" Grid.ColumnSpan="3" Margin="5">
+                <!--<Grid MinHeight="100" Grid.ColumnSpan="3" Margin="5">
```
stmoy, line 25: Is this commented out code needed? If not, please remove.

XamlControlsGallery/SamplePages/SamplePage7.xaml.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace AppUIBasics.SamplePages
+{
+    public sealed partial class SamplePage7 : Page
+    {
+        public SamplePage7()
+        {
+            this.InitializeComponent();
+        }
+
+        public void PrepareConnectedAnimation(ConnectedAnimationConfiguration config)
```
stmoy, line 15: Does this page need the ConnectedAnimation stuff? If not, please remove. My guess is that it's copy-pasta from another sample page? If it's not needed, please remove.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
```diff
@@ -74,6 +75,10 @@
             <RowDefinition Height="Auto"/>
             <RowDefinition Height="Auto"/>
             <RowDefinition Height="Auto"/>
+            <RowDefinition Height="Auto"/>
```
stmoy, line 78: I wonder if this whole thing could be replaced with a stack panel?

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
```diff
@@ -375,5 +379,56 @@
...
+            <Grid>
+                <Grid.RowDefinitions>
+                    <RowDefinition Height="Auto"/>
+                    <RowDefinition Height="Auto"/>
+                </Grid.RowDefinitions>
+
+                <TextBlock TextWrapping="WrapWholeWords"
+                           Margin="0,0,0,12"
+                           Grid.Row="0"
+                           Text="NavigationView supports hierarchy in Left, LeftCompact, and Top display modes. Note that Left or Top mode works best if your parent items should navigate to specific pages, rather than just serve as headings. To expand the sub-menu without selecting the parent item, click the chevron. Choose between the three modes to see how each works." />
```
stmoy, line 405: Is there a mode wherein the parent page does NOT result in navigation? Should that be a separate sample?

XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
```diff
@@ -171,7 +172,24 @@ private void NavigationView_SelectionChanged7(Microsoft.UI.Xaml.Controls.Navigat
                 contentFrame7.Navigate(pageType, null, args.RecommendedNavigationTransitionInfo);
             }
         }
- 
+
+        private void NavigationView_SelectionChanged8(Microsoft.UI.Xaml.Controls.NavigationView sender, Microsoft.UI.Xaml.Controls.NavigationViewSelectionChangedEventArgs args)
+        {
+            if (args.IsSettingsSelected)
```
stmoy, line 178: This code looks copy+pasted from previous SelectionChanged handlers. Do we need so many - or is it possible to refactor this into a more straightforward chunk of shared code?

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
anawishnoff: Changed it to some icons that have a natural hierarchy and make more sense.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
anawishnoff: I cleaned up this function so it added functionality for my sample (it was originally created to only be an event handler for radio buttons on another sample). Do you recommend that I combine all of the left/top/leftcompact pane clicked functions into one and just have a bunch of if statements?

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
anawishnoff: Maybe? I had assumed Yulia made it a Grid for a reason so I didn't want to change it. But I can look into it.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
anawishnoff: Yeah, it's possible that a top-level nav item doesn't actually go anywhere - just serves as a "heading" for its children sub-tree. I can change one of the top-level items in this sample so that it doesn't actually navigate anywhere - not sure if it warrants a whole other sample

XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
anawishnoff: Because the content frames aren't a property of the NavigationView, I don't see a way to combine these into one SelectionChanged handler when the sender can't give us info on which frame to update. @ojhad let me know if you have an easier solution to this.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
anawishnoff: Done- changed to StackPanel in upcoming commit.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
stmoy: No worries if this can't be easily refactored.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
stmoy: This added a little bit of confusion - why is one of the headers clickable and not the other? Maybe the text of the item could be updated to state which item has a page associated, and which is just a heading for its child sub tree? How is this documented in the guidance doc (could we use the same words here)?

stmoy: On that note: in LeftCompact mode, the user experience isn't optimal in the scenario where I'm already on Child A, and want to navigate to Child B. I need to navigate through Parent 1 even though I don't want to.

I guess this is more feedback about the feature itself. Do we have guidance docs for this scenario? CC: @Ranjeshj for FYI

XamlControlsGallery/ControlPages/NavigationViewPage.xaml.cs
stmoy: Gotcha. I just saw some duplicated code and was curious if we could optimize it. I'd prefer this than a bunch of if statements.

XamlControlsGallery/ControlPages/NavigationViewPage.xaml
ranjeshj: @stmoy do you mean arrow key navigation ? @ojhad fixed some keyboard navigation issues which should be in the latest pre-release related to arrow key navigation. Curious if you are trying with or without those changes.

anawishnoff: We have guidance docs currently in PR. But yeah, since left compact mode creates a flyout for the child menu, you have to navigate through the parent if you want to get to another child item.

I can update the headers to be more descriptive and try and show why one is just a header but the other one is clickable.

anawishnoff: From guidance doc:

When providing users with a hierarchical tree of navigation options, you may choose to make parent items non-selectable, for example when your app doesn't have a destination page associated with that parent item. To prevent an item from showing the selection indicator when invoked, set its SelectsOnInvoked property to False.

anawishnoff: After your comments and playing with the sample, I think I'm going to add this bit of guidance to the docs:

If your parent items are selectable, it's recommend you use the Left-Expanded or Top pane display modes. LeftCompact mode will cause the user to navigate to the parent item in order to open the child subtree every time its invoked.

PR #429, opened 2020-05-15
Fixed stray whitespaces in multiline code samples

Sergio0694:
## Description

This PR fixes leftover whitespace characters at the end of individual lines within multiline samples.

## Motivation and Context

Having leftover whitespaces makes the UX worse when copy-pasting samples into other projects to try them out, as you need to manually check line by line whether the code has unwanted characters.

## How Has This Been Tested?

- From code, by testing the generated `string` before/after the new logic added in this PR.
- In the app, by highlighting and copy/pasting various code samples with this issue.

**Before:**
image
**After:**
image

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlExample.xaml.cs
+++ b/XamlControlsGallery/ControlExample.xaml.cs
@@ -12,6 +12,7 @@ using ColorCode;
 using ColorCode.Common;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text.RegularExpressions;
 using Windows.Foundation;
 using Windows.Foundation.Metadata;
@@ -259,6 +260,9 @@ namespace AppUIBasics
         {
             // Trim out stray blank lines at start and end.
             sampleString = sampleString.TrimStart('\n').TrimEnd();
+
+            // Also trim out spaces at the end of each line
+            sampleString = string.Join('\n', sampleString.Split('\n').Select(s => s.TrimEnd()));
 
             // Perform any applicable substitutions.
             sampleString = SubstitutionPattern.Replace(sampleString, match =>


XamlControlsGallery/ControlExample.xaml.cs
```diff
@@ -260,6 +261,9 @@ private void FormatAndRenderSampleFromString(string sampleString, ContentPresent
             // Trim out stray blank lines at start and end.
             sampleString = sampleString.TrimStart('\n').TrimEnd();
 
+            // Also trim out spaces at the end of each line
+            sampleString = string.Join('\n', sampleString.Split('\n').Select(s => s.TrimEnd()));
```
chingucoding, line 265: While I am not a fan to add even more processing for every sample, this is probably the best solution to fix this.

Sergio0694: Thanks for the review 👍

Yeah I agree that it'd be nice to minimize the amount of processing per sample, but this shouldn't really add much overhead anyway. Technically it's possible to optimize this with a single allocation (and also making it faster), by leveraging `ArrayPool<T>` and `Span<T>` APIs (specifically, `ReadOnlySpan<T>.IndexOf` and `ReadOnlySpan<T>.Trim`, though it seemed a bit overkill for this scenario and it would've required the additional `System.Memory` dependency, so I just left it at that for now 😄

chingucoding: I think in the long term, instead of running the processing every time the sample is opened, we should have an automatic check that tells us what we need to change, so we can simply have the sample be "correct" and we don't have to run any correction steps on the users end.

stmoy: +1 to @chingucoding - the code change here is a great "catch-all", but even better would be if our samples were updated to be correct meaning this code never needs to actually be run. But for now, this is a reasonable change - thanks @sergio0694!

Sergio0694: Happy to help! 😊
I believe there should be a stylecop tool for that, the Windows Community Toolkit has that and it shows warnings whenever a line has any trailing whitespaces. You could probably just reuse that one, or at least use that as a starting point 👍

chingucoding: Congrats @Sergio0694 on getting your first PR on the XCG merged!

Does stylecop also do that for x:String in XAML? Or does it do that for every file, regardless of type and content?

Sergio0694: Not a stylecop expert, but I believe it can be configured to work on specific file types.

PR #457, opened 2020-05-29
Fix for Issue#2457 - Crash when rearranging TabView tabs bound to UIElements

RBrid:
Fixes Issue #2457 in microsoft/microsoft-ui-xaml repo.

## Description
The TabViewItem's DataContext includes a UIElement (in this case a Frame) and when a tab gets dropped during a re-ordering operation that element gets reparented into a new ListView item container while it is still used in an old container because of an animation. This causes the "Element already parented" exception.
(I also reordered the 'using' statements at the top of the file)

## Motivation and Context
The TabViewItem's DataContext must either not contain a UIElement or the inner ListView's ItemContainerTransitions collection must not include AddDeleteThemeTransition/ContentThemeTransition animations. The combination of both causes the crash. The ListView by itself exhibits the same behavior - this is not TabView-specific.

## How Has This Been Tested?
Adhoc testing with 19H1 with System animations turned on.

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/TabViewPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/TabViewPage.xaml.cs
@@ -1,14 +1,16 @@
 using System;
+using System.Collections.ObjectModel;
+using AppUIBasics.SamplePages;
+using AppUIBasics.TabViewPages;
+using Microsoft.UI.Xaml.Controls;
+using Windows.ApplicationModel.Core;
+using Windows.UI.Core;
+using Windows.UI.ViewManagement;
+using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Input;
-using Microsoft.UI.Xaml.Controls;
-using AppUIBasics.SamplePages;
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/TabViewPage.xaml.cs
```diff
@@ -26,10 +28,37 @@ public sealed partial class TabViewPage : Page
         public TabViewPage()
         {
             this.InitializeComponent();
+            Loaded += TabViewPage_Loaded;
 
             InitializeDataBindingSampleData();
         }
 
+        private void TabViewPage_Loaded(object sender, RoutedEventArgs e)
```
stmoy, line 36: This is a heavy hammer; yes, it avoids the crash, but this chunk of code is divorced from the problem sample. I'm following up over on the main thread.

If the resulting guidance is "disable the animation for the TabView", I think we'll want to localize the code to disable to the specific sample itself so it is more obvious what is happening and also to help increase copy-paste'ability.

stmoy:
We're investigating a framework-side fix for this common case. Closing this PR for now.

PR #471, opened 2020-06-11
Allowing keyboard input for the ScrollViewer

RBrid:
User could not use Ctrl +/- to zoom in and out of the ScrollViewer in the ScrollViewer sample page.
Arrows, PageUp, PageDown, etc did not work either to scroll its Image. The IsTabStop=True setting allows the control to take focus & keyboard input works now.

This is for internal bug 26337703 - Cannot set focus to ScrollViewer.

## Testing
Ad-hoc testing of the ScrollViewer sample page.

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ScrollViewerPage.xaml
+++ b/XamlControlsGallery/ControlPages/ScrollViewerPage.xaml
@@ -18,7 +18,7 @@
     <StackPanel>
         <local:ControlExample x:Name="Example1" HeaderText="Content inside of a ScrollViewer.">
 
-            <ScrollViewer x:Name="Control1" Height="270" ZoomMode="Disabled" IsVerticalScrollChainingEnabled="True"
+            <ScrollViewer x:Name="Control1" Height="270" ZoomMode="Disabled" IsTabStop="True" IsVerticalScrollChainingEnabled="True"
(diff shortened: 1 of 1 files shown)


chingucoding:
Is it expected that when `HorizontalScrollBarVisibility` is set `Hidden`, that scrolling to the left and right doesn't work persistently:

gif

RBrid:
@chingucoding, no that is not normal & I could not repro this.  @stmoy, can you repro this? Thanks.

stmoy:
@RBrid - I am able to reproduce this issue. 

Steps:
1. On the ScrollViewer page in the Xaml Controls Gallery app
1. Set ZoomMode = Enabled
1. Zoom in a bit using the Zoom slider
1. Set ScrollbarVisibility.Horizontal = Visible, or Hidden
1. Tab to the image to give it focus
1. Hold the down arrow key to scroll the image down
1. Release the down arrow key and quickly press the right arrow key

Expected:
Image starts scrolling to the right from the current vertical position

Actual:
Image starts scrolling to the right from the top of the image (ie. it looks like it jumps to the top)

Note that if you wait a second between pressing down and pressing right, the behavior does not repro for me.

stmoy:
Followed up with @RBrid - it looks like this is indeed a framework issue and should not block this PR.

PR #476, opened 2020-06-18
Prevent crashes on ImplicitAnimation page by checking for NaN

### LINKED ISSUE #475: [CRASH] Motion > Implicit Transition > Empty Opacity field
**Describe the bug**
I tried to click on the cross to empty the field and see what happens when clicking the button... welp, it crashes !

1. Go to Motion > Implicit Transitions
2. Empty the Opacity field
3. Click on `Set Opacity` button

**Expected behavior**
Setting opacity to 0.

**Actual behavior**
App crashes.

**Screenshots**
image
image
I cannot show the app crashing...

**Version Info**

Latest (1.2.15.0)

Windows 10 version:
- [ ] Insider Build (xxxxx)
- [X] November 2019 Update (18363)
- [ ] May 2019 Update (18362)
- [ ] Creators Update (15063)

Device form factor:
- [X] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Question**
Is it done in purpose ?

chingucoding:
## Description

The page crashed when setting the opacity and other values in samples on the page since a cleared NumberBox will report double.NaN as its value which is not a valid value for those cases. This is now fixed by wrapping the "get call" and setting the NumberBox value to 0 if its NaN.
## Motivation and Context

Fixes #475 
## How Has This Been Tested?

Tested manually by clearing the NumberBox values and clicking the "set value buttons" on the samples.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ImplicitTransitionPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/ImplicitTransitionPage.xaml.cs
@@ -46,7 +46,7 @@ namespace AppUIBasics.ControlPages
 
             RotationRectangle.CenterPoint = new System.Numerics.Vector3((float)RotationRectangle.ActualWidth / 2, (float)RotationRectangle.ActualHeight / 2, 0f);
 
-            RotationRectangle.Rotation = (float)EnsureValueIsNumber(RotationNumberBox);
+            RotationRectangle.Rotation = EnsureValueIsNumber(RotationNumberBox);
         }
         private void ScaleButton_Click(object sender, RoutedEventArgs e)
         {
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/ImplicitTransitionPage.xaml.cs
```diff
@@ -137,6 +137,15 @@ private void BackgroundButton_Click(object sender, RoutedEventArgs e)
             }
         }
 
+        private double EnsureValueIsNumber(NumberBox numberBox)
```
stmoy, line 140: Could this return a float instead? Then we wouldn't need to cast above.

chingucoding: Good point, fixed now.

PR #497, opened 2020-07-09
Fix issue with ConnectedAnimation crashing due to selection being null

### LINKED ISSUE #494: Pressing Navigate in the Connected Animation page crashes the app
**Describe the bug**

The Default connected animation is causing a crash.

1. Go to 'Connected Animation page
2. scroll down to A simple connected animation
3. Click the sample's Navigate button (without changing the radio button selection. If you click on one of the configurations first, the crash does not occur)

**Expected behavior**

Animation plays

**Actual behavior**

The app crashes

**Screenshots**

image

**Version Info**

1.2.15.0

**Additional context**

chingucoding:
## Description

The sample crashed as on a freshly loaded page, the selecteditem of the RadioButtons was null. By setting SelectedIndex in XAML, the RadioButtons now always know the selected item, and thus fixing the crash.
## Motivation and Context

Closes #494 
## How Has This Been Tested?

Opened ConnectedAnimationPage and pressed the "Navigate" button. App did not crash.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ConnectedAnimationPage.xaml
+++ b/XamlControlsGallery/ControlPages/ConnectedAnimationPage.xaml
@@ -29,8 +29,8 @@
                     <Button Content="Navigate" Click="NavigateButton_Click" HorizontalAlignment="Stretch" />
 
                     <contract7Present:TextBlock Text="Configurations" Style="{ThemeResource BaseTextBlockStyle}" Margin="0,6" />
-                    <muxccontract7Present:RadioButtons x:Name="ConfigurationPanel">
-                        <contract7Present:RadioButton Content="Default" IsChecked="True" />
+                    <muxccontract7Present:RadioButtons x:Name="ConfigurationPanel" SelectedIndex="0">
+                        <contract7Present:RadioButton Content="Default" />
                         <contract7Present:RadioButton Content="Gravity" />
                         <contract7Present:RadioButton Content="Direct" />
                         <contract7Present:RadioButton Content="Basic" />


stmoy:
Thank you for the fix. I'm wondering if this is a platform bug. CC: @ranjeshj @kikisaints @YuliKl - thoughts?

chingucoding:
> Thank you for the fix. I'm wondering if this is a platform bug. CC: @ ranjeshj @ kikisaints @ YuliKl - thoughts?

I'm inclined to say "no" here, because it isn't really like we are setting the selection on the RadioButtons, but we are rather telling the single RadioButton to be checked. I am not sure if this is a very common scenario or if it is "expected", after all "Checked" is only something you would use if you use multiple RadioButton, but asking the RadioButtons Control what is selected would be more in the realm of "RadioButtons" handle selection.

Those two seem more or less exclusive to me.

stmoy:
If the bar for RadioButtons control is "we expect it to work wherever StackPanel was used", then this would violate that bar. (Not sure if this is the bar, hence my question to Ranjesh/Kiki/Yulia)

I guess it's a bit of a semantic question - should RadioButton be responsible for knowing which thing is selected, or should RadioButtons? Previously it was RadioButton, but now I don't know if it is "either/or" or **only** RadioButtons. I think this change suggests the latter, but I don't know if that has been well articulated.

YuliKl:
This had felt like an app bug to me when I opened it. But I defer to @kikisaints for clarification on how these controls were designed to work.

ranjeshj:
> If the bar for RadioButtons control is "we expect it to work wherever StackPanel was used", then this would violate that bar. (Not sure if this is the bar, hence my question to Ranjesh/Kiki/Yulia)
> 
> I guess it's a bit of a semantic question - should RadioButton be responsible for knowing which thing is selected, or should RadioButtons? Previously it was RadioButton, but now I don't know if it is "either/or" or **only** RadioButtons. I think this change suggests the latter, but I don't know if that has been well articulated.

This is interesting. Selection on the RadioButtons control is conveyed through the checked state of the RadioButton. Should changing checked state trigger a selection change? What if you set two RadioButtons to checked? @StephenLPeters do you remember discussing this issue ?

stmoy:
Even though this smells like a platform bug to me, it's not likely to be fixed anytime soon. I'll accept this PR in the meantime while we discuss if this is desired behavior or not.

PR #500, opened 2020-07-11
RadioButtons cleanup

stmoy:
## Description

This change further delineates the samples for RadioButton vs. RadioButtons. Previously they shared an icon, were found in the same category, etc. - this change moves RadioButtons to Layouts and updates the wording for both samples.  

## Motivation and Context

RadioButton and RadioButtons are related, but conceptually different. This change further differentiates the two samples. 

## How Has This Been Tested?

Manual testing. 

image

image

## Types of changes


stmoy:
CC: @chingucoding

XamlControlsGallery/ControlPages/RadioButtonPage.xaml
```diff
@@ -15,35 +15,25 @@
...
-    &lt;TextBlock Text="Options:"/&gt;
-    &lt;RadioButton x:Name="Option1RadioButton" Content="Option 1" Checked="Option1RadioButton_Checked"/&gt;
-    &lt;RadioButton x:Name="Option2RadioButton" Content="Option 2" Checked="Option2RadioButton_Checked"/&gt;
-    &lt;RadioButton x:Name="Option3RadioButton" Content="Option 3" Checked="Option3RadioButton_Checked"/&gt;
-&lt;/StackPanel&gt;
+&lt;muxc:RadioButtons&gt;
+    &lt;RadioButton Content="Option 1" Checked="RadioButton_Checked"/&gt;
+    &lt;RadioButton Content="Option 2" Checked="RadioButton_Checked"/&gt;
+    &lt;RadioButton Content="Option 3" Checked="RadioButton_Checked"/&gt;
+&lt;/muxc:RadioButtons&gt;
```
chingucoding, line 34: If we are switching back to RadioButtons as panel, we should set the header property.

XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -894,12 +894,11 @@
         {
           "UniqueId": "RadioButton",
           "Title": "RadioButton",
-          "Subtitle": "A control that allows a user to select a single option from a group of options. When radio buttons are grouped together they are mutually exclusive.",
+          "Subtitle": "A control that allows a user to select a single option from a group of options.",
           "ImagePath": "ms-appx:///Assets/RadioButton.png",
-          "Description": "Use RadioButtons to let a user choose between mutually exclusive, related options.",
+          "Description": "Use RadioButtons to let a user choose between mutually exclusive, related options. Generally contained within a RadioButtons group control.",
```
chingucoding, line 899: Are RadioButton_s supposed to always used within a RadioButtons group control? Is there a benefit to using a RadioButton instead of something like a TextBlock or simply a string when I am using the RadioButtons control?

From a simplicity point of view, putting RadioButton_s inside a RadioButtons control seems a bit strange, after all those RadioButton technically have their own checked state etc, while the containing RadioButtons also has their own selectionchanged.

XamlControlsGallery/DataModel/ControlInfoData.json
```diff
@@ -1376,6 +1349,31 @@
             "VariableSizedWrapGrid",
             "RelativePanel"
           ]
+        },
+                {
+          "UniqueId": "RadioButtons",
+          "Title": "RadioButtons",
+          "Subtitle": "A control that displays a group of related RadioButton controls with keyboarding and accessibility support.",
+          "ImagePath": "ms-appx:///Assets/RadioButtons.png",
+          "Description": "A control that displays a group of related RadioButton controls with keyboarding and accessibility support.",
```
chingucoding, line 1358: Should we also mention that it allows passing simple text or other UI elements and it automagically wraps those? This wording seems a bit like it doesn't add much to existing RadioButton_s.

XamlControlsGallery/DataModel/ControlInfoData.json
stmoy: Good question. The [guidance doc](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/radio-button) doesn't mention including things that are _not_ RadioButton controls. I was under the impression that RadioButton is the main thing to be used within a RadioButtons, hence these samples. 

Automagically converting text to RadioButtons is a cool function, though I can't tell if it's a feature or a bug from the guidance docs 😅 @kikisaints can you help clarify?

As it pertains to this PR - I'd like to get this pulled in before I submit a new version of the Gallery to the Store this week. Automagic conversion should probably have its own sample pending Kiki's input, so I don't think this feedback blocks this PR.

XamlControlsGallery/DataModel/ControlInfoData.json
stmoy: Discussed here: https://github.com/microsoft/Xaml-Controls-Gallery/pull/500#discussion_r453809040

XamlControlsGallery/DataModel/ControlInfoData.json
chingucoding: I agree that this PR shouldn't be blocked on this. Regarding the guidance docs, I think the "automagic wrapping" seems intentional, after all there is an example that uses plain strings instead of RadioButtons and another example binding to a collection of objects which are not RadioButton controls.

stmoy: Good point. I see what you mean now. I'll update the description.

stmoy:
I'm planning on pulling this in early afternoon today - @chingucoding, are you ok with the PR now that I've made some changes?

PR #518, opened 2020-07-29
Add ScreenshotMode

teaP:
This change adds a toggleswitch for ScreenshotMode to the settings page, as well as the ability to choose where the screenshots will end up. 

image

When ScreenshotMode is on, the example pages hide the xaml text, add padding around the control, change the background color so you can exactly see the area where the screenshot will be taken, and show a few extra controls:

image

The screenshot button takes an immediate screenshot of the related control and saves it with the best name it can cobble together. The delay button takes a screenshot after 3 seconds, allowing you time to open a flyout or take pictures of hover/press effects. (The mechanism of these two is entirely different because there's no one good solution; also, unfortunately, neither one will capture theme shadows.) You can also change the padding around the control in order to make space for flyouts.

In order to keep status text out of the screenshot, I added an Output section to ControlExample and moved status text in various different control pages into that area. This is the only change to non-screenshot mode that should be visible.


XamlControlsGallery/ControlExample.xaml.cs
```diff
@@ -350,5 +366,177 @@ private void OnValueChanged(ControlExampleSubstitution sender, object e)
...
+                ScreenshotStatusTextBlock.Text = i.ToString();
+                await Task.Delay(1000);
+            }
+            ScreenshotStatusTextBlock.Text = "Image captured";
+
+            // AppRecordingManager is desktop-only, and its use here is quite hacky,
+            // but it is able to capture popups (though not theme shadows).
+
+            bool isAppRecordingPresent = ApiInformation.IsTypePresent("Windows.Media.AppRecording.AppRecordingManager");
+            if (isAppRecordingPresent)
```
chingucoding, line 422: If the AppRecordingManager is not present, maybe we should have a fallback in place, instead of just doing nothing. How about calling `TakeScreenshot` in that case?

XamlControlsGallery/ControlExample.xaml
```diff
@@ -46,6 +46,20 @@
...
+            <VisualStateGroup>
+                <VisualState x:Name="NormalMode"/>
+                <VisualState x:Name="ScreenshotMode">
+                    <VisualState.Setters>
+                        <Setter Target="CodeScrollViewer.Visibility" Value="Collapsed"/>
+                        <Setter Target="ControlPresenter.Padding" Value="{ThemeResource ControlElementScreenshotModePadding}"/>
+                        <Setter Target="ControlPresenter.HorizontalAlignment" Value="Center"/>
+                        <Setter Target="ScreenshotControls.Visibility" Value="Visible"/>
+
+                        <Setter Target="ExampleContainer.Background" Value="Gray"/>
```
YuliKl, line 59: I'm not a huge fan of hardcoded colors, maybe a theme resource would be better?

XamlControlsGallery/App.xaml
```diff
@@ -59,6 +59,8 @@
             <Thickness x:Key="PageHeaderDefaultPadding">12,28,12,0</Thickness>
             <Thickness x:Key="PageHeaderMinimalPadding">-4,0,12,0</Thickness>
 
+            <Thickness x:Key="ControlElementScreenshotModePadding">67</Thickness>
```
chingucoding, line 62: Is there a specific reason that 67 is being used here or does it just look good?

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -18,13 +18,12 @@
...
         <local:ControlExample x:Name="Example1" HeaderText="A simple Button with text content.">
             <local:ControlExample.Example>
-                <RelativePanel>
-                    <Button x:Name="Button1" Content="Standard XAML button" Click="Button_Click"
-                            IsEnabled="{x:Bind DisableButton1.IsChecked.Value.Equals(x:False), Mode=OneWay}" />
-                    <TextBlock x:Name="Control1Output" Margin="8,0,0,0" FontFamily="Global User Interface"
-                            RelativePanel.RightOf="Button1" RelativePanel.AlignVerticalCenterWith="Button1" />
-                </RelativePanel>
+                <Button x:Name="Button1" Content="Standard XAML button" Click="Button_Click"
+                    IsEnabled="{x:Bind DisableButton1.IsChecked.Value.Equals(x:False), Mode=OneWay}" />
```
YuliKl, line 22: Wait, we can do cool things like this in markup? ❤
🌈The more you know 🌈

XamlControlsGallery/SettingsPage.xaml
```diff
@@ -55,12 +56,25 @@
...
+
+                <TextBlock
+                    Style="{StaticResource TitleTextBlockStyle}"
+                    Margin="0,40,0,0"
+                    FontWeight="Normal"
+                    Text="Screenshots" />
+                <StackPanel Margin="0,10,0,0">
+                    <ToggleSwitch x:Name="screenshotModeToggle" Header="Screenshot mode" Toggled="screenshotModeToggle_Toggled" OffContent="Off" OnContent="On" Margin="0,0,0,4"/>
+                    <Button Content="Set screenshot folder" Click="FolderButton_Click"/>
+                    <TextBlock x:Name="screenshotFolderTextBlock" IsTextSelectionEnabled="True"/>
```
chingucoding, line 75: I think it would be awesome if you could also open the folder from inside the XAML Controls Gallery by clicking on a button, which would open the folder in explorer.

XamlControlsGallery/SettingsPage.xaml.cs
```diff
@@ -129,5 +140,19 @@ private void navigationLocation_SelectionChanged(object sender, SelectionChanged
         {
             NavigationOrientationHelper.IsLeftMode = navigationLocation.SelectedIndex == 0;
         }
+
+        private async void FolderButton_Click(object sender, RoutedEventArgs e)
+        {
+            FolderPicker folderPicker = new FolderPicker();
+            folderPicker.SuggestedStartLocation = PickerLocationId.PicturesLibrary;
+            folderPicker.FileTypeFilter.Add(".png");
```
chingucoding, line 148: We need this because otherwise the file picker crashes right? Maybe we could add a comment for that?

XamlControlsGallery/ControlExample.xaml.cs
teaP: Good call, I did just that.

XamlControlsGallery/ControlExample.xaml
teaP: Agreed, fixed!

XamlControlsGallery/App.xaml
teaP: It's the value I was given by the documentation team.

XamlControlsGallery/ControlPages/ButtonPage.xaml
teaP: 😀

XamlControlsGallery/SettingsPage.xaml
teaP: Good idea! I changed it to a link.

XamlControlsGallery/SettingsPage.xaml.cs
teaP: Yes, it's entirely meaningless as far as I can tell, but if you don't add at least one file type it gets sad. I added a comment.

chingucoding:
Just making sure, is it expected that the screenshot also shows the header of the samples?

image

XamlControlsGallery/App.xaml
chingucoding: Alright, I see. Thanks for the clarification.

teaP:
> Just making sure, is it expected that the screenshot also shows the header of the samples?

Whaaaaat no, it isn't. How did you get that to happen?

chingucoding:
Steps to repro:
1. Enable screenshot mode
2. Visit Buttons page
3. Click on "Take screenshot with delay"

That gave me the posted image.

teaP:
It does that every time? I can't repro at all.

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -18,13 +18,12 @@
...
             </local:ControlExample.Example>
+            <local:ControlExample.Output>
+                <TextBlock x:Name="Control1Output" FontFamily="Global User Interface"/>
+            </local:ControlExample.Output>
```
chingucoding, line 26: I am not sure if this "styling" is ideal. Maybe we can move the output somewhere else? The top sample looks a bit strange in my opinion.

image

chingucoding:
> It does that every time? I can't repro at all. With other examples too?

It happens with all samples I have tested, but only when using the "take screenshot with delay" option. The "direct" screenshot works fine.

teaP:
> > It does that every time? I can't repro at all. With other examples too?
> 
> It happens with all samples I have tested, but only when using the "take screenshot with delay" option. The "direct" screenshot works fine.

Wow, I can't fathom what's happening. Is there anything about your computer setup that could be related -- any scaling going on? Actually, can you do me a favor? If you could go into the TakeScreenshotWithDelay method and comment out this line:

```
// Delete intermediate file
await screenshotFile.DeleteAsync();
```

and then check out what the full-window screenshot looks like?

chingucoding:
Here are the screenshots:

appScreenshot
DropDownButton_Simple

Could it be that it is related to the NavigationView being in top mode?

teaP:
> Could it be that it is related to the NavigationView being in top mode?

That's exactly what it is, yes. I totally can't imagine why yet but I'll dig into it. Thank you very much!

teaP:
OK, I fixed the issue -- the thing is that when NavView is on the left, it extends into the titlebar, so I didn't have to take into account the height of the titlebar in the screenshot, but when it's on top that isn't the case. Unfortunately, there seems to be no way to get the magic 32px number so I just hardcoded it in.

YuliKl:
> Unfortunately, there seems to be no way to get the magic 32px number so I just hardcoded it in.

I think that can be had from [CoreApplicationViewTitleBar Height](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.core.coreapplicationviewtitlebar.height?view=winrt-19041)

teaP:
> > Unfortunately, there seems to be no way to get the magic 32px number so I just hardcoded it in.
> 
> I think that can be had from [CoreApplicationViewTitleBar Height](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.core.coreapplicationviewtitlebar.height?view=winrt-19041)

You would absolutely think that and I spent a very long amount of time trying to make that work, but the bottom line is that it only returns 0 if you aren't extending into the title bar, which is the exactly situation in which I need the number 32 instead. :/

chingucoding:
Given that the NavigationView hardcodes the 32px, and this is shipped to all apps, I think in our case, it is also fine to hardcode that margin/difference of 32px.

YuliKl:
> > I think that can be had from [CoreApplicationViewTitleBar Height](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.core.coreapplicationviewtitlebar.height?view=winrt-19041)
> 
> You would absolutely think that and I spent a quite an amount of time trying to make it work, but the bottom line is that it only returns 0 if you aren't extending into the title bar, which is the exact situation in which I need the number 32 instead. :/

Amazing. Maybe @oldnewthing can explain to us why this is the case...

oldnewthing:
The purpose of CoreApplicationViewTitleBar.Height is to tell you how tall to draw your replacement title bar. If you're not extending, then you shouldn't draw a replacement title bar at all, so it reports zero. There is no "What would the title bar height be *if I were extending* (even though I'm not)", but maybe you can extend, grab the value, and then stop extending.

```cs
        async Task<double> GetTitleBarHeightAsync()
        {
            var titleBar = CoreApplication.GetCurrentView().TitleBar;
            if (titleBar.ExtendViewIntoTitleBar)
            {
                return titleBar.Height;
            }
            var source = new TaskCompletionSource<double>();
            TypedEventHandler<CoreApplicationViewTitleBar, object> handler = (s, e) =>
            {
                source.TrySetResult(titleBar.Height);
            };
            titleBar.LayoutMetricsChanged += handler;
            titleBar.ExtendViewIntoTitleBar = true;
            var height = await source.Task;
            titleBar.LayoutMetricsChanged -= handler;
            titleBar.ExtendViewIntoTitleBar = false;
            return height;
        }
```

teaP:
> The purpose of CoreApplicationViewTitleBar.Height is to tell you how tall to draw your replacement title bar. If you're not extending, then you shouldn't draw a replacement title bar at all, so it reports zero. There is no "What would the title bar height be _if I were extending_ (even though I'm not)", but maybe you can extend, grab the value, and then stop extending.

It makes sense -- needing to know the size of the non client area is for sure an edge case. :)

chingucoding:
This is very nitpicky, but I just noticed this border here, and I am wondering if it is desired:

image

teaP:
> This is very nitpicky, but I just noticed this border here, and I am wondering if it is desired:

It does seem that the crop is off by a pixel here or there sometimes... I made adjustments to the cropping area, please let me know if you can still repro!

XamlControlsGallery/Helper/UIHelper.cs
```diff
@@ -1,12 +1,16 @@
...
 using System.Linq;
+using Windows.Storage;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Media;
 
 namespace AppUIBasics.Helper
 {
     public static class UIHelper
     {
+        public static bool IsScreenshotMode { get; set; }
```
chingucoding, line 11: Does it make sense to store this in the UIHelper?

XamlControlsGallery/ControlExample.xaml.cs
```diff
@@ -350,5 +367,188 @@ private void OnValueChanged(ControlExampleSubstitution sender, object e)
...
+        {
+            TakeScreenshot();
+        }
+
+        private void ScreenshotDelayButton_Click(object sender, RoutedEventArgs e)
+        {
+            TakeScreenshotWithDelay();
+        }
+
+        private async void TakeScreenshot()
```
chingucoding, line 381: How about moving this onto a separate helper class, to which we pass the UIElement? This is not really logic related to the ControlExample, so it might be better to have it in a separate class.

XamlControlsGallery/Package.appxmanifest
```diff
@@ -41,5 +41,6 @@
   </Applications>
   <Capabilities>
     <Capability Name="internetClient" />
+    <uap6:Capability Name="graphicsCapture"/>
```
chingucoding, line 44: Why did we only update the non debug package file? Is it because the debug version will never be published to the store?

XamlControlsGallery/ControlExample.xaml.cs
teaP: I think it's fine -- the screenshot controls are part of the ControlExample and there's not that much of it anyway, and no one else is using it.

XamlControlsGallery/Helper/UIHelper.cs
teaP: It's UI, and it's helping. :D

XamlControlsGallery/Package.appxmanifest
teaP: I didn't realize there was another one! I added the capability there as well.

XamlControlsGallery/ControlExample.xaml.cs
chingucoding: Fair enough.

XamlControlsGallery/Helper/UIHelper.cs
chingucoding: You are technically correct, the best kind of correct!

XamlControlsGallery/ControlPages/ButtonPage.xaml
stmoy: Agreed that the styling is not ideal - but not sure how to fix it. Open to suggestions! We'll go with this styling for now, but we can always update it later.

XamlControlsGallery/XamlControlsGallery.csproj
```diff
@@ -1217,7 +1217,11 @@
     <None Include="testcert.pfx" />
     <None Include="XamlControlsGallery_TemporaryKey.pfx" />
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <SDKReference Include="WindowsDesktop, Version=10.0.18362.0">
+      <Name>Windows Desktop Extensions for the UWP</Name>
```
stmoy, line 1222: Why is this needed?

teaP: AppRecordingManager is a desktop only API, but it's also the only thing that captures popups in situ.

XamlControlsGallery/ControlExample.xaml.cs
```diff
@@ -350,5 +367,188 @@ private void OnValueChanged(ControlExampleSubstitution sender, object e)
...
+                    imageName = fileName + ".png";
+                }
+            }
+            else if (!String.IsNullOrWhiteSpace(Name))
+            {
+                // Put together the page name and the control example name
+                UIElement uie = this;
+                while (uie != null && !(uie is Page))
+                {
+                    uie = VisualTreeHelper.GetParent(uie) as UIElement;
```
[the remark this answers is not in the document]
teaP: Is TabView the only one? Can I just add x:Names there? There's not much else to go on as far as names go.... the other option is to name it after its index into it's parent's children, maybe.

XamlControlsGallery/ControlExample.xaml.cs
```diff
@@ -350,5 +367,188 @@ private void OnValueChanged(ControlExampleSubstitution sender, object e)
...
+        }
+
+        private void ControlPaddingBox_KeyUp(object sender, Windows.UI.Xaml.Input.KeyRoutedEventArgs e)
+        {
+            if (e.Key == Windows.System.VirtualKey.Enter && !String.IsNullOrWhiteSpace(ControlPaddingBox.Text))
+            {
+                // Evaluate the text as padding
+                string[] strs = ControlPaddingBox.Text.Split(new char[] { ' ', ',' });
+                double[] nums = new double[4];
+                for (int i = 0; i < strs.Length; i++)
```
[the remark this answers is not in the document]
teaP: That's fair. Fixed.

XamlControlsGallery/ControlExample.xaml.cs
```diff
@@ -350,5 +367,188 @@ private void OnValueChanged(ControlExampleSubstitution sender, object e)
...
+        }
+
+        private void ControlPaddingChangedCallback(DependencyObject sender, DependencyProperty dp)
+        {
+            ControlPaddingBox.Text = ControlPresenter.Padding.ToString();
+        }
+
+        private void ControlPaddingBox_KeyUp(object sender, Windows.UI.Xaml.Input.KeyRoutedEventArgs e)
+        {
+            if (e.Key == Windows.System.VirtualKey.Enter && !String.IsNullOrWhiteSpace(ControlPaddingBox.Text))
```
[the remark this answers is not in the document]
teaP: Sure! Done.

XamlControlsGallery/ControlExample.xaml.cs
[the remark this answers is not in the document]
stmoy: There are a lot of examples:
- AppBarButton (Reveal style)
- AppBarToggleButton
- AutomationProperties 
- Button (Reveal style, AccentStyle)

etc. - there's a bunch, this is just the first few in alphabetical order.

Indexing in parent's children seems reasonable; I was going to suggestion "don't overwrite old pictures" if possible (so you'd have "Button1.png, Button2.png" or some other poorly named thing).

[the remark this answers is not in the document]
teaP: Fixed, I added the index after the page name.

XamlControlsGallery/ControlExample.xaml
```diff
@@ -80,23 +95,41 @@
                     <RowDefinition Height="Auto" />
                 </Grid.RowDefinitions>
                 <Grid.ColumnDefinitions>
+                    <ColumnDefinition />
```
stmoy, line 98: Should this be Width="*"?

It seems like something in this change is causing the samples to be horizontally squished. 

Before:
image


After:
image

stmoy: Just kidding - I actually think the next column needs to have Width = Auto.

What is happening right now is that neither the sample or the output column have their width set, so I think they both default to *, when the output column should actually collapse. 

That said, I don't know if that makes the screenshots too wide for some samples.

PR #524, opened 2020-08-07
Update samples that use the RichEditBox

chingucoding:
## Description

A few changes:
- Remove code that reapplied rtf when switching focus
- Switch default color from black to gray

I am not happy with the default foreground being gray, however this is the easy approach to fix the issue with black text on black background in dark theme. Having those depend on the current theme will make the samples definitely more complex, that is why I opted out of doing this for now.
 

## Motivation and Context

Update samples as the recent changes to the RichEditBox control broke them.

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/SplitButtonPage.xaml
+++ b/XamlControlsGallery/ControlPages/SplitButtonPage.xaml
@@ -1,4 +1,4 @@
-﻿<Page x:Class="AppUIBasics.ControlPages.SplitButtonPage" 
+<Page x:Class="AppUIBasics.ControlPages.SplitButtonPage" 
       xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
       xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" 
       xmlns:local="using:AppUIBasics"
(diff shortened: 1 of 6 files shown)


XamlControlsGallery/ControlPages/SplitButtonPage.xaml
```diff
@@ -22,7 +22,7 @@
                 </Grid.ColumnDefinitions>
 
                 <muxc:SplitButton x:Name="myColorButton" AutomationProperties.Name="Font color" Padding="0" MinHeight="0" MinWidth="0" VerticalAlignment="Top" Click="myColorButton_Click">
-                    <Rectangle x:Name="CurrentColor" Width="{StaticResource SwatchSize}" Height="{StaticResource SwatchSize}" Fill="Black" Margin="0"/>
+                    <Rectangle x:Name="CurrentColor" Width="{StaticResource SwatchSize}" Height="{StaticResource SwatchSize}" Fill="Gray" Margin="0"/>
```
stmoy, line 25: Instead of Gray, could we make the default some obviously obnoxious color like Teal? I'm worried that Gray will make the control look like it's in the Disabled state, but it's not. If we choose an obviously customized color, it'll help indicate that a non-default color has been set.

chingucoding: Yeah, gray wasn't a great choice. I've switched to green now as it is one of the colors we had in the selection, that also is fairly  visible in both themes, and isn't to aggressive to the eye.

PR #535, opened 2020-08-21
Fix issue with resizing having horrible performance

### LINKED ISSUE #529: Resizing SplitButton page is very laggy
**Describe the bug**
Open app to SplitButton page

Attempt to resize or move the window

**Expected behavior**
Window resizes/moves correctly

**Actual behavior**
Laggy resize or maximize

Doesn't seem to affect other pages

chingucoding:
## Description

The issue was that setting the foreground is very expensive, however changing it to the value it already is doesn't have any benefit for us.

Through performance debugger, I found out that we spend 60% of CPU time on these two lines when resizing the window. By adding this check, resizing the window now works as expected again.
## Motivation and Context

Fixes #529 
## How Has This Been Tested?

Tested through resizing page and looking at perf log.
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
@@ -208,7 +208,10 @@ namespace AppUIBasics.ControlPages
 
         private void Editor_TextChanged(object sender, RoutedEventArgs e)
         {
-            editor.Document.Selection.CharacterFormat.ForegroundColor = currentColor;
+            if(editor.Document.Selection.CharacterFormat.ForegroundColor != currentColor)
+            {
+                editor.Document.Selection.CharacterFormat.ForegroundColor = currentColor;
+            }
         }
     }
 }
--- a/XamlControlsGallery/ControlPages/SplitButtonPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/SplitButtonPage.xaml.cs
@@ -50,7 +50,10 @@ namespace AppUIBasics.ControlPages
 
         private void MyRichEditBox_TextChanged(object sender, RoutedEventArgs e)
         {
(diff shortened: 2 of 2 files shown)


teaP:
Isn't it also really weird that setting the color to something it already is does.... anything?

chingucoding:
Yes, that also confuses me here. It's a bit strange that manually checking if we actually modify the value prevents us from having such a high CPU usage. I would have expected the control, document or selection object on the document to skip that step if the old value is the same as the new value.

stmoy:
@ranjeshj @chrisglein - it smells to me like there is a bug in the text control; what's the best way to handle this? I think we can take this PR in the short term, but how should we track this bug in the control itself?

chrisglein:
> @ranjeshj @chrisglein - it smells to me like there is a bug in the text control; what's the best way to handle this? I think we can take this PR in the short term, but how should we track this bug in the control itself?

Given the fix yeah this seems like a bug. Setting a property value to the same value it already was should not cause invalidations or re-renders. My guess is that the change notifications happening on the text services RichEditBox.Document object (Or just the [ITextCharacterFormat](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Text.ITextCharacterFormat?view=winrt-19041) is very course and doesn't actually check if the value has really changed. Probably a bug in richedit, but the first step would be to file an issue on WinUI and link this PR.

chingucoding:
> Given the fix yeah this seems like a bug. Setting a property value to the same value it already was should not cause invalidations or re-renders. My guess is that the change notifications happening on the text services RichEditBox.Document object (Or just the ITextCharacterFormat is very course and doesn't actually check if the value has really changed. Probably a bug in richedit, but the first step would be to file an issue on WinUI and link this PR.

Created an issue over at WinUI for this: https://github.com/microsoft/microsoft-ui-xaml/issues/3225

stmoy (approved):

PR #540, opened 2020-08-23
Fix crash with itemsrepeater and end key in color demo

chingucoding:
## Description

The problem is that with end key being hit, we scroll a lot very fast resulting in some events saying "Render item x", our code saying "Awesome item x selected, let's quickly get that" and ItemsRepeater being "Item x? I'v already scrolled to item x+5000, so here is null".
## Motivation and Context

Fixes the following issue at the WinUI repo: https://github.com/microsoft/microsoft-ui-xaml/issues/2107
## How Has This Been Tested?

Tested manually by pressing end key.
## Types of changes


ranjeshj (approved):

anawishnoff:
@chingucoding We'd like to merge this in but it looks like there's a test failing. Would you be able to pull from master and push again?

chingucoding:
The CI pipeline wasn't too reliable in the last few months and I think it was disabled now, but maybe @stmoy knows more about this. Either way, you should be able to merge it now.

PR #565, opened 2020-09-29
Update AutoSuggestBox sample to be more illustrative

### LINKED ISSUE #509: Doc Alignment: Update AutoSuggestBox Sample
**Describe the bug**

The current AutoSuggestBox page has a sample where the control simply appends "1" and "2" to the query. This isn't correct guidance, and it also doesn't show the scenario of "No Results Found". Update the ASB page to search from a list rather than modifying the results based on the query. 

https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/auto-suggest-box#dos-and-donts

chingucoding:
## Description

The AutoSuggestBox sample now searches a wonderful list of cats and shows "No items found" when there was not a match to align with the official AutoSuggestBox docs.

## Motivation and Context

Closes #509 
## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/AutoSuggestBoxPage.xaml
+++ b/XamlControlsGallery/ControlPages/AutoSuggestBoxPage.xaml
@@ -1,4 +1,4 @@
-﻿<Page x:Name="pageRoot" x:Class="AppUIBasics.ControlPages.AutoSuggestBoxPage"
+<Page x:Name="pageRoot" x:Class="AppUIBasics.ControlPages.AutoSuggestBoxPage"
         xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:local="using:AppUIBasics"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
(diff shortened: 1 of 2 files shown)


XamlControlsGallery/ControlPages/AutoSuggestBoxPage.xaml.cs
```diff
@@ -21,23 +21,138 @@ namespace AppUIBasics.ControlPages
...
+                    return cat.ToLower().Contains(key);
+                });
+                if(found)
+                {
+                    suitableItems.Add(cat);
+                }
+            }
+            if(suitableItems.Count == 0)
+            {
+                suitableItems.Add("No items found");
```
stmoy, line 151: Double checking: I would expect this text would be handled by AutoSuggestBox (so that it can be consistent, localizable, etc.) - is that not the case? If not and we need to supply our own text, please use "No results found", as this is what the main search box inside the Gallery reports.

chingucoding: Unfortunately, this is something we need to handle. If the collection of items is empty, the AutoSuggestBox just shows no flyout at all. Updated the wording now to "No results found".

XamlControlsGallery/ControlPages/AutoSuggestBoxPage.xaml
```diff
@@ -19,6 +19,47 @@
...
+            return cat.ToLower().Contains(key);
+        });
+        if(found)
+        {
+            suitableItems.Add(cat);
+        }
+    }
+    if(suitableItems.Count == 0)
+    {
+        suitableItems.Add("No items found");
```
stmoy, line 51: Please update the text here as well.

chingucoding: Always forget to update that too, fixed now.

stmoy (approved):
Thank you for these updates!

PR #566, opened 2020-10-01
Add button wrapping sample

### LINKED ISSUE #508: Doc Alignment: Button Wrapping Example
**Describe the bug**

Doc alignment: add a sample to Button page that shows the recommended wrapping guidance. 

https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/buttons#button-content

chingucoding:
## Description

This adds a sample to the button page showing how to handle large text inside buttons.
## Motivation and Context

Closes #508 
## How Has This Been Tested?

Tested manually.
image

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ButtonPage.xaml
+++ b/XamlControlsGallery/ControlPages/ButtonPage.xaml
@@ -60,12 +60,12 @@
         </local:ControlExample>
         <local:ControlExample HeaderText="Wrapping Buttons with large content">
             <StackPanel MaxWidth="500">
-                <TextBlock Text="The following buttons have content that would get clipped if we are not wrapping the buttons properly." Margin="0,0,0,8" TextWrapping="Wrap"/>
-                <TextBlock Text="The first option to mitigate cut of content is to place them underneath each other:" Margin="0,0,0,8" TextWrapping="Wrap"/>
+                <TextBlock Text="The following buttons' content may get clipped if we don't pay careful attention to their layout containers." Margin="0,0,0,8" TextWrapping="Wrap"/>
+                <TextBlock Text="One option to mitigate clipped content is to place Buttons underneath each other, allowing for more space to grow horizontally:" Margin="0,0,0,8" TextWrapping="Wrap"/>
                 <Button HorizontalAlignment="Stretch" Margin="0,0,0,5">This is some text that is too long and will get cut off</Button>
                 <Button HorizontalAlignment="Stretch">This is another text that would result in being cut off</Button>
 
-                <TextBlock Text="Another option is to increase the height of the buttons:" Margin="0,8,0,8"/>
+                <TextBlock Text="Another option is to explicitly wrap the Button's content" Margin="0,8,0,8"/>
                 <StackPanel Orientation="Horizontal" HorizontalAlignment="Center">
                     <Button MaxWidth="240" Margin="0,0,8,0" Height="60">
                         <TextBlock Text="This is some text that is too long and will get cut off" TextWrapping="WrapWholeWords"/>
--- a/XamlControlsGallery/ControlPages/ButtonPage.xaml
+++ b/XamlControlsGallery/ControlPages/ButtonPage.xaml
@@ -58,6 +58,45 @@
                 </x:String>
             </local:ControlExample.Xaml>
         </local:ControlExample>
+        <local:ControlExample HeaderText="Wrapping Buttons with large content">
+            <StackPanel MaxWidth="500">
+                <TextBlock Text="The following buttons have content that would get clipped if we are not wrapping the buttons properly." Margin="0,0,0,8" TextWrapping="Wrap"/>
+                <TextBlock Text="The first option to mitigate cut of content is to place them underneath each other:" Margin="0,0,0,8" TextWrapping="Wrap"/>
+                <Button HorizontalAlignment="Stretch" Margin="0,0,0,5">This is some text that is too long and will get cut off</Button>
+                <Button HorizontalAlignment="Stretch">This is another text that would result in being cut off</Button>
+
+                <TextBlock Text="Another option is to increase the height of the buttons:" Margin="0,8,0,8"/>
+                <StackPanel Orientation="Horizontal" HorizontalAlignment="Center">
+                    <Button MaxWidth="240" Margin="0,0,8,0" Height="60">
+                        <TextBlock Text="This is some text that is too long and will get cut off" TextWrapping="WrapWholeWords"/>
+                    </Button>
+                    <Button MaxWidth="240" Height="60">
+                        <TextBlock Text="This is another text that would result in being cut off" TextWrapping="WrapWholeWords"/>
+                    </Button>
+                </StackPanel>
+            </StackPanel>
+            <local:ControlExample.Xaml>
+                <x:String xml:space="preserve">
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -58,6 +58,45 @@
...
+        <local:ControlExample HeaderText="Wrapping Buttons with large content">
+            <StackPanel MaxWidth="500">
+                <TextBlock Text="The following buttons have content that would get clipped if we are not wrapping the buttons properly." Margin="0,0,0,8" TextWrapping="Wrap"/>
+                <TextBlock Text="The first option to mitigate cut of content is to place them underneath each other:" Margin="0,0,0,8" TextWrapping="Wrap"/>
+                <Button HorizontalAlignment="Stretch" Margin="0,0,0,5">This is some text that is too long and will get cut off</Button>
+                <Button HorizontalAlignment="Stretch">This is another text that would result in being cut off</Button>
+
+                <TextBlock Text="Another option is to increase the height of the buttons:" Margin="0,8,0,8"/>
+                <StackPanel Orientation="Horizontal" HorizontalAlignment="Center">
+                    <Button MaxWidth="240" Margin="0,0,8,0" Height="60">
```
YuliKl, line 70: You actually don't want to hardcode the Height. With large, this will still get cut off:
image

If you need a height, use MinHeight. Or you can simply leave off that property and the Button (inside a Horizontal StackPanel, at least) will adjust automatically.

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -58,6 +58,45 @@
                 </x:String>
             </local:ControlExample.Xaml>
         </local:ControlExample>
+        <local:ControlExample HeaderText="Wrapping Buttons with large content">
+            <StackPanel MaxWidth="500">
```
YuliKl, line 62: You don't need to specify this MaxWidth, either. It actually causes problems with large text:
image

If you simply let the StackPanel manage its own width automatically, things look much better.

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -58,6 +58,45 @@
...
+                <TextBlock Text="The following buttons have content that would get clipped if we are not wrapping the buttons properly." Margin="0,0,0,8" TextWrapping="Wrap"/>
```
YuliKl, line 63:
```suggestion
                <TextBlock Text="The following buttons' content may get clipped if we don't pay careful attention to their layout containers." Margin="0,0,0,8" TextWrapping="Wrap"/>
```

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -58,6 +58,45 @@
...
+                <TextBlock Text="The first option to mitigate cut of content is to place them underneath each other:" Margin="0,0,0,8" TextWrapping="Wrap"/>
```
YuliKl, line 64:
```suggestion
                <TextBlock Text="One option to mitigate clipped content is to place Buttons underneath each other, allowing for more space to grow horizontally:" Margin="0,0,0,8" TextWrapping="Wrap"/>
```

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -58,6 +58,45 @@
...
+                <Button HorizontalAlignment="Stretch" Margin="0,0,0,5">This is some text that is too long and will get cut off</Button>
+                <Button HorizontalAlignment="Stretch">This is another text that would result in being cut off</Button>
+
+                <TextBlock Text="Another option is to increase the height of the buttons:" Margin="0,8,0,8"/>
```
YuliKl, line 68:
```suggestion
                <TextBlock Text="Another option is to explicitly wrap the Button's content" Margin="0,8,0,8"/>
```

XamlControlsGallery/ControlPages/ButtonPage.xaml
chingucoding: Great suggestion, thank you! Fixed now.

XamlControlsGallery/ControlPages/ButtonPage.xaml
chingucoding: Same here, great suggestion, fixed that now.

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -58,6 +58,45 @@
...
+                <x:String xml:space="preserve">
+&lt;StackPanel MaxWidth="500"&gt;
+    &lt;TextBlock Text="The following buttons have content that would get clipped if we are not wrapping the buttons properly." Margin="0,0,0,8" TextWrapping="Wrap"/&gt;
+    &lt;TextBlock Text="The first option to mitigate cut of content is to place them underneath each other:" Margin="0,0,0,8" TextWrapping="Wrap"/&gt;
+    &lt;Button HorizontalAlignment="Stretch" Margin="0,0,0,5"&gt;This is some text that is too long and will get cut off&lt;/Button&gt;
+    &lt;Button HorizontalAlignment="Stretch"&gt;This is another text that would result in being cut off&lt;/Button&gt;
+
+    &lt;TextBlock Text="Another option is to increase the height of the buttons:" Margin="0,8,0,8"/&gt;
+    &lt;StackPanel Orientation="Horizontal" HorizontalAlignment="Center"&gt;
+        &lt;Button MaxWidth="240" Margin="0,0,8,0" Height="60"&gt;
```
YuliKl, line 88: Don't forget to update the sample code :)

chingucoding: Very good point! I think I updated the sample correctly now.

XamlControlsGallery/ControlPages/ButtonPage.xaml
```diff
@@ -58,6 +58,45 @@
...
+                    <Button MaxWidth="240">
+                        <TextBlock Text="This is another text that would result in being cut off" TextWrapping="WrapWholeWords"/>
+                    </Button>
+                </StackPanel>
+            </StackPanel>
+            <local:ControlExample.Xaml>
+                <x:String xml:space="preserve">
+&lt;StackPanel&gt;
+    &lt;TextBlock Text="The following buttons' content may get clipped if we don't pay careful attention to their layout containers." Margin="0,0,0,8" TextWrapping="Wrap"/&gt;
+    &lt;TextBlock Text="One option to mitigate clipped content is to place Buttons underneath each other, allowing for more space to grow horizontally:" Margin="0,0,0,8" TextWrapping="Wrap"/&gt;
```
YuliKl, line 82: Actually, we probably don't need this description text to show up in the sample code.

Probably best to just isolate the two Button lines inside their StackPanels and leave it at that.

YuliKl (approved):

XamlControlsGallery/ControlPages/ButtonPage.xaml
chingucoding: Hmm good point. But how easy is it for users to understand what the sample code is about if it just shows a few buttons in StackPanels? With the description in there, it's clear what the purpose of those specific StackPanels and buttons is. Though the actual sample is directly above it, so this might not be an issue after all.

Issue #570, opened 2020-10-09
Add ARM64 Support


stmoy:
ARM64 is an increasingly important endpoint. This issue tracks the work required to make the Gallery run on ARM64.

Some potential "gotchas":
- XCG depends on some packages that may or may not work on ARM64
- How well does the toolchain support ARM64?

chingucoding:
Regarding the dependencies:
- WinUI: Supports ARM64
- ColorCode.Uwp: Probably supports it?
- Win2D.Uwp: Supports ARM64 with version 1.24.0, we are currently using 1.23.0
- Microsoft.NETCore.UniversalWindowsPlatform: Added support with 6.2.0-Preview1-26502-02, we are currently using 6.1.9
- Microsoft.Xaml.Behaviors.Uwp.Managed: Does NOT support ARM64 right now

So ARM64 is currently blocked on the behaviors SDK not supporting ARM64.

stmoy:
Thanks for this investigation @chingucoding! I wonder if we could see who owns Behaviors and see if they can upgrade to ARM64? CC: @michael-hawker do you recall who owns Behaviors?

Alternatively we could look at removing the Behaviors dependency. Any thoughts on how tightly it's coupled?

chingucoding:
The only reasone we use the behaviors library is to animate the header parallaxing on the "All controls page" and the "What's new page". Beside that, it doesn't seem to be used anywhere else in the XCG. I'm note quite sure how easy it would be to implement the   header scrolling behavior without that library though,

michael-hawker:
@azchohfi is working on the ARM64 story as well, I believe he's been working with the Behaviors folks as well already on this.

ColorCode should be working as we use it in the toolkit and we made the toolkit support ARM64. 🙂

azchohfi:
ARM64 support have been added, we just haven't shipped a new version yet.
https://github.com/microsoft/XamlBehaviors/pull/190

azchohfi:
Btw, Microsoft.Xaml.Behaviors.Uwp.**Managed** already supports ARM64.
We just shipped a new version of Microsoft.Xaml.Behaviors.Uwp.**Native** that supports ARM64:
https://www.nuget.org/packages/Microsoft.Xaml.Behaviors.Uwp.Native/1.0.8

PR #587, opened 2020-11-04
[winui2] Update project files to be ready for ARM64

chingucoding:
## Description

Updated the packages to use the latest versions and have ARM64 compatible versions as such.
Also added ARM64 as configuration to the project and solution.
## Motivation and Context

First step in adding ARM64 support, see #570 
## How Has This Been Tested?

Built project in ARM64 however not tested on an arm64 device.
## Types of changes


Poopooracoocoo:
anything blocking this??

bpulliam:
I'll leave the same comment on both arm64 updates: given the age of the PR, either resolve the merge conflicts with the current HEAD of winui2 or create a new PR.  We should support this, and either method gets us towards that goal.

bpulliam (changes requested):
I'll leave the same comment on both arm64 updates: given the age of the PR, either resolve the merge conflicts with the current HEAD of winui2 or create a new PR.  We should support this, and either method gets us towards that goal.

PR #596, opened 2020-11-17
Added determinate mode for progress ring

MarissaMatt:
Added determinate mode example to the progress ring page. 

## Description
Added an example of how to set the IsIndeterminate property to change the progress ring in to a determinate mode progress ring that fills based on the value property. 

## How Has This Been Tested?
Ran on my local machine using the 2.5 pre-release package. 

## Types of changes
New feature

### DIFF
--- a/XamlControlsGallery/ControlPages/ProgressRingPage.xaml
+++ b/XamlControlsGallery/ControlPages/ProgressRingPage.xaml
@@ -20,7 +20,7 @@
     mc:Ignorable="d">
     <StackPanel>
         <local:ControlExample HeaderText="An indeterminate progress ring.">
-            <muxc:ProgressRing IsActive="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" Height="130" Width="130" VerticalAlignment="Top" Margin="10,10,0,0"/>
+            <muxc:ProgressRing IsActive="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" Height="60" Width="60" VerticalAlignment="Top" Margin="10,10,0,0"/>
             <local:ControlExample.Options>
                 <StackPanel>
                     <ToggleSwitch x:Name="ProgressToggle" Header="Toggle work" OffContent="Do work" OnContent="Working" IsOn="True" />
@@ -37,7 +37,7 @@
         </local:ControlExample>
         <local:ControlExample HeaderText="A determinate progress ring.">
             <StackPanel x:Name="Control2" Orientation="Horizontal">
-                <muxc:ProgressRing Width="130" x:Name="ProgressRing2" IsIndeterminate="False"/>
+                <muxc:ProgressRing Width="60" x:Name="ProgressRing2" IsIndeterminate="False"/>
                 <TextBlock x:Name="Control2Output" Style="{ThemeResource OutputTextBlockStyle}" Width="60" TextAlignment="Center" />
                 <TextBlock x:Name="ProgressLabel"  Text="Progress" VerticalAlignment="Center" Margin="0,0,10,0"/>
                 <muxc:NumberBox x:Name="ProgressValue" AutomationProperties.LabeledBy="{Binding ElementName=ProgressLabel}" Minimum="0" Maximum="100" SpinButtonPlacementMode="Inline" ValueChanged="ProgressValue_ValueChanged" VerticalAlignment="Center"/>
--- a/XamlControlsGallery/ControlPages/ProgressRingPage.xaml
+++ b/XamlControlsGallery/ControlPages/ProgressRingPage.xaml
@@ -20,10 +20,10 @@
     mc:Ignorable="d">
     <StackPanel>
         <local:ControlExample HeaderText="An indeterminate progress ring.">
-            <muxc:ProgressRing IsActive="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" Height="60" Width="60" VerticalAlignment="Top" Margin="10,10,0,0"/>
+            <muxc:ProgressRing IsActive="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" AutomationProperties.Name="Progress image" Height="60" Width="60" VerticalAlignment="Top" Margin="10,10,0,0"/>
             <local:ControlExample.Options>
                 <StackPanel>
-                    <ToggleSwitch x:Name="ProgressToggle" Header="Toggle work" OffContent="Do work" OnContent="Working" IsOn="True" />
+                    <ToggleSwitch x:Name="ProgressToggle" AutomationProperties.Name="Progress options" Header="Toggle work" OffContent="Do work" OnContent="Working" IsOn="True" />
                 </StackPanel>
             </local:ControlExample.Options>
             <local:ControlExample.Xaml>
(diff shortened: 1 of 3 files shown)


XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -38,6 +38,15 @@
                 <local:ControlExampleSubstitution Key="IsActive" Value="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" />
             </local:ControlExample.Substitutions>
 
+        </local:ControlExample>
```
chingucoding, line 41: I think parts of the sample is missing, namely the sample and the options for said sample. This XAML wouldn't compile.

MarissaMatt: @chingucoding that's weird I tried adding it again is it compiling this time? This is also using the 2.5 pre-release package.

chingucoding: It's compiling fine now and works as expected, thank you!

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -20,15 +20,12 @@
     mc:Ignorable="d">
     <StackPanel>
         <local:ControlExample HeaderText="An indeterminate progress ring.">
-
-            <muxc:ProgressRing IsActive="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" Height="60" Width="60" VerticalAlignment="Top" Margin="10,10,0,0"/>
-
+            <muxc:ProgressRing IsActive="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" Height="130" Width="130" VerticalAlignment="Top" Margin="10,10,0,0"/>
```
chingucoding, line 23: Height and width of 130 looks a bit out of place in my opinion, maybe we could choose a smaller size?

YuliKl: I tend to agree. This feels a little big, and I'd like to see us stay on 4px grid. So 120 would already be a better size from that perspective.

MarissaMatt: ok no problem I can switch it back to 60.

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -37,7 +34,22 @@
             <local:ControlExample.Substitutions>
                 <local:ControlExampleSubstitution Key="IsActive" Value="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" />
             </local:ControlExample.Substitutions>
-
+        </local:ControlExample>
+        <local:ControlExample HeaderText="A determinate progress ring.">
+            <StackPanel x:Name="Control2" Orientation="Horizontal">
+                <muxc:ProgressRing Width="130" x:Name="ProgressRing2" IsIndeterminate="False"/>
```
YuliKl, line 40: Would also be good to understand what scenarios you expect apps to use a determinate ring for. The scenario should help influence the right size in the sample.

MarissaMatt: Good idea! I'll update the page description so it's similar to progress bar.

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -37,7 +34,22 @@
...
+                <TextBlock x:Name="Control2Output" Style="{ThemeResource OutputTextBlockStyle}" Width="60" TextAlignment="Center" />
+                <TextBlock x:Name="ProgressLabel"  Text="Progress" VerticalAlignment="Center" Margin="0,0,10,0"/>
```
YuliKl, line 42: NumberBox has a header property. Maybe use that instead of this ProgressLabel?

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -37,7 +34,21 @@
             <local:ControlExample.Substitutions>
                 <local:ControlExampleSubstitution Key="IsActive" Value="{x:Bind ProgressToggle.IsOn, Mode=OneWay}" />
             </local:ControlExample.Substitutions>
-
+        </local:ControlExample>
+        <local:ControlExample HeaderText="A determinate progress ring.">
```
YuliKl, line 38: I recommend running the Accessibility Insights app on your changes before submitting a PR. It's showing that both the ProgressRing and the NumberBox are missing automation names.

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -37,7 +34,21 @@
...
+            <StackPanel x:Name="Control2" Orientation="Horizontal">
+                <muxc:ProgressRing Width="60" x:Name="ProgressRing2" IsIndeterminate="False"/>
+                <TextBlock x:Name="Control2Output" Style="{ThemeResource OutputTextBlockStyle}" Width="60" TextAlignment="Center" />
```
YuliKl, line 41: Is this TextBlock actually used? I'm not seeing any control output...

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -37,7 +34,21 @@
...
+                <muxc:NumberBox x:Name="ProgressValue" AutomationProperties.LabeledBy="{Binding ElementName=ProgressLabel}" Minimum="0" Maximum="100" SpinButtonPlacementMode="Inline" Header="Progress" ValueChanged="ProgressValue_ValueChanged" VerticalAlignment="Center"/>
```
YuliKl, line 42: nit: this line is starting to get really long. Please wrap to make it easier on reviewers. Scrolling in GitHub is no fun ;)

YuliKl, line 42: There's a known NumberBox bug that it's not announcing its header as its UAI name. The LabeledBy bit would have gotten around that bug...but you got rid of ProgressLabel on my recommendation. So I suggest 

```suggestion
                <muxc:NumberBox x:Name="ProgressValue" AutomationProperties.Name="Progress amount" Minimum="0" Maximum="100" SpinButtonPlacementMode="Inline" Header="Progress" ValueChanged="ProgressValue_ValueChanged" VerticalAlignment="Center"/>
```

YuliKl, line 42: I had logged another NumberBox bug where specifying no value results in the flippers being disabled. I find this very annoying. Adding Value="0" works around this problem.

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -37,7 +34,21 @@
...
+            </StackPanel>
+            <local:ControlExample.Xaml>
+                <x:String>
+                    &lt;muxc:ProgressRing Width="130" Value="$(DeterminateProgressValue)" /&gt;
```
YuliKl, line 46: Missing the crucial "IsIndeterminate="False"" bit :)

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
MarissaMatt: It's separating the NumberBox from the ring.

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
MarissaMatt: Is there a way to fix that? I followed the sample from ProgressBar.

YuliKl: For every control, you can always set AutomationProperties.Name="something useful" (where the name, ideally, is not just the control name repeated)

XamlControlsGallery/ControlPages/ProgressRingPage.xaml
YuliKl: In that case, you just want a margin. Delete the TextBlock and update the ProgressRing to something like

```
                <muxc:ProgressRing Width="60" x:Name="ProgressRing2" IsIndeterminate="False" Margin="0,0,60,0"/>
```

MarissaMatt:
@YuliKl I updated everything you gave feedback on above.

YuliKl (approved):

stmoy (approved):

PR #603, opened 2020-11-28
Fixed incorrect wording in description of TimePicker control

r2d2rigo:
## Description
The description text on how to use the `TimePicker` control shows the following wrong text:

> Use a TimePicker to let users set a time in your app, for example to set a reminder. The TimePicker displays three controls for **month, day, and** and AM/PM.

It seems to be an oversight when copying the same description from the `DateTime` control description.

## Motivation and Context
This fixes a minor help text issue that may confuse new users of the app.

## How Has This Been Tested?
Untested since it's a wording change.

## Types of changes


stmoy (approved):
Thank you for fixing this! 🎉

mdtauk:
In the UK, locale time is 24 hours, which doesn't display the AM/PM designator.

Not sure if it's worth including that fact in the description however

Issue #618, opened 2020-12-27
suggestion: show interaction possibilities between controls


Jay-o-Way:
**Is this a totally new sample, an addition to an existing sample page, or a modification of an existing sample page?**
- [ ] New sample
- [ ] Addition to existing sample
- [x] Modification to existing sample

**If this is an addition/modification to an existing sample page, which one?**

The page that shows the TOGGLE SWITCH.

**Describe the sample**

This page shows two simple toggle switches. The second one determines the state of a connected ProgressRing. On the right side, there are a few links to some other "related controls". The ProgressRing is not amongst them. Since the direct relation (the IsActive property with a bind to the toggle) is _not_ shown on the page (or the code) in the Controls Gallery app, it would be nice to have a link to the page for the ProgressRing, hoping to find a code example that shows how to use the `IsActive="{x:Bind ToggleSwitch2.IsOn, Mode=OneWay}"` function.

**Is your sample request related to a problem? Please describe it**

Nope.

**Additional context**

For people that are still new to this coding, it would be interesting to see how two different controls can interact with each other. In this case, the `IsActive` property can be an example for methods.

chingucoding:
I'm not entirely sure if I understand this correctly. Would it help you if the toggle switch sample code to include the binding between the ToggleSwitch and the ProgressRing?

@stmoy Should we update the sample to use the new WinUI ProgressRing?

Jay-o-Way:
> I'm not entirely sure if I understand this correctly.

I was afraid of that. Here's a quick drawing.
![image](https://user-images.githubusercontent.com/65828559/103181302-06bdc780-48a0-11eb-80e3-bbe11d56a1a6.png)
Since the ProgressRing is _used_ in an example, I'm thinking it might be handy to have a direct link to that control, right next to it.

> Would it help you if the toggle switch sample code to include the binding between the ToggleSwitch and the ProgressRing?

Yes, that too. I'm still pretty new to this and I think this would show the relation between these two controls. You can find it in the [online source code](https://github.com/microsoft/Xaml-Controls-Gallery/blob/master/XamlControlsGallery/ControlPages/ToggleSwitchPage.xaml), but including it in the example code in the app would be handy.

chingucoding:
[...]

PR #619, opened 2020-12-28
Update ToggleSwitchPage.xaml

Jay-o-Way:
## Description

Included the ProgressRing (and the Stackpanel) in the example code.

## Motivation and Context

Addressing issue #618
The example shows a ProgressRing (next to the ToggleSwitch) and I thought it would be nice to include this in the example code, so that people who are new to code (like me) can see how the ToggleSwitch influences the ProgressRing.

## How Has This Been Tested?

Locally with visual studio.

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ToggleSwitchPage.xaml
+++ b/XamlControlsGallery/ControlPages/ToggleSwitchPage.xaml
@@ -35,9 +35,13 @@
             </StackPanel>
             <local:ControlExample.Xaml>
                 <x:String>
-                    &lt;ToggleSwitch Header="Toggle work" OffContent="Do work" OnContent="Working" IsOn="$(isOn)$(isOff)" /&gt;
+                    &lt;StackPanel Orientation=&quot;Horizontal&quot;&gt;
+                    &lt;ToggleSwitch Header=&quot;Toggle work&quot; OffContent=&quot;Do work&quot; OnContent=&quot;Working&quot; IsOn=&quot;$(isOn)$(isOff)&quot; /&gt;
+                    &lt;ProgressRing IsActive=&quot;{x:Bind ToggleSwitch2.IsOn, Mode=OneWay}&quot; Width=&quot;130&quot;/&gt;
+                    &lt;/StackPanel&gt;
                 </x:String>
             </local:ControlExample.Xaml>
+
             <local:ControlExample.Substitutions>
                 <local:ControlExampleSubstitution Key="isOn" IsEnabled="{x:Bind ToggleSwitch2.IsOn, Mode=OneWay}"
                                                   Value= "True"/>
--- a/XamlControlsGallery/ControlPages/ToggleSwitchPage.xaml
+++ b/XamlControlsGallery/ControlPages/ToggleSwitchPage.xaml
@@ -34,11 +34,11 @@
                 <ProgressRing IsActive="{x:Bind ToggleSwitch2.IsOn, Mode=OneWay}" Width="130"/>
             </StackPanel>
             <local:ControlExample.Xaml>
-                <x:String>
-                    &lt;StackPanel Orientation=&quot;Horizontal&quot;&gt;
-                    &lt;ToggleSwitch Header=&quot;Toggle work&quot; OffContent=&quot;Do work&quot; OnContent=&quot;Working&quot; IsOn=&quot;$(isOn)$(isOff)&quot; /&gt;
-                    &lt;ProgressRing IsActive=&quot;{x:Bind ToggleSwitch2.IsOn, Mode=OneWay}&quot; Width=&quot;130&quot;/&gt;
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/ToggleSwitchPage.xaml
```diff
@@ -35,9 +35,13 @@
             </StackPanel>
             <local:ControlExample.Xaml>
                 <x:String>
-                    &lt;ToggleSwitch Header="Toggle work" OffContent="Do work" OnContent="Working" IsOn="$(isOn)$(isOff)" /&gt;
+                    &lt;StackPanel Orientation=&quot;Horizontal&quot;&gt;
+                    &lt;ToggleSwitch Header=&quot;Toggle work&quot; OffContent=&quot;Do work&quot; OnContent=&quot;Working&quot; IsOn=&quot;$(isOn)$(isOff)&quot; /&gt;
+                    &lt;ProgressRing IsActive=&quot;{x:Bind ToggleSwitch2.IsOn, Mode=OneWay}&quot; Width=&quot;130&quot;/&gt;
+                    &lt;/StackPanel&gt;
```
chingucoding, line 41: The code sample is not correctly formatted in the app: 
Image shown ill formatted code

You can get the code to be correctly formatted by adding `xml:space="preserve"` to the x:String in line 37 and manually formatting the text then. You can find an example how to "format" this here: https://github.com/microsoft/Xaml-Controls-Gallery/blob/7faea5a5cb74a9b9db7c6a7d291f9b890ce06420/XamlControlsGallery/ControlPages/AppBarSeparatorPage.xaml#L33-L41

The only important thing is adding the xml:preserver and that the formatted code then starts at the left. Thank you for your PR!

chingucoding:
@Jay-o-Way If you think that this PR fixes issue #618 sufficiently, you can add "Fix #618" to your PR description. That way, when this PR gets accepted and merged, GitHub will automatigically close issue #618 for you 😄

chingucoding (approved):
Looks good to me, thank you!

XamlControlsGallery/ControlPages/ToggleSwitchPage.xaml
Jay-o-Way: I couldn't really see it as it would be in the app, because there stil was a little issue when I tried to debug/run it here.
This doesn't address the "Related controls" section, but that's a different subject and I'll leave that up to the colleague you already mentioned.

chingucoding: Oh hmm, you shouldn't have problems running the app when freshly cloned. Do you remember what the issue was, maybe we can fix it or find a workaround.

Jay-o-Way: The problem was missing (.exe) files in a folder
image

All the folders `Debug` and `Release` (inside `ARM`, `x64`, and `x86`) are empty. Probably just a "local" cause?

chingucoding: That's weird. Have you tried rebuilding the solution? If you try to try to debug the app on your machine, it should build the app; maybe that didn't work correctly? Though I have to say that I haven't seen this message before, the fact that the output folders seem to be empty is definitely unusual.

Jay-o-Way: LOL, I was using the thing on the single xaml file instead of the project. My bad. Never mind 🤐 `#StillLearning`

Jay-o-Way:
Any news?

stmoy (approved):
Thank you for submitting this change!

Jay-o-Way:
@stmoy Just curious, but do you have an idea when the Windows app will be updated? It shows no update when I look in Microsoft Store and the information there says "Release date 9-10-2017" and using "WinUI version 2.2". 🤷🏻‍♂️

stmoy:
Whoops. I suspect that the Store page is out of date - the most recent version was published Dec 04 2020 and was updated with the latest version of WinUI 2.5. The Settings page in the app itself has up-to-date information, can you please double check there as well?

We generally push new versions of the app to the store to coincide with WinUI releases. Since this change came after the 2.5 release in December, that's probably why the change isn't shown in the app yet.

Jay-o-Way:
> The Settings page in the app itself has up-to-date information, can you please double check there as well?

It says version **1.2.17.0**. Is that good?

stmoy:
Yup, that's the version of the app itself. The WinUI version isn't highlighted on the Settings page, but that could be a good thing to highlight separately. 

I'm not sure how to update the release date - I think that was the original "first time submitted to the Store" date.

Poopooracoocoo:
@stmoy could you please provide all of this info in the app and in the ms store? it's so hard to determine whether XCG has received an update or not.

PR #624, opened 2021-01-07
Add support to samples to "copy link to this page" with appropriate protocol activation parameters

### LINKED ISSUE #623: Add support to samples to "copy link to this page" with appropriate protocol activation parameters
As far as I can tell, the only way to generate a deep link into a specific page is to construct the protocol activation URI manually. It would be helpful to be able to quickly generate links to specific pages within the app and add them to your clipboard. This would make it easier to reference specific samples in emails, chat, support tickets, issues, documentation, etc.

There are a number of options that come to mind (could implement one or more):
* Add context menu on NavigationViewItems
* Add button near existing "Toggle Theme" button on each page
* <kbd>Ctrl</kbd> + <kbd>C</kbd> on any page (when text is not selected) could copy URI to clipboard

grochocki:
## Description
* Adds support to copy links to clipboard for sample groups and items
  * Added context menu to NavigationViewItems
  * Added button to ItemsPage
* Fixes issue in protocol activation preventing html links that strips one of the slashes ("///" -> "//"), which some applications, like Outlook, does automatically. These links will now launch properly.
* Fixes issue that was preventing NavigationView visuals from matching navigation due to protocol launch on first launch.

## Motivation and Context
It would be helpful to be able to quickly generate links to specific pages within the app and add them to your clipboard. This would make it easier to reference specific samples in emails, chat, support tickets, issues, documentation, etc.

Fixes #623 

## How Has This Been Tested?
For each of:
* What's New
* All Controls
* At least one sample group (e.g., Basic Input)
* At least one sample item (e.g., Button)

I manually tested:
* Generating a link on right-click
* Generating a link via sample page (sample items only)
* Pasting into Notepad (plaintext)
* Pasting into Word (html)
* Protocol launch with app closed
* Protocol launch with app open

Context Menu
image

Button
image

Paste Results:
image

## Types of changes


stmoy:
Love this change overall - thanks @grochocki. One specific concern I have is that it isn't immediately obvious what to do with the output of the copied URI. In the plain-text case, for example, the user will need to enter that text into the Run dialog in order for it to do anything useful. This is mostly mitigated by the HTML-generated text in the word example above, but that depends on the user pasting the link into an app that understands what to do with it.

I think there are a few ways to mitigate this concern:

- Include some text somewhere after the user clicks a link that says "Pasting this link into the Run dialog will open the app to this page" (or something)
- Removing or minimizing the copy button on each page (maybe to just an icon?) 

Open to ideas. Also open to being told that this isn't actually as much of a concern as I think it is 😉

grochocki:
> Also open to being told that this isn't actually as much of a concern as I think it is 😉

I am not (yet) convinced this is that big of a concern. If it were a consumer app, sure, but if we are optimizing for developer audience, I feel like it is fairly low risk.

> Include some text somewhere after the user clicks a link that says "Pasting this link into the Run dialog will open the app to this page" (or something)

I am open to something like this, though, anyway! I will noodle on it a bit.

> Removing or minimizing the copy button on each page (maybe to just an icon?)

I thought about this originally, and just the icon (or the icon + "copy") might imply that you are copying selected content/text. I am open to other suggestions, but I think we may want to make _what_ you are copying explicit in the action. Could reduce slightly to "Copy Sample Link", perhaps?

stmoy:
> > Also open to being told that this isn't actually as much of a concern as I think it is 😉
> 
> I am not (yet) convinced this is that big of a concern. If it were a consumer app, sure, but if we are optimizing for developer audience, I feel like it is fairly low risk.

Fair enough, but it does bear mentioning that this app is also used by folks just learning about Xaml, designers, and is our "booth" app that we use to show off the power of the platform. It's not just devs.
 
> > Include some text somewhere after the user clicks a link that says "Pasting this link into the Run dialog will open the app to this page" (or something)
> 
> I am open to something like this, though, anyway! I will noodle on it a bit.

One idea: this info could be hosted in a flyout that either opens when the user clicks the button, or maybe inside a tooltip if the user hovers over the button.

> > Removing or minimizing the copy button on each page (maybe to just an icon?)
> 
> I thought about this originally, but just the icon (or the icon + "copy") might imply that you are copying selected content/text. I am open to other suggestions, but I think we may want to make _what_ you are copying explicit in the action. Could reduce slightly to "Copy Sample Link", perhaps?

Good point, and I was pondering this as well. What if we swapped from "Copy" to "Link" for both the icon and the verbiage? Something like "Generate link to page"? 

Thinking aloud: what if when you clicked the button, it opens up a flyout or TeachingTip that looks something like the share experience in Word including a link, a button that copies the link to the clipboard, etc.

image

chingucoding:
Re Page link on pages:

What about updating the "page links" section that links to the source code to include that link? They all link to the page one way or another depending on what you are interested in, being able to get the link to the sample page could fit in there.

stmoy:
@Chingucoding

That also seems like a reasonable place for this link as well, though the header will need to be updated from "View page code on GitHub" to something else - but I don't have a good idea off the top of my head.

grochocki:
> Fair enough, but it does bear mentioning that this app is also used by folks just learning about Xaml, designers, and is our "booth" app that we use to show off the power of the platform. It's not just devs.

Great point!

> One idea: this info could be hosted in a flyout that either opens when the user clicks the button, or maybe inside a tooltip if the user hovers over the button.

> Thinking aloud: what if when you clicked the button, it opens up a flyout or TeachingTip that looks something like the share experience in Word including a link, a button that copies the link to the clipboard, etc.

I see a couple possible issues with the flyout approach. First, I would be concerned about it getting in the way of quick copies -- every copy would be two+ clicks. Also, unlike Word, there are not any _other_ sharing options that really necessitate the flyout -- it would just be a surface to put instructional text, which gives me a "not the right tool for the job" kind of feeling.

The tooltip/teaching tip addresses both of those shortcomings and is along the lines of what I was thinking originally. For example, on first click, pop a teaching tip automatically. Subsequent copies would be one click. If you want to see the help text again, hover/click on a little info icon.

Similar to this pattern:
image

> What if we swapped from "Copy" to "Link" for both the icon and the verbiage? Something like "Generate link to page"?

♥

> What about updating the "page links" section that links to the source code to include that link? They all link to the page one way or another depending on what you are interested in, being able to get the link to the sample page could fit in there.

I considered this as well, but all of these links facilitate navigation, per the guidance:
> Only use hyperlinks for navigation; don't use them for other actions.

I'd also be worried about discoverability. Both among a sea of links, but also because the content is shifted to the bottom on all but the largest layouts:
image

chingucoding:
I don't think that opening a flyout to be able to copy the text is the best way to handle this. Copying the URL is a non destructive operation and doesn't need confirmation.  I think a good way would be to have the button copy and then show a ToolTip with "Copied link" or similar which hides after a certain amount of time. TeachingTip is usually quite large and feels a bit overkill if next to the button. Not sure though how a TeachingTip at the bottom would look like, maybe that is another option to confirm the copying.

> I'd also be worried about discoverability. Both among a sea of links, but also because the content is shifted to the bottom on all but the largest layouts:

True though you can still get the link by right clicking on the item in the navigation, though of course people need to know of this option first.

grochocki:
Based on the feedback above, I tried to strike a balance between keeping the copy action quick, while also helping users familiarize themselves with this functionality:
* Updated default string to "Generate Link to Page" and after clicked, updates to "Copied to Clipboard".
* Added Teaching Tip that will always be displayed until the user chooses "Don't show again"
* Teaching Tip can be reset in Settings

Teaching Tip:
image

New "Reset teaching tip" setting:
image

In action:
mVxH70ErB8

XamlControlsGallery/PageHeader.xaml
```diff
@@ -75,6 +75,26 @@
                     IsOpen="False"
                     IsSticky="False"
                     Visibility="Collapsed">
+                    <AppBarButton x:Name="CopyLinkButton" Click="OnCopyLinkButtonClick">
+                        <AppBarButton.Icon>
+                            <FontIcon x:Name="CopyLinkButtonIcon" Margin="0,-2,0,0" Glyph="&#xE71B;" />
```
stmoy, line 80: Could this be replaced with a SymbolIcon with Symbol="Copy"?

XamlControlsGallery/PageHeader.xaml
```diff
@@ -75,6 +75,26 @@
...
+                                x:Name="CopyLinkButtonTeachingTip"
+                                Title="Quickly reference this sample!"
+                                PreferredPlacement="Bottom"
+                                Subtitle="Share with others or paste this link into the Run dialog to open the app to this page directly."
+                                ActionButtonContent="Don't show again"
+                                ActionButtonClick="OnCopyDontShowAgainButtonClick"
+                                CloseButtonContent="Got it!"
+                                Target="{x:Bind CopyLinkButton}">
+                                <muxc:TeachingTip.HeroContent>
+                                    <Image Source="/Assets/CopyLinkTeachingTip.png" />
```
stmoy, line 93: Nice touch!

XamlControlsGallery/PageHeader.xaml
grochocki: The checkmark is part of MDL2, but not part of the Symbol Icon enumeration, so using FontIcon per [guidance](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.symbol?view=winrt-19041#remarks). Could update to a different set of icons, if you think that would be better.

chingucoding: If we want to use the "SymbolIcon" Copy, we can use  glyph `&#xE8C8;` but since copy is available as symbol, we can also use a SymbolIcon here.

grochocki: Are there any technical reasons to prefer SymbolIcon over FontIcon in this case? IMO, the checkmark best conveys the confirmation that the generated link was copied to clipboard, and it aligns with the feedback above (an earlier iteration used the Copy glyph), so if there is not a strong reason for choosing over the other, I think I'd prefer to keep it FontIcon to enable this experience (unless you all disagree, of course, and want to revert back to the original copy UX):

image

image

stmoy: Oh, thank you for clarifying that this is the check glyph. My preference for SymbolIcon is from a readability point of view, not a technical one. I agree that checkmark seems appropriate here.

FWIW, it looks like the check glyph is called "Accept" in the SymbolIcon enumeration - maybe that could be used?

grochocki: > it looks like the check glyph is called "Accept" in the SymbolIcon enumeration

Ha - yep! Inconsistent naming across MDL2 and SymbolIcon enum. Happy to make this change.

stmoy (approved):
Looks good to me! @Chingucoding, do you have any feedback? Thanks for adding this @grochocki - this is great!

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
```diff
@@ -215,6 +260,12 @@ private void OnGamepadAdded(object sender, Gamepad e)
             _isGamePadConnected = Gamepad.Gamepads.Any();
         }
 
+        private void OnNavigationViewControlLoaded(object sender, RoutedEventArgs e)
+        {
+            // Delay necessary to ensure NavigationView visual state can match navigation
+            Task.Delay(500).ContinueWith(_ => this.NavigationViewLoaded?.Invoke(), TaskScheduler.FromCurrentSynchronizationContext());
```
chingucoding, line 266: I'm not sure if this is the best way to ensure this.

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
```diff
@@ -136,16 +135,49 @@ public bool CheckNewControlSelected()
             return _newControlsMenuItem.IsSelected;
         }
 
+        public void EnsureNavigationSelection(string id)
```
chingucoding, line 138: Why do we need this? Are we relying on the selection to get the correct URI?

chingucoding (commented):
Thank you for implementing this, that's really awesome! I've left a few comments, I'm not sure if I correctly understand why there is a lot of code surrounding ensuring selection.

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
grochocki: > Are we relying on the selection to get the correct URI?

No -- we are using the navigation parameters to guarantee that the NavigationView visual state matches the right sample page content that is displayed from protocol activations.

The original protocol activation code has a few issues with handling navigation to ItemPages, particularly on cold launch. Depending on whether you are navigating to a group or an item, you can reliably get into a bad state where the NavigationView visual state does not match the page content that is displayed.

XamlControlsGallery/Navigation/NavigationRootPage.xaml.cs
grochocki: This is needed in conjunction with the change above to prevent a bad visual state where a group is expanded, but no item in the group has the selection visual. There are some reports of NavigationViewItem selection not always working the first time, which is what I was experiencing when testing protocol activation cold launches. It seems to be timing-related, and possibly impacted by the animation to expand the group.

Alternative solutions here might be to explore delaying protocol activation handling or possibly NavigationViewItem creation, but both of those would have an impact on launch perf. This one-time delayed call to EnsureNavigationSelection fixes the visual timing issue on cold launch without impacting launch perf.

chingucoding: Oof I see. Is this purely related to a timing issue or possible a bug with NavigationView (namely https://github.com/microsoft/microsoft-ui-xaml/issues/3149 or https://github.com/microsoft/microsoft-ui-xaml/issues/3015)? If the latter, can you add a comment indicating that or possible create an issue to remove this once it's fixed in WinUI? That would be great!

chingucoding (approved):
Looks good to me; I wish we wouldn't need the ensure selection logic, but it seems it's a limitation/bug in NavigationView so it's fine. Thank you @grochocki for adding these features!

stmoy:
:shipit:

PR #625, opened 2021-01-14
HC changes so controls look more accurate

YuliKl:
The app was showing a few high contrast colors that were muddying up the accuracy of the control's appearance. To make Xaml controls show a more accurate representation of themselves, made the following two changes.

- Changed HighContrastAdjustment to None. 
Before:
HCAdjustment_Before
After: 
HCAdjustment_Updated

- Updated controls examples to stop drawing ButtonFace color.
After:
ExampleControlPage_Updated

## How Has This Been Tested?

Manually verified in regular and high contrast.

## Types of changes

### DIFF
--- a/XamlControlsGallery/AllControlsPage.xaml
+++ b/XamlControlsGallery/AllControlsPage.xaml
@@ -17,7 +17,8 @@
     xmlns:behavior="using:AppUIBasics.Behaviors"
     xmlns:interactivity="using:Microsoft.Xaml.Interactivity"
     xmlns:controls="using:Windows.UI.Xaml.Controls"
-    NavigationCacheMode="Enabled">
+    NavigationCacheMode="Enabled"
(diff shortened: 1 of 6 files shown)


stmoy (approved):

stmoy:
The code looks ok to me, but the first "After" screenshot seems more difficult to read (purple on teal). Is that expected? Is this attempting to demonstrate that the default colors in HC are actually more difficult to read by default?

YuliKl:
> The code looks ok to me, but the first "After" screenshot seems more difficult to read (purple on teal). Is that expected? Is this attempting to demonstrate that the default colors in HC are actually more difficult to read by default?

The purple on teal is coming from my own custom HC style - one where all 8 colors are unique so there's no ambiguity about which brush is being used. So being harder to read is my own fault, and won't impact end-users unless they make equally poor color choices as I did :)

PR #627, opened 2021-01-15
Add PipsPager Sample Pages

gabbybilka:
## Description
Add PipsPager sample, check out [spec PR](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/104) for most up to date information.
image

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/PipsPagerPage.xaml
+++ b/XamlControlsGallery/ControlPages/PipsPagerPage.xaml
@@ -10,9 +10,44 @@
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
 
     <StackPanel>
-        <local:ControlExample x:Name="Example1" HeaderText="PipsPager with options to change its orientation and button visibility.">
+        <local:ControlExample x:Name="Example1" HeaderText="PipsPager integrated with a FlipView">
             <local:ControlExample.Example>
-                <muxc:PipsPager x:Name="TestPipsPager1"/>
+                <StackPanel>
+                    <FlipView x:Name="Gallery" MaxWidth="400" Height="270" ItemsSource="{x:Bind Pictures}">
+                        <FlipView.ItemTemplate>
+                            <DataTemplate x:DataType="x:String">
+                                <Image Source="{x:Bind Mode=OneWay}"/>
+                            </DataTemplate>
+                        </FlipView.ItemTemplate>
+                    </FlipView>
+                    <muxc:PipsPager x:Name="FlipViewPipsPager"
+                        HorizontalAlignment="Center"
+                        Margin="0, 10, 0, 0"
+                        NumberOfPages="{x:Bind Pictures.Count}"
+                        SelectedPageIndex="{x:Bind Path=Gallery.SelectedIndex, Mode=TwoWay}" />
+                </StackPanel>
+            </local:ControlExample.Example>
+            <local:ControlExample.Xaml>
+                <x:String xml:space="preserve"> 
+&lt;StackPanel&gt;
+    &lt;FlipView x:Name="Gallery" MaxWidth="400" Height="270" ItemsSource="{x:Bind Pictures}"&gt;
+        &lt;FlipView.ItemTemplate&gt;
+            &lt;DataTemplate x:DataType="x:String"&gt;
+                &lt;Image Source="{x:Bind Mode=OneWay}" /&gt;
+            &lt;/DataTemplate&gt;
+        &lt;/FlipView.ItemTemplate&gt;
+    &lt;/FlipView&gt;
+    &lt;muxc:PipsPager x:Name="FlipViewPipsPager"
+        Margin="140, 10, 0, 0"
+        NumberOfPages="{x:Bind Pictures.Count}"
+        SelectedPageIndex="{x:Bind Path=Gallery.SelectedIndex, Mode=TwoWay}" /&gt;
+&lt;/StackPanel&gt;
+                </x:String >
+            </local:ControlExample.Xaml>
+        </local:ControlExample>
+        <local:ControlExample x:Name="Example2" HeaderText="PipsPager with options to change its orientation and button visibility.">
+            <local:ControlExample.Example>
+                <muxc:PipsPager x:Name="TestPipsPager2"/>
             </local:ControlExample.Example>
             <local:ControlExample.Options>
                 <StackPanel>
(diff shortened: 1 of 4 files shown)


XamlControlsGallery/ControlPages/PipsPagerPage.xaml
```diff
@@ -0,0 +1,85 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:AppUIBasics"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <StackPanel>
+        <local:ControlExample x:Name="Example1" HeaderText="PipsPager with options to change its orientation and button visibility.">
```
stmoy, line 13: I'd recommend moving this sample to the end. Our general approach is to have the more "real world" examples at the top, and the "API in action" samples at the bottom (as seen in NavView).

XamlControlsGallery/ControlPages/PipsPagerPage.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+            "Assets/SampleMedia/LandscapeImage8.jpg",
+        };
+        public PipsPagerPage()
+        {
+            this.InitializeComponent();
+        }
+
+        private void OrientationComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            if (TestPipsPager1 == null) return;
```
stmoy, line 27: Is this check necessary?

XamlControlsGallery/ControlPages/PipsPagerPage.xaml
```diff
@@ -0,0 +1,85 @@
...
+                        <FlipView.ItemTemplate>
+                            <DataTemplate>
+                                <Image Source="{Binding}"/>
+                            </DataTemplate>
+                        </FlipView.ItemTemplate>
+                    </FlipView>
+                    <muxc:PipsPager x:Name="FlipViewPipsPager"
+                              Margin="140, 10, 0, 0"
+                              NumberOfPages="{x:Bind Pictures.Count}"
+                              SelectedPageIndex="{Binding ElementName=Gallery, Path=SelectedIndex, Mode=TwoWay}" />
```
stmoy, line 63: General question: this uses old-style Binding instead of x:Bind. Is it possible to use x:Bind with PipsPager?

XamlControlsGallery/ControlPages/PipsPagerPage.xaml
```diff
@@ -0,0 +1,85 @@
...
+                <StackPanel>
+                    <FlipView x:Name="Gallery" MaxWidth="400" Height="270" ItemsSource="{x:Bind Pictures}">
+                        <FlipView.ItemTemplate>
+                            <DataTemplate>
+                                <Image Source="{Binding}"/>
+                            </DataTemplate>
+                        </FlipView.ItemTemplate>
+                    </FlipView>
+                    <muxc:PipsPager x:Name="FlipViewPipsPager"
+                              Margin="140, 10, 0, 0"
```
stmoy, line 61: Magic numbers 😱 - should we have some guidance that says "put your pager x,y,z away from the paging element"?

YuliKl:
Are all these borders around the pips and arrows by design? Things look pretty crowded.

XamlControlsGallery/ControlPages/PipsPagerPage.xaml
YuliKl: Shouldn't PipsPager be horizontally centered relative to the FlipView? I know you can't do that in a vertical StackPanel, may just need to replace it with a Grid with rows.

XamlControlsGallery/ControlPages/PipsPagerPage.xaml.cs
gabbybilka: Likely not, since this event should only trigger when the PipsPager is valid as otherwise the selection wouldn't change. Good point.

XamlControlsGallery/ControlPages/PipsPagerPage.xaml
gabbybilka: I was able to horizontally center the PipsPager in the vertical StackPanel. The magic "top" margin number does still exist.

gabbybilka:
> Are all these borders around the pips and arrows by design? Things look pretty crowded.

No, they should be fixed and removed in a later visual update.

XamlControlsGallery/ControlPages/PipsPagerPage.xaml
```diff
@@ -0,0 +1,85 @@
...
+&lt;StackPanel&gt;
+    &lt;FlipView x:Name="Gallery" MaxWidth="400" Height="270" ItemsSource="{x:Bind Pictures}"&gt;
+        &lt;FlipView.ItemTemplate&gt;
+            &lt;DataTemplate x:DataType="x:String"&gt;
+                &lt;Image Source="{x:Bind Mode=OneWay}" /&gt;
+            &lt;/DataTemplate&gt;
+        &lt;/FlipView.ItemTemplate&gt;
+    &lt;/FlipView&gt;
+    &lt;muxc:PipsPager x:Name="FlipViewPipsPager"
+        Margin="140, 10, 0, 0"
```
YuliKl, line 41: Would be good to show the same changes in the sample code, too.

XamlControlsGallery/ControlPages/PipsPagerPage.xaml
```diff
@@ -0,0 +1,85 @@
...
+                    <FlipView x:Name="Gallery" MaxWidth="400" Height="270" ItemsSource="{x:Bind Pictures}">
+                        <FlipView.ItemTemplate>
+                            <DataTemplate x:DataType="x:String">
+                                <Image Source="{x:Bind Mode=OneWay}"/>
+                            </DataTemplate>
+                        </FlipView.ItemTemplate>
+                    </FlipView>
+                    <muxc:PipsPager x:Name="FlipViewPipsPager"
+                        HorizontalAlignment="Center"
+                        Margin="0, 10, 0, 0"
```
YuliKl, line 25:
```suggestion
                        Margin="0, 12, 0, 0"
```

This likely doesn't actually matter, but I try to keep to a 4px design grid. And at least in the past, 12px and 24 px have been the two most common margin sizes that design recommends.

XamlControlsGallery/ControlPages/PipsPagerPage.xaml
gabbybilka: Updated to 12px to stay in the 4px grid, are there other controls that have recommendations in spacing between it and another control?

YuliKl: As a rule of thumb, 12px is a good amount of space between controls on a page, 24px is a good margin around the exterior of a page (between content and edge of window). You may want to use 24px in a page when establishing visual groupings via whitespace.

PR #631, opened 2021-01-29
Fix indentation on SemanticZoomPage.xaml

srjheam:
## Description
Add properly indentation to `ItemTemplate` property in `SemanticZoom.ZoomedView` at `SemanticZoomPage.xaml`

## Motivation and Context
I took a look at SemanticZoom and found this annoying indentation error 😅

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/SemanticZoomPage.xaml
+++ b/XamlControlsGallery/ControlPages/SemanticZoomPage.xaml
@@ -47,7 +47,7 @@
             <SemanticZoom x:Name="Control1" Height="500">
                 <SemanticZoom.ZoomedInView>
                     <GridView ItemsSource="{x:Bind cvsGroups.View}" ScrollViewer.IsHorizontalScrollChainingEnabled="False" SelectionMode="None" GotFocus="List_GotFocus"
-                        ItemTemplate="{StaticResource ZoomedInTemplate}">
+                              ItemTemplate="{StaticResource ZoomedInTemplate}">
                         <GridView.GroupStyle>
                             <GroupStyle HeaderTemplate="{StaticResource ZoomedInGroupHeaderTemplate}" />
                         </GridView.GroupStyle>


XamlControlsGallery/ControlPages/SemanticZoomPage.xaml
```diff
@@ -47,7 +47,7 @@
             <SemanticZoom x:Name="Control1" Height="500">
                 <SemanticZoom.ZoomedInView>
                     <GridView ItemsSource="{x:Bind cvsGroups.View}" ScrollViewer.IsHorizontalScrollChainingEnabled="False" SelectionMode="None" GotFocus="List_GotFocus"
-                        ItemTemplate="{StaticResource ZoomedInTemplate}">
+                              ItemTemplate="{StaticResource ZoomedInTemplate}">
```
chingucoding, line 50: Can you elaborate why this is better and how the previous one was an "error", especially since this would introduce a third way to format in this file (firs one being the previous value and the second one being continuing where the last property started)?

srjheam: I think this is the right way to indent because it's used by Visual Studio and Microsoft itself in its other examples. But nothing prevents the identification of being this way:
```xaml
<GridView ItemsSource="{x:Bind cvsGroups.View}" ScrollViewer.IsHorizontalScrollChainingEnabled="False" SelectionMode="None" GotFocus="List_GotFocus" ItemTemplate="{StaticResource ZoomedInTemplate}">
```
or that way:
```xaml
<GridView ItemsSource="{x:Bind cvsGroups.View}"
          ScrollViewer.IsHorizontalScrollChainingEnabled="False"
          SelectionMode="None"
          GotFocus="List_GotFocus"
          ItemTemplate="{StaticResource ZoomedInTemplate}">
```

But I still think my changes are following .NET conventions and standards..

chingucoding: I don't think there is a right way at all for this, only personal preferences. While this style is being used in some places, it is not used in others, e.g. in the WinUI code base or the XAML sample [on this page](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/xaml-overview).

If it's a standard or convention and you can share where this is pointed out, gladly do so. I wasn't able to find anything from Microsoft on this. Also if it's the "right way", why not change all indentation in that file?

srjheam: > While this style is being used in some places, it is not used in others, e.g. in the WinUI code base or the XAML sample [on this page](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/xaml-overview).

I'm sorry, but I didn't get the point. This page only shows a different way to indent, nothing wrong.

> If it's a standard or convention and you can share where this is pointed out, gladly do so. I wasn't able to find anything from Microsoft on this.

As I said before, that pattern may be seen at Visual Studio, Microsoft Docs, Examples, and so on. As in [this example here](https://github.com/microsoft/Xaml-Controls-Gallery/blob/4212531e54058e75a2daaba761cfc89e85fc2afb/XamlControlsGallery/ControlPages/AcrylicPage.xaml#L122-L123), [here](https://github.com/microsoft/Xaml-Controls-Gallery/blob/4212531e54058e75a2daaba761cfc89e85fc2afb/XamlControlsGallery/ControlPages/CalendarViewPage.xaml#L38-L40) and in many other places. 

> Also if it's the "right way", why not change all indentation in that file?

I'm not saying that this is the only way, what I mean is that it's is just one way. The indentation in rest of that file is correct, following other patterns, but still correct as my fix.

chingucoding: > I'm sorry, but I didn't get the point. This page only shows a different way to indent, nothing wrong.

If there is nothing wrong with that style, why not leave the code as is as it is following that pattern of one indentation level?

> As I said before, that pattern may be seen at Visual Studio, Microsoft Docs, Examples, and so on. As in [this example here](https://github.com/microsoft/Xaml-Controls-Gallery/blob/4212531e54058e75a2daaba761cfc89e85fc2afb/XamlControlsGallery/ControlPages/AcrylicPage.xaml#L122-L123), [here](https://github.com/microsoft/Xaml-Controls-Gallery/blob/4212531e54058e75a2daaba761cfc89e85fc2afb/XamlControlsGallery/ControlPages/CalendarViewPage.xaml#L38-L40) and in many other places.

Yes the pattern is used in other files, but the pattern you are removing is also used in other files. And it is also used in the official documentation e.g. [on this page](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/xaml-syntax-guide) and [this page](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/x-bind-markup-extension). So it is inaccurate to say that it is an "indentation error". There are multiple ways to indent code and both are fine, saying that one is faulty, improper or wrong is not fine though.

> I'm not saying that this is the only way, what I mean is that it's is just one way. The indentation in rest of that file is correct, following other patterns, but still correct as my fix.

Saying that the previous one was an "indentation error" and that you chose the "proper" way kind of implies though that the previous pattern (also used in this file) was not correct.

srjheam: > And it is also used in the official documentation e.g. on this page and this page.

These examples use the exactly pattern I am tying to point out: parameter bellow parameter, all aligned.

chingucoding: > These examples use the exactly pattern I am tying to point out: parameter bellow parameter, all aligned.

I thought this was about how much you indent the code (e.g. one tab relative to the start, two tabs or length of the typename) not how these align. If you want to achieve that, why not have every parameter take it's own line like the examples show?

srjheam: Issue closed.

PR #633, opened 2021-02-03
Fix DependencyProperty registration

ajpinedam:
DependencyProperty Registration with the wrong name as Parameter

## Description

There is a DependencyProperty Registration with the wrong Name. 

## Motivation and Context

This was causing some crashes due to a Name being registered multiple times.

## How Has This Been Tested?

Running the App and making sure this change does not break other areas

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlExample.xaml.cs
+++ b/XamlControlsGallery/ControlExample.xaml.cs
@@ -84,7 +84,7 @@ namespace AppUIBasics
             {
                 return string.Empty;
             }
-            
+
             return value.ToString();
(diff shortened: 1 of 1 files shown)


chingucoding (approved):
Looks good to me. Thank you for fixing this @ajpinedam !

stmoy (approved):

stmoy:
Thanks for fixing this @ajpinedam! This was a silly mistake 🤦‍♂️

I've approved/merged this in, but I'm curious where this was crashing?

PR #654, opened 2021-03-18
update WinUI version to latest preview

ranjeshj:
Updating WinUI to latest pre-release version - 
https://github.com/microsoft/microsoft-ui-xaml/releases/tag/v2.6.0-prerelease.210315002

### DIFF
--- a/XamlControlsGallery/XamlControlsGallery.csproj
+++ b/XamlControlsGallery/XamlControlsGallery.csproj
@@ -1213,7 +1213,7 @@
       <Version>6.1.9</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.UI.Xaml">
-      <Version>2.6.0-prerelease.210217002</Version>
+      <Version>2.6.0-prerelease.210315002</Version>
     </PackageReference>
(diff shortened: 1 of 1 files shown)


chingucoding (approved):
:shipit:

stmoy (approved):

Poopooracoocoo:
@stmoy bruhhh where's the release at? 🥺

XCG doesn't need to be on the stable ver of WinUI, does it? I only use XCG every so often to test WinUI in the first place.

stmoy:
@Poopooracoocoo - I hear you. I know it's been awhile since we've pushed an update to the Store. It's unfortunately a very manual process today that takes more time than we'd like, which is why we've only been updating alongside major releases of WinUI. Additionally, we're very focused on getting WinUI 3 ready which is also a big effort and affects the Gallery. Realistically, we're planning on updating alongside WinUI 2.6 release and WinUI 3 releases; if I'm able to pull something together earlier than that, I'll let you know.

Poopooracoocoo:
@stmoy and just so we're clear, WinUI isn't popular. The entire purpose of XCG is for testing it.

Would be good if you could pull something together sooner :)

PR #666, opened 2021-04-01
Make the RichEditBox to be able to use Microsoft::UI::Xaml::CommandBarFlyout

licanhua:
When integrating with the latest WinUI, the sender is Microsoft::UI::Xaml::TextCommandBarFlyout other than Windows::UI::Xaml::CommandBarFlyout
It blocked the CommandBarFlyout pop up

### DIFF
--- a/XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
@@ -45,7 +45,7 @@ namespace AppUIBasics.ControlPages
             }
             else
             {
-                muxc::CommandBarFlyout muxFlyout = sender as muxc::CommandBarFlyout;
+                muxc.CommandBarFlyout muxFlyout = sender as muxc.CommandBarFlyout;
                 if (muxFlyout != null && muxFlyout.Target == REBCustom)
                 {
                     AppBarButton myButton = new AppBarButton
--- a/XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
@@ -19,6 +19,7 @@ using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Input;
 using Windows.UI.Xaml.Media;
+using Mux = Windows.UI.Xaml.Controls;
 
 namespace AppUIBasics.ControlPages
 {
@@ -34,7 +35,7 @@ namespace AppUIBasics.ControlPages
         private void Menu_Opening(object sender, object e)
         {
             CommandBarFlyout myFlyout = sender as CommandBarFlyout;
-            if (myFlyout.Target == REBCustom)
+            if (myFlyout != null && myFlyout.Target == REBCustom)
             {
                 AppBarButton myButton = new AppBarButton
                 {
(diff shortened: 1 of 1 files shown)


chingucoding (approved):
Looks good to me.

Edit: Nevermind, should have take a closer look at this...

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -34,14 +35,28 @@ public RichEditBoxPage()
...
-            if (myFlyout.Target == REBCustom)
+            if (myFlyout != null && myFlyout.Target == REBCustom)
             {
                 AppBarButton myButton = new AppBarButton
                 {
                     Command = new StandardUICommand(StandardUICommandKind.Share)
                 };
                 myFlyout.PrimaryCommands.Add(myButton);
             }
+            else
```
stmoy, line 46: What is the purpose of this code? It doesn't look to do much. Is the goal to expand on the RichEditBox sample to show how to integrate a CommandBarFlyout? Would this sample be better suited to the CommandBarFlyout page?

stmoy (changes requested):

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -34,14 +35,28 @@ public RichEditBoxPage()
...
+            {
+                Mux::CommandBarFlyout muxFlyout = sender as Mux::CommandBarFlyout;
```
chingucoding: Don't think that that works in C#.

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -19,6 +19,7 @@
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Input;
 using Windows.UI.Xaml.Media;
+using Mux = Windows.UI.Xaml.Controls;
```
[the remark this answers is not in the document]
chingucoding: C# code almost entirely writes out Microsoft.UI.Xaml.Controls instead of a named import, advised to do this here too.

YuliKl:
I'm confused. Why do we need code changes in XCG? Shouldn't the CommandBarFlyout showing up just happen for default RichEditBox, no app-side code required? (@ranjeshj @llongley)

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
stmoy: Agreed. Does this build?

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
licanhua: It's added a share button to the commandbarflyout. If I remove `else`, it does have the share button

image

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
licanhua: yes. it passed the compilation and no error in run time

licanhua:
> I'm confused. Why do we need code changes in XCG? Shouldn't the CommandBarFlyout showing up just happen for default RichEditBox, no app-side code required? (@ranjeshj @llongley)


I don't know why the app didn't crash. In debug mode, myFlyout is null.
For some reason, I think we didn't use mux::CommandBarFlyout in richeditbox before, and new winui fixed the issue.

xcg has problem in below code:
```
if (myFlyout.Target == REBCustom)
```

XamlControlsGallery/ControlPages/RichEditBoxPage.xaml.cs
YuliKl: Right, one of the XCG samples shows how to add a custom button to the default text edit menu. That should indeed be an app-side fix.

But the original issue was that the text edit menus weren't showing at all. That problem I would have expected to be fixed in WinUI.

So maybe we need two unrelated PRs? XCG should be showing the default floaty at least in some samples.

YuliKl: This sample should just work:
image

This sample may need an additional fix in XCG:
image

licanhua: I didn't see winui has problem. Maybe VS compiler issue?

I expect the app crashed when myFlyout is null, but it's not. Instead, XCG doesn't show the default floaty

You can verify it locally to change
```
if (myFlyout.Target == REBCustom)
```

to
```
if (myFlyout != null && myFlyout.Target == REBCustom)
```

ranjeshj (approved):
:shipit:

PR #689, opened 2021-05-04
Use muxc:PersonPicture

licanhua:
Use muxc:PersonPicture instead of wuxc:PersonPicture

### DIFF
--- a/XamlControlsGallery/ControlPages/PersonPicturePage.xaml
+++ b/XamlControlsGallery/ControlPages/PersonPicturePage.xaml
@@ -28,7 +28,7 @@
             </local:ControlExample.Options>
             <local:ControlExample.Xaml>
                 <x:String>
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/PersonPicturePage.xaml
```diff
@@ -12,7 +12,7 @@
         <local:ControlExample x:Name="Example1" HeaderText="Select different looks for the person picture.">
 
             <local:ControlExample.Example>
-                <PersonPicture x:Name="personPicture" Height="300"
+                <muxc:PersonPicture x:Name="personPicture" Height="300"
```
chingucoding, line 15: Please also update the sample code in line 31.

chingucoding (changes requested):

ranjeshj (approved):

PR #708, opened 2021-06-20
Fix small issues around the animated ItemsRepeater sample

### LINKED ISSUE #579: Narrator does not announce the colors getting selected in ItemsRepeater.
**Describe the bug**

1. Turn on Narrator
2. Launch XAML Controls Gallery.
3. Activate ItemsRepeater from home page, navigate to header "Animated Scrolling and Content Display" and select the colors using "caps+arrow" keys and observe.

**Expected behavior**

Narrator should announce the information whether the item got selected or not along with describing other attributes.

**Actual behavior**

Narrator only announces the button name when activating it but tells nothing whether it is selected or not.

**Screenshots**

image

**Version Info**

Gallery version (found on Settings page): 1.2.13.0

**Additional context**

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_queries/edit/25440841)

chingucoding:
## Description

Fix issue where the selected element was not indicated to users using assistive technology.

Also switched to ViewChanged event as the selected color was faulty with the ViewChanging event.

## Motivation and Context

Closes #579 
## How Has This Been Tested?

Test manually
## Types of changes


XamlControlsGallery/ControlPages/ItemsRepeaterPage.xaml.cs
```diff
@@ -291,6 +299,17 @@ private void Animated_ScrollViewer_ViewChanging(object sender, ScrollViewerViewC
...
         }
 
+        private void SetUIANamesForSelectedEntry(Button selectedItem)
+        {
+            if (LastSelectedColorButton != null && LastSelectedColorButton.Content is string content)
+            {
+                AutomationProperties.SetName(LastSelectedColorButton, content);
+            }
+
+            AutomationProperties.SetName(selectedItem, (string)selectedItem.Content + " , selected");
```
anawishnoff, line 309: Do we need a null check for selected item?

XamlControlsGallery/ControlPages/ItemsRepeaterPage.xaml.cs
```diff
@@ -279,9 +284,12 @@ private void GetButtonSize(object sender, RoutedEventArgs e)
             AnimatedBtnMargin = AnimatedBtn.Margin;
         }
 
-        private void Animated_ScrollViewer_ViewChanging(object sender, ScrollViewerViewChangingEventArgs e)
+        private void Animated_ScrollViewer_ViewChanged(object sender, ScrollViewerViewChangedEventArgs e)
```
anawishnoff, line 287: Why the change (not against it, just curious)?

chingucoding: ViewChanging does not fire once scrolling is done but rather in between. So if you scroll fast, the item in center and the displayed color are actually different (sometimes it isn't even on the screen anymore):

Selected color not matching the visually selected color

XamlControlsGallery/ControlPages/ItemsRepeaterPage.xaml.cs
chingucoding: You could but the code would fail on different areas if the passed parameter would be null so there is no benefit in adding that I think.

anawishnoff (approved):
Same conflict issue as the other PR - otherwise approved!

PR #709, opened 2021-06-24
Update to WinUI 2.6

kmahone:


StephenLPeters (approved):

XamlControlsGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -15,18 +15,41 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
     xmlns:local="using:AppUIBasics"
-    HighContrastAdjustment="None">
+    HighContrastAdjustment="None"
+    muxc:BackdropMaterial.ApplyToRootOrPageBackground="True">
```
chingucoding, line 19: Is this code supposed to work? On my machine, I'm getting a compiler error due to BackDropMaterial.ApplyToRootOrPageBackground not being an attached property for type page. @kmahone @StephenLPeters

ranjeshj: @chingucoding This is an API that was added recently. This would need the insider sdk to work.

chingucoding: And in non insider SDKs, this line is supposed to not compile then?

ranjeshj: @chingucoding I think this should work even without the insider sdk. Are you still hitting this ? Can you share the build error ?

chingucoding: Can't repro this now that the XCG is using 2.6.1, if I downgrad to 2.6.0 it fails with error "The attachable property 'ApplyToRootOrPageBackground' was not found in type 'BackdropMaterial'.".
Since I can't repro this with 2.6.1, I guess there is nothing to fix here anymore is there?

PR #716, opened 2021-06-28
Update MaxTestedVersion to >= Windows 11

jevansaks:
There isn't an SDK for Windows 11 yet but we have tested the app on Windows 11 so set the MaxTestedVersion = 22200 so we get the latest behavior. We can remove/undo this when the SDK is available.


jlaanstra:
@jevansaks shouldn't this be 22000 instead of 22200?

jevansaks:
You're right, but it's also ok because 22000 is the _smallest_ number we could have that works but anything higher is technically ok too.

PR #730, opened 2021-07-18
Update ListView Drop handler to be more generalizable

MichaelKintscher:
Modified the "ListViews with Drag, Drop, and Reordering Support" sample to make it more generalizable from the specific setup in this app. Also increased clarity in comments, and removed a typo.

## Description

Modified the Drop handler for the "ListViews with Drag, Drop, and Reordering Support" sample's two ListViews. The insertion index is now initialized to zero. Only if there are items in the target ListView, does the calculation of the insertion index proceed. Otherwise the calculation is skipped and the index retains its initial value of zero. Also updated the comments to remove some ambiguity in describing what the code is doing. Finally, corrected a typo by removing an excess semicolon at the end of what was line 174 (now is line 170).

## Motivation and Context

Previous version assumed height of ListViewItems in both source and target ListViews were identical. A sample ListViewItem (sampleItem) was selected from the source ListView. The height of this sampleItem was then measured to calculate the insertion index for the dropped items. If the item heights in the two ListViews were different, the insertion index would be incorrectly calculated, resulting in the item being inserted in the wrong place, possibly off-screen and possibly confusing the user.

Came across this issue when trying to follow the sample to implement this drag and drop functionality between two ListViews. The ListViews in this case did not have identical ListViewItem heights.

## How Has This Been Tested?

Copied the ContactListViewTemplate DataTemplate in the Page.Resources and modified the copy to assign a different height to the Grid.. Assigned this modified DataTemplate to the ItemTemplate property on DragDropListView2. Built and ran the code. Dragged and dropped items between the DragDropListView and DragDropListView2, in both directions. Items were not inserted at the correct positions.

Repeated the same procedure with the changes made in this pull request to the Drop event handler for the two ListViews. Items were now inserted at the correct positions, regardless of size differences between the two ListViews' item templates.

Tested using Visual Studio 2019 Version 16.10.0 on a 2017 Surface Pro. Confirmed the only two references to the changed function were the two ListViews (DragDropListView and DragDropListView2), so no other area of the code should be affected.

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/ListViewPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/ListViewPage.xaml.cs
@@ -152,36 +152,37 @@ namespace AppUIBasics.ControlPages
                     // Find the insertion index:
                     Windows.Foundation.Point pos = e.GetPosition(target.ItemsPanelRoot);
 
-                    // Find which ListView is the target, find height of first item
-                    ListViewItem sampleItem;
-                    if (target.Name == "DragDropListView")
+                    // If the target ListView has items in it, use the heigh of the first item
+                    //      to find the insertion index.
+                    int index = 0;
+                    if (target.Items.Count != 0)
                     {
-                        sampleItem = (ListViewItem)DragDropListView2.ContainerFromIndex(0);
+                        // Get a reference to the first item in the ListView
+                        ListViewItem sampleItem = (ListViewItem)target.ContainerFromIndex(0);
+
+                        // Adjust itemHeight for margins
+                        double itemHeight = sampleItem.ActualHeight + sampleItem.Margin.Top + sampleItem.Margin.Bottom;
+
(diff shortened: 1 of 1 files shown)


chingucoding (approved):
Thank you for updating this sample! Changes look good to me.

Usually, for our samples, we also show the C# source code for them to make it easier for users to get started. This sample does not have this so far, if you like you could also add this code to the sample. Of course you don't have to do this, the changes are already very good!

Also, great to see you thoroughly tested this!

anawishnoff:
Thank you @MichaelKintscher, this is a great fix and I appreciate the thorough explanation. I agree that we usually try to add the C#/code-behind to the sample on the page as well, but for samples like this with more complicated logic it's fine to leave it out. I'll go ahead and approve/merge this now.

@chingucoding Thank you for all of your help in looking through these PRs as well! Much appreciated.

anawishnoff (approved):

PR #781, opened 2021-09-24
Add AutomationProperties.Name to NumberBox

### LINKED ISSUE #325: Name Property missing for Progress text box present under header 'A determinate Progress bar'.
**Describe the bug**

2. Activate ProgressBar from home page.
3. Navigate to text box under header A determinate progress bar

**Expected behavior**

Name Property should be there for text box,

**Actual behavior**

Text box name property missing.

**Screenshots**

image

**Version Info**

1.2.12.0

**Additional context**

Copied from [internal bug](https://dev.azure.com/microsoft/OS/_workitems/edit/24649907)

karenbtlai:
## Description

Add AutomationProperties.Name to NumberBox to pass all Accessibility Insights test

## Motivation and Context

Fixes #325

### DIFF
--- a/XamlControlsGallery/ControlPages/ProgressBarPage.xaml
+++ b/XamlControlsGallery/ControlPages/ProgressBarPage.xaml
@@ -47,9 +47,15 @@
                 <muxc:ProgressBar Width="130" x:Name="ProgressBar2"/>
                 <TextBlock x:Name="Control2Output" Style="{ThemeResource OutputTextBlockStyle}" Width="60" TextAlignment="Center" />
                 <TextBlock x:Name="ProgressLabel"  Text="Progress" VerticalAlignment="Center" Margin="0,0,10,0"/>
-                <muxc:NumberBox x:Name="ProgressValue" AutomationProperties.LabeledBy="{Binding ElementName=ProgressLabel}"
-                    Minimum="0" Maximum="100" Value="0"
-                    SpinButtonPlacementMode="Inline" ValueChanged="ProgressValue_ValueChanged"/>
+                <muxc:NumberBox
+                    x:Name="ProgressValue"
+                    AutomationProperties.LabeledBy="{Binding ElementName=ProgressLabel}"
+                    AutomationProperties.Name="NumberBox controlling ProgressBar2 value"
+                    Minimum="0"
+                    Maximum="100"
+                    Value="0"
+                    SpinButtonPlacementMode="Inline"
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/ProgressBarPage.xaml
```diff
@@ -47,9 +47,15 @@
                 <muxc:ProgressBar Width="130" x:Name="ProgressBar2"/>
                 <TextBlock x:Name="Control2Output" Style="{ThemeResource OutputTextBlockStyle}" Width="60" TextAlignment="Center" />
                 <TextBlock x:Name="ProgressLabel"  Text="Progress" VerticalAlignment="Center" Margin="0,0,10,0"/>
-                <muxc:NumberBox x:Name="ProgressValue" AutomationProperties.LabeledBy="{Binding ElementName=ProgressLabel}"
-                    Minimum="0" Maximum="100" Value="0"
-                    SpinButtonPlacementMode="Inline" ValueChanged="ProgressValue_ValueChanged"/>
+                <muxc:NumberBox
+                    x:Name="ProgressValue"
+                    AutomationProperties.LabeledBy="{Binding ElementName=ProgressLabel}"
+                    AutomationProperties.Name="NumberBox controlling ProgressBar2 value"
```
chingucoding, line 53: Shouldn't setting the LabeledBy property be enough here? Could this actually be an issue with the NumberBox control?

karenbtlai: LabeledBy applies the automationProperties.Name property to NumberBox but not the textbox, but mysteriously, the .Name property applies the name to both. This behaviour is duplicated with ComboBox. This is a temporary fix while I investigate the discrepancies between the two properties.

chingucoding: I see, very strange that it works for one of the properties but not for the other. Could we template bind to the labeled by property maybe to mitigate this?

chingucoding (approved):
Fine to merge this as workaround until there is a proper fix on NumberBox. Thank you for fixing this!

XamlControlsGallery/ControlPages/ProgressBarPage.xaml
karenbtlai: One option is to templatebind InputBox to the header name like ComboBox, but that still doesn't solve the issue if there is no header available.

ranjeshj (approved):

PR #810, opened 2021-11-02
Add copy code button to code samples

chingucoding:
## Description

This updates the code samples to also have a "copy sample code" button to make it easier to copy sample code.
To make development of this and future features easier, the code samples have been moved into their own control.
## How Has This Been Tested?

Tested manually.
Video of copy code button showcasing it's tooltip and the confirmation message when clicking on it.
## Types of changes


XamlControlsGallery/App.xaml
```diff
@@ -122,6 +122,7 @@
                 <Setter Property="HorizontalContentAlignment" Value="Stretch" />
             </Style>
 
+            <x:Double x:Key="TeachingTipMinWidth">100</x:Double>
```
chingucoding, line 125: We need this since we need to override the resource on a level that is above the flyout layer which only is the App.xaml since the text for the copy code button is too short and we would have a gap.

XamlControlsGallery/XamlControlsGallery.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29728.190
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31825.309
```
chingucoding, line 4: Those are from using VS 2022, happy to revert those changes if desired.

chingucoding:
Sorry about the large PR @gregwoo-microsoft but I think that this step would make future changes a lot easier. Happy to hear your thoughts about the copy code button and it's UX!

gregwoo-microsoft:
Thanks a lot @chingucoding! I ran the XCG and the changes look great! I want to spend more time to review/better understand your code changes before I approve this PR. Amazing work! This is awesome :)

XamlControlsGallery/Controls/ControlExample.xaml.cs
```diff
@@ -1,577 +1,417 @@
...
+        }
+
+        public static readonly DependencyProperty MinimumUniversalAPIContractProperty = DependencyProperty.Register("MinimumUniversalAPIContract", typeof(int), typeof(ControlExample), new PropertyMetadata(null));
+        public int MinimumUniversalAPIContract
+        {
+            get { return (int)GetValue(MinimumUniversalAPIContractProperty); }
+            set { SetValue(MinimumUniversalAPIContractProperty, value); }
+        }
+
+        private ILanguage CSharpLanguage = Languages.CSharp;
```
gregwoo-microsoft, line 191: Commenting these lines to highlight diff since Github isn't highlighting them

XamlControlsGallery/Controls/ControlExample.xaml.cs
```diff
@@ -1,577 +1,417 @@
...
+            set { SetValue(HorizontalContentAlignmentProperty, value); }
+        }
+
+        public static readonly DependencyProperty MinimumUniversalAPIContractProperty = DependencyProperty.Register("MinimumUniversalAPIContract", typeof(int), typeof(ControlExample), new PropertyMetadata(null));
+        public int MinimumUniversalAPIContract
+        {
+            get { return (int)GetValue(MinimumUniversalAPIContractProperty); }
+            set { SetValue(MinimumUniversalAPIContractProperty, value); }
+        }
+
```
gregwoo-microsoft, line 190: Main change is from 190 to 209. Commenting these lines to highlight diff since Github isn't highlighting them

gregwoo-microsoft:
This looks great Marcel! Thanks a lot for this! I changed the copied text to "Copied!" (one problem though is that I can't get it to be at the center of the control tip). I think that this phrasing is more concise than "Successfully copied sample code!" and is also more subtle/discrete. Let me know what you think :)

XamlControlsGallery/XamlControlsGallery.sln
chingucoding: Reverted the changes now :)

chingucoding (commented):
> This looks great Marcel! Thanks a lot for this! I changed the copied text to "Copied!" (one problem though is that I can't get it to be at the center of the control tip). I think that this phrasing is more concise than "Successfully copied sample code!" and is also more subtle/discrete. Let me know what you think :)

Very good idea! The problem with centering is that the TeachingTip control has it's own thoughts of how content should be (we are using the subtitle and that is bound to the left). I've updated the overriding of the TeachingTipMinWidth to 50 so that the TeachingTIp is only as large as the text and we don't have any gaps there.

XamlControlsGallery/App.xaml
```diff
@@ -122,6 +122,7 @@
                 <Setter Property="HorizontalContentAlignment" Value="Stretch" />
             </Style>
 
+            <x:Double x:Key="TeachingTipMinWidth">50</x:Double>
```
chingucoding, line 125: @StephenLPeters, is there a better way to override this than having this resource in the App.xaml file? I tried setting this in TeachingTip.ResourceDictionary however that did not work (or I did something wrong there). Maybe it's worth introducing properties so one can override those properties on a per control basis (unless lightweight styling using the controls ResourceDictionary works, then not).

gregwoo-microsoft: This seems to be the best way to do it for now (https://stackoverflow.com/questions/59542235/unable-to-set-width-height-property-of-teaching-tip-in-uwp)

chingucoding: Guess it's just one of those popup layer problems that we can't get around for now, thanks for checking @gregwoo-microsoft! Maybe this is worth adding a property on the teachingtip, but that's a discussion for a different time.

gregwoo-microsoft: Yes I agree. I'll approve the changes then :)

gregwoo-microsoft (approved):
Thank you Marcel 💯

PR #831, opened 2021-12-12
Add adjustable slider properties in the slider example

HO-COOH:
## Description
Make the example2 in Sliders Page manually adjustable.

## Motivation and Context
I was tinkering with sliders in my own project and got some exceptions. Then I checked the slider examples but found the example doesn't provide manual settings for manipulating the slider properties, so I decide to add them.

## How Has This Been Tested?
Manually test on the GUI.

from this
image
to this
image

## Types of changes


XamlControlsGallery/ControlPages/SliderPage.xaml
```diff
@@ -30,17 +31,53 @@
...
+
+                    <TextBlock Text="StepFrequency:" Margin="0,5,0,0" Grid.Row="2" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="StepFrequencyValue" Grid.Row="2" Grid.Column="1" Margin="10,5,0,0" Value="10" SpinButtonPlacementMode="Compact" ValueChanged="StepFrequencyValue_ValueChanged"/>
+
+                    <TextBlock Text="SmallChange:" Margin="0,5,0,0" Grid.Row="3" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="SmallChangeValue" Grid.Row="3" Grid.Column="1" Margin="10,5,0,0" Value="10" SpinButtonPlacementMode="Compact" ValueChanged="SmallChangeValue_ValueChanged"/>
+
+                    <TextBlock Text="LargeChange:" Margin="0,5,0,0" Grid.Row="4" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="LargeChangeValue" Grid.Row="4" Grid.Column="1" Margin="10,5,0,0" Value="100" SpinButtonPlacementMode="Compact" ValueChanged="LargeChangeValue_ValueChanged"/>
+                </Grid>
```
chingucoding, line 62: I would place these values inside the ControlExample.Options part of the ControlExample. An example of this can be found here: https://github.com/microsoft/Xaml-Controls-Gallery/blob/master/XamlControlsGallery/ControlPages/NavigationViewPage.xaml#L231-L239

XamlControlsGallery/ControlPages/SliderPage.xaml.cs
```diff
@@ -30,5 +30,63 @@ public SliderPage()
...
+            }
+            Slider2.Maximum = args.NewValue;
+        }
+
+        private void StepFrequencyValue_ValueChanged(Microsoft.UI.Xaml.Controls.NumberBox sender, Microsoft.UI.Xaml.Controls.NumberBoxValueChangedEventArgs args)
+        {
+            try
+            {
+                Slider2.StepFrequency = (float)args.NewValue;
+            }
```
chingucoding, line 61: Can this even fail? If so, can we guard against this in a better way or just use x:Bind for this instead?

XamlControlsGallery/ControlPages/SliderPage.xaml.cs
```diff
@@ -30,5 +30,63 @@ public SliderPage()
...
+            catch (Exception)
+            {
+                Slider2.StepFrequency = (Slider2.Maximum - Slider2.Minimum) / 100.0;
```
chingucoding, line 64: This seems a bit arbitrary, out of curiosity, is there a specific reason for this?

XamlControlsGallery/ControlPages/SliderPage.xaml.cs
```diff
@@ -30,5 +30,63 @@ public SliderPage()
...
+            }
+        }
+
+        private void SmallChangeValue_ValueChanged(Microsoft.UI.Xaml.Controls.NumberBox sender, Microsoft.UI.Xaml.Controls.NumberBoxValueChangedEventArgs args)
+        {
+            try
+            {
+                Slider2.SmallChange = (float)args.NewValue;
+            }
```
chingucoding, line 73: Same as above, can this fail or can we guard better against this or can we even use x:Bind?

XamlControlsGallery/ControlPages/SliderPage.xaml.cs
```diff
@@ -30,5 +30,63 @@ public SliderPage()
...
+            {
+                Slider2.SmallChange = (Slider2.Maximum - Slider2.Minimum) / 100.0;
+            }
+        }
+
+        private void LargeChangeValue_ValueChanged(Microsoft.UI.Xaml.Controls.NumberBox sender, Microsoft.UI.Xaml.Controls.NumberBoxValueChangedEventArgs args)
+        {
+            try
+            {
+                Slider2.LargeChange = (float)args.NewValue;
```
chingucoding, line 84: Same as above, can this fail or can we guard better against this or can we even use x:Bind?

XamlControlsGallery/ControlPages/SliderPage.xaml.cs
```diff
@@ -30,5 +30,63 @@ public SliderPage()
...
+
+        private void SmallChangeValue_ValueChanged(Microsoft.UI.Xaml.Controls.NumberBox sender, Microsoft.UI.Xaml.Controls.NumberBoxValueChangedEventArgs args)
+        {
+            try
+            {
+                Slider2.SmallChange = (float)args.NewValue;
+            }
+            catch (Exception)
+            {
+                Slider2.SmallChange = (Slider2.Maximum - Slider2.Minimum) / 100.0;
```
chingucoding, line 76: Same as above, is there a specific reason for this exact formula?

XamlControlsGallery/ControlPages/SliderPage.xaml.cs
```diff
@@ -30,5 +30,63 @@ public SliderPage()
...
+
+        private void LargeChangeValue_ValueChanged(Microsoft.UI.Xaml.Controls.NumberBox sender, Microsoft.UI.Xaml.Controls.NumberBoxValueChangedEventArgs args)
+        {
+            try
+            {
+                Slider2.LargeChange = (float)args.NewValue;
+            }
+            catch (Exception)
+            {
+                Slider2.LargeChange = (Slider2.Maximum - Slider2.Minimum) / 10.0;
```
chingucoding, line 88: Same as above, is there a specific reason for this exact formula?

chingucoding (changes requested):
Very good idea @HO-COOH , thank you for adding this feature!

I left a few comments on things that might be better done in a different way.

XamlControlsGallery/ControlPages/SliderPage.xaml
HO-COOH: > Thanks for that information.

HO-COOH:
Now using `ControlExample.Options`.
image

Indeed using binding seems simpler and it doesn't seems to throw exceptions even with exceptional inputs, like maximum greater than minimum. Would dive in more.

gregwoo-microsoft:
Thank you @HO-COOH for implementing this! This is a great idea. Also, thank you @chingucoding for providing feedback. I'll have a deeper look at this PR later today before merging it 😄

XamlControlsGallery/ControlPages/SliderPage.xaml
```diff
@@ -27,21 +28,65 @@
...
+                    <muxc:NumberBox x:Name="MaximumValue" Grid.Row="1" Grid.Column="1" Margin="10,5,0,0" Value="1000" SpinButtonPlacementMode="Compact"/>
+
+                    <TextBlock Text="StepFrequency:" Margin="0,5,0,0" Grid.Row="2" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="StepFrequencyValue" Grid.Row="2" Grid.Column="1" Margin="10,5,0,0" Value="10" SpinButtonPlacementMode="Compact"/>
+
+                    <TextBlock Text="SmallChange:" Margin="0,5,0,0" Grid.Row="3" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="SmallChangeValue" Grid.Row="3" Grid.Column="1" Margin="10,5,0,0" Value="10" SpinButtonPlacementMode="Compact"/>
+
+                    <TextBlock Text="LargeChange:" Margin="0,5,0,0" Grid.Row="4" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="LargeChangeValue" Grid.Row="4" Grid.Column="1" Margin="10,5,0,0" Value="100" SpinButtonPlacementMode="Compact"/>
```
chingucoding, line 60: A11y wise, this is not ideal. Either you should set the `AutomationProperties.LabeledBy` property on the NumberBoxes to the textblocks or set the `AutomationProperties.Name` property of these NumberBoxes. Alternatively, we could use the NumberBox's header and maybe update the layout into a two column layout where each column contains NumberBoxes.

XamlControlsGallery/ControlPages/SliderPage.xaml
```diff
@@ -27,21 +28,65 @@
...
+                    <Grid.RowDefinitions>
+                        <RowDefinition/>
+                        <RowDefinition/>
+                        <RowDefinition/>
+                        <RowDefinition/>
+                        <RowDefinition/>
+                    </Grid.RowDefinitions>
+                    <Grid.ColumnDefinitions>
+                        <ColumnDefinition Width="Auto"/>
+                        <ColumnDefinition Width="Auto"/>
```
chingucoding, line 44: Nit: Set the `MinWidth` property so that the column doesn't resize upon changes that go from 3 digit to 4 digit numbers.

XamlControlsGallery/ControlPages/SliderPage.xaml
HO-COOH: Resolved, but not sure how does it work tho. Adding `AutomationProperties.Name` seems like it makes the narrator speaks its name upon focus, while adding `AutomationProperties.LabeledBy` seems doing nothing.

chingucoding: Dang it, I think this is an issue with NumberBox and how it handles UIA. Guess we have to fall back to using the AutomationProperties.Name property, sorry about the confusion.

HO-COOH: Should be good now.

XamlControlsGallery/ControlPages/SliderPage.xaml
```diff
@@ -27,21 +28,65 @@
...
+
+                    <TextBlock x:Name="StepFrequencyLabel" Text="StepFrequency:" Margin="0,5,0,0" Grid.Row="2" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="StepFrequencyValue" Grid.Row="2" Grid.Column="1" Margin="10,5,0,0" Value="10" SpinButtonPlacementMode="Compact" AutomationProperties.Name="{x:Bind StepFrequencyValue.Name}"/>
+
+                    <TextBlock x:Name="SmallChangeLabel" Text="SmallChange:" Margin="0,5,0,0" Grid.Row="3" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="SmallChangeValue" Grid.Row="3" Grid.Column="1" Margin="10,5,0,0" Value="10" SpinButtonPlacementMode="Compact" AutomationProperties.Name="{x:Bind SmallChangeValue.Name}"/>
+
+                    <TextBlock x:Name="LargeChangeLabel" Text="LargeChange:" Margin="0,5,0,0" Grid.Row="4" Grid.Column="0"/>
+                    <muxc:NumberBox x:Name="LargeChangeValue" Grid.Row="4" Grid.Column="1" Margin="10,5,0,0" Value="100" SpinButtonPlacementMode="Compact" AutomationProperties.Name="{x:Bind LargeChangeValue.Name}"/>
```
chingucoding, line 60: You should set the `AutomationProperties.Name` property to something like "Minimum" instead of `{x:Bind MinimumValue.Name}`. Also you should set the `AutomationProperties.AccessibilityView` property to `Raw`. Sorry for the ping-pong on this PR 😅

HO-COOH: Learned something from `AutomationProperties`. Should be good now

gregwoo-microsoft:
Thanks a lot @HO-COOH and @chingucoding! This is great. I've added the "Minimum" property to the StepFrequencyValue NumberBox to prevent crashes when set to 0. While I was looking at the slider page, I've also slightly modified the sample underneath to snap to the tick marks.

Before I merge your changes, I have a question regarding the "LargeChange" property. I can't manage to make it work on my end (clicking on the slider moves the thumb to the area where I clicked. I expected it to move by the "LargeChange" value instead). Let me know if you are experiencing the same thing or not.

chingucoding:
I'm also experiencing this @gregwoo-microsoft. However I'm not sure what the purpose of `LargeChange` is and how it is supposed to work since I didn't find any documentation on this. My guess would be that this exists for UIA purposes and make navigation that way easier.

gregwoo-microsoft:
I agree that the documentation is not clear on this. I'll see what I can do to update it.

Based on what I found online (https://wpf.2000things.com/2014/03/10/1025-small-and-large-changes-to-the-value-of-a-slider/) the LargeChange property is supposed to indicate how much a slider should move when clicking on either side of the thumb. Right now when you click on either side, the thumb jumps to that location instead of being incremented by the LargeChange value.

I think that it would be best to remove the "Small Change" and "LargeChange" knobs from this PR. I am afraid that it may cause more harm than good by creating a lot of confusion. What do you think?

chingucoding:
I see, that makes sense, thank you for help @gregwoo-microsoft ! It kind of makes sense to remove it since it can be confusing, however on the other hand, there isn't really much left from the API if we remove the small and large change inputs.

gregwoo-microsoft:
@stmoy what are your thoughts on this?

XamlControlsGallery/ControlPages/SliderPage.xaml
```diff
@@ -27,31 +28,75 @@
...
                 </x:String>
             </local:ControlExample.Xaml>
+            <local:ControlExample.Substitutions>
+                <local:ControlExampleSubstitution Key="MinimumValue" Value="{x:Bind MinimumValue.Value, Mode=OneWay}"/>
+                <local:ControlExampleSubstitution Key="MaximumValue" Value="{x:Bind MaximumValue.Value, Mode=OneWay}"/>
+                <local:ControlExampleSubstitution Key="StepFrequencyValue" Value="{x:Bind StepFrequencyValue.Value, Mode=OneWay}"/>
+                <local:ControlExampleSubstitution Key="SmallChangeValue" Value="{x:Bind SmallChangeValue.Value, Mode=OneWay}"/>
+                <local:ControlExampleSubstitution Key="LargeChangeValue" Value="{x:Bind LargeChangeValue.Value, Mode=OneWay}"/>
+                <local:ControlExampleSubstitution Key="Value" Value="{x:Bind Slider2.Value, Mode=OneWay}"/>
+            </local:ControlExample.Substitutions>
```
chingucoding, line 87: Sorry for the delay on this @HO-COOH . Given that the large change property does not change much, could you remove it from the sample? Without it, this PR would be ready to merge I think.

HO-COOH:
@chingucoding What's the problem?

chingucoding (approved):
My bad, it seemed that I was viewing changes of the latest commits and not the entire PR. 

PR looks good to me.

chingucoding:
@gregwoo-microsoft What are your thoughts, can we merge this?

PR #832, opened 2021-12-14
Remove misleading, unused property "IsSelected" in TreeView sample

Marv51:
## Description
This PR removes the `IsSelected` property, on a sample view model on the TreeView page. This was unused code before.

At least to me, the property strongly implies that it might be used with TreeView to bind to the `IsSelected` property of the TreeViewItem. This however, [does not work](https://github.com/microsoft/microsoft-ui-xaml/issues/125). So I think this is a bad sample property and should be removed.

## Motivation and Context
See description above

## How Has This Been Tested?
Tested manually, app still builds, runs and TreeView page still works.

none

## Types of changes

- [x] Minor

### DIFF
--- a/XamlControlsGallery/ControlPages/TreeViewPage.xaml.cs
+++ b/XamlControlsGallery/ControlPages/TreeViewPage.xaml.cs
@@ -199,22 +199,6 @@ namespace AppUIBasics.ControlPages
             }
         }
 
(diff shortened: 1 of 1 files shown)


gregwoo-microsoft (approved):
Good catch! I don't see any reference to this code either. Thank you! :)

PR #840, opened 2022-01-11
Fix splitbutton keyboard accessability.

StephenLPeters:
Fix a bug where the items in the split button drop down were not keyboard invokable, which was (inappropriately) being flagged as a sev 1 accessibility issue for Winui.


gregwoo-microsoft (approved):
Thank you @StephenLPeters for pushing this. This also simplifies our code which is awesome 💯

PR #868, opened 2022-02-08
Update TextBoxPage.xaml to fix color contrast ratio 

### LINKED ISSUE #867: TextBox card: Color contrast ratio is less than 4.5:1 for some text content in the app.
Repro Steps:

1. Launch the WinUI app
2. Navigate to search box using tab key.
3. Type 'TextBox' in search box and hit enter.
4. Now navigate to TextBox card and activate it.
5. Check color contrast ratio for " I am super excited to be here!".

Actual Result:
Color contrast ratio is lessthan 4.5:1 for some text content in the app.

Expected Result:
Color contrast ratio is greaterthan or equal to 4.5:1 for some text content in the app.

User Impact: 

People with low vision are not able to consume the text which is having low contrast ratio

Windows version:
- [ ] Insider Build (xxxxx)
- [ ] May 2021 Update (19043)
- [ ] October 2020 Update (19042)
- [ ] May 2020 Update (19041)
- [ ] November 2019 Update (18363)
- [ ] May 2019 Update (18362)
- [ ] Creators Update (15063)

Device form factor:
- [ ] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

gregwoo-microsoft:
## Description

Fixes #867 

## Types of changes

### DIFF
--- a/XamlControlsGallery/ControlPages/TextBoxPage.xaml
+++ b/XamlControlsGallery/ControlPages/TextBoxPage.xaml
@@ -34,14 +34,14 @@
         </local:ControlExample>
         <local:ControlExample x:Name="Example3" HeaderText="A read-only TextBox with various properties set.">
             <TextBox Text="I am super excited to be here!" IsReadOnly="True" FontFamily="Arial" FontSize="24"
-                     FontStyle="Italic" CharacterSpacing="200" Foreground="#669DFF" 
+                     FontStyle="Italic" CharacterSpacing="200" Foreground="#5178BE" 
                      AutomationProperties.Name="customized TextBox"/>
             <local:ControlExample.Xaml>
                 <x:String xml:space="preserve">
 &lt;TextBox Text="I am super excited to be here!"
     AutomationProperties.Name="customized TextBox" IsReadOnly="True"
     FontFamily="Arial" FontSize="24" FontStyle="Italic"
-    CharacterSpacing="200" Foreground="#669DFF" /&gt;
+    CharacterSpacing="200" Foreground="#5178BE" /&gt;
                 </x:String>
             </local:ControlExample.Xaml>
         </local:ControlExample>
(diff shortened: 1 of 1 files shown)


XamlControlsGallery/ControlPages/TextBoxPage.xaml
```diff
@@ -34,14 +34,14 @@
         </local:ControlExample>
         <local:ControlExample x:Name="Example3" HeaderText="A read-only TextBox with various properties set.">
             <TextBox Text="I am super excited to be here!" IsReadOnly="True" FontFamily="Arial" FontSize="24"
-                     FontStyle="Italic" CharacterSpacing="200" Foreground="CornflowerBlue" 
+                     FontStyle="Italic" CharacterSpacing="200" Foreground="#669DFF"
```
chingucoding, line 37: It seems that this color does not have sufficient contrast in light theme: 
Screenshot if contrast ratio between light theme background color and the color #669DFF

gregwoo-microsoft: Good catch Marcel! Thanks for the feedback. Do you know what color would work for both light and dark themes? I tried to find one that worked for both but I don't think that it is possible. Wondering if there is a way to change the color based on the theme. Let me know what you think.

chingucoding: I am not entirely sure if there is a color that would have a contrast ratio of 4.5 to both the light and dark background. Alternatively, we could use a theme resource that would change colors based on theme. If you like, I can also dig deeper and try to find a color that could fit the criteria.

chingucoding: One thing that just came to mind is if we actually need to meet the 4.5 contrast ratio of if the 3.0 ratio is enough since we have a large font size here.

gregwoo-microsoft: That's a good point. In this case I believe that #5178BE should be good for both light and dark modes.
image
image


Let me know what you think :)

chingucoding: That colors looks good to me!

chingucoding (approved):

XamlControlsGallery/ControlPages/TextBoxPage.xaml
gregwoo-microsoft: Just updated the color code :)

chingucoding: Oh right, guess I was a bit to eager to approve the PR. LGTM to me now :)

PR #874, opened 2022-02-18
Fix TabView Sample Tear Behaviour Bug

karenbtlai:
## Description

When there is only one tab left in the TabView Windowing sample, tearing it outside of window crashes app. This is because it still closes and opens a new window when on the last tab, but an error is thrown during .Close() as you can't close the only tab in the window. 

Simply disabling `CanReorderTabs` and `CanDragTabs` when `TabItems.Count == 1` suffices.

## Motivation and Context

Fixes internal bug: 38125264

## Types of changes

### DIFF
--- a/XamlControlsGallery/TabViewPages/TabViewWindowingSamplePage.xaml.cs
+++ b/XamlControlsGallery/TabViewPages/TabViewWindowingSamplePage.xaml.cs
@@ -39,6 +39,17 @@ namespace AppUIBasics.TabViewPages
                     Window.Current.Close();
                 }
             }
+            // If there is only one tab left, disable dragging and reordering of Tabs.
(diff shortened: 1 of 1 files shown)


gregwoo-microsoft:
I agree with @gabbybilka. I think that users shouldn't be able to drag out a tab if it's the only one in the window.

karenbtlai:
Edge has it so that the tab also becomes a drag region. Unfortunately it's a more involved feature and requires more experimentation, so should be a separate fix.

ranjeshj (approved):

ranjeshj:
> Edge has it so that the tab also becomes a drag region. Unfortunately it's a more involved feature and requires more experimentation, so should be a separate fix.

@karenbtlai can you create a XCG bug for this in github ?

PR #883, opened 2022-03-17
Set up CI with Azure Pipelines

gregwoo-microsoft:


chingucoding:
Closing since it is targeting the WinUI 2 branch which is only for critical fixes.

PR #900, opened 2022-04-08
Remove the transient vs not behavior based on pointer over in CommandBarFlyout page. Can introduce confusion to narrator users

ranjeshj:
Always be transient or not for a type of interaction. Switching it based on pointer over can be confusing for keyboard/narrator users.

Internal bug: [Bug 38293416](https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/38293416): [XCG WinUI3] WinUI Accessibility->CommandBarFlyout card: See more menu list is expanding automatically when image is invoked.


bkudiess (approved):
LGTM

PR #901, opened 2022-04-08
Update name (WinUI 3 Controls Gallery --> WinUI 3 Gallery)

gregwoo-microsoft:
## Types of changes


bpulliam:
Changing the title of the application is a good idea, but we should keep the directory structure consistent between the two versions (WinUI 2 and WinUI 3) to facilitate cherry-picks and looking at differences. Let's see how we can minimize the differences between the two branches.

bpulliam:
Changing the title of the application is a good idea, but we should keep the directory structure consistent between the two versions (WinUI 2 and WinUI 3) to facilitate cherry-picks and looking at differences. Let's see how we can minimize the differences between the two branches.

gregwoo-microsoft:
@bpulliam I just updated the main folder name to "WinUIGallery" for both the WinUI 2 and the WinUI 3 Gallery. Let me know what you think!

chingucoding:
Out of curiosity, is there a good reason to have a WinUI 3 and a WinUI 2 branch? If I were to fix a bug, I think it would be easier for me if I can make one pull request that targets both the WinUI 3 and WinUI 2 version at the same time. What do you think @gregwoo-microsoft and @bpulliam ?

gregwoo-microsoft:
That's a great point and we once considered that option but there were 2 main drawbacks to doing this:
1) The main branch would take up 2x more space (as both WinUI Galleries would be there)
2) We are increasingly focused on developing the WinAppSDK so having the WinUI 3 Gallery in the main branch better reflects our current focus & investments

Let me know what you think!

gregwoo-microsoft:
That's a great point and we once considered that option but there were 2 main drawbacks to doing this:
1) The main branch would take up 2x more space (as both WinUI Galleries would be there)
2) We are increasingly focused on developing the WinAppSDK so having the WinUI 3 Gallery in the main branch better reflects our current focus & investments

Let me know what you think!

chingucoding:
Thank you for the insights, that makes sense. Given those drawbacks, I think naming the both the same and keeping them on separate branches is a good way to move forward here!

gregwoo-microsoft:
Closing this as it is a copy

PR #910, opened 2022-04-18
Re-enable AnimatedVisualPlayer sample

kmahone:
The AnimatedVisualPlayer sample was disabled because the required Win2D component was not available at the time for WinUI3. The [Microsoft.Graphics.Win2D](https://www.nuget.org/packages/Microsoft.Graphics.Win2D) nuget package is available and supports what we need.


ranjeshj (approved):
:shipit:

PR #926, opened 2022-05-03
Add a page for system backgrounds / mica

bpulliam:
## Description
This change adds a "System Backdrops (Mica/Acrylic)" page in XCG. Some details:

The System Backdrops page has "(Mica/Acrylic)" in the title so it shows up if the user types Mica or Acrylic in the XCG search box.
There are ~100 lines of code required in the app to use Mica or Acrylic. (We want to make this much easier in WinAppSDK 1.2.) The page shows all of the C# code to do this.
The page has a couple buttons to launch a separate window which shows Mica/Acrylic. We could change the XCG window itself to have transparency where needed to make this work, but I wanted something more self-contained which developers can more easily copy for their app.
The sample currently accounts for the Window getting activated/deactivated and the initial Light/Dark theme of the app. It does not have an ActualThemeChanged handler to support the theme changing, nor does it check for high contrast or watch for high contrast changes.
The sample shows the need to explicitly Dispose() (or Close() in C++) the MicaController/DesktopAcrylicController before the window closes. If this isn't done, then the controller will likely try to get/set the underlying SystemBackdrop property on the window, which will crash when called too late on window close.
The sample currently does not call DispatcherQueueController.ShutdownQueueAsync -- we just release our reference when the window closes and leave it running. (I think we'd need to duplicate the IDispatcherQueueController interface into C# if we wanted to call this, and would need additional code to ref-count to know when we are ready to do this.)

## Motivation and Context
Show the use of a new feature in WinAppSDK 1.1-preview 3!

## Types of changes


MikeHillberg (approved):

XamlControlsGallery/ControlPagesSampleCode/SystemBackdrops/SystemBackdropsSample1.txt
```diff
@@ -0,0 +1,58 @@
...
+    // Make sure any Mica/Acrylic controller is disposed so it doesn't try to
+    // use this closed window.
+    if (m_micaController != null)
+    {
+        m_micaController.Dispose();
+        m_micaController = null;
+    }
+    this.Activated -= Window_Activated;
+    m_configurationSource = null;
+}
```
dotMorten, line 57: The C++ sample closes `dispatcherQueueController` by calling `ShutdownQueueAsync` https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/system-backdrop-controller
Is this not needed for C#?

codendone: Ideally the call would be made to ensure clean shutdown, but in practice it isn't required as long as the *Controller object is Closed/Disposed. Making the `ShutdownQueueAsync` call is non-trivial because it would require defining the COM interface in C#. The code here opts for simplicity.

PR #934, opened 2022-05-09
WinUI 2 Gallery Visual Revamp

gregwoo-microsoft:
## Types of changes


XamlControlsGallery/App.xaml
```diff
@@ -84,8 +84,8 @@
 
             <SolidColorBrush x:Key="GridViewHeaderItemDividerStroke" Color="Transparent"/>
 
-            <x:String x:Key="ControlsName">All controls</x:String>
-            <x:String x:Key="NewControlsName">What's New</x:String>
+            <x:String x:Key="AppTitleName">WinUI 2 Gallery</x:String>
+            <x:String x:Key="WinUIVersion">SDK X.X</x:String>
```
chingucoding, line 88: I think we should specifiy the WinUI 2 version being used. Unfortunately, I didn't find a way we could extrapolate this during runtime to avoid having to manually do this though I haven't looked at this that deep.

```suggestion
            <x:String x:Key="WinUIVersion">SDK 2.8</x:String>
```

gregwoo-microsoft: Yes I agree with you!

bpulliam (approved):

PR #935, opened 2022-05-11
Fix project TargetFramework

bpulliam:
## Description
Provide a default TargetFrameworkMoniker based on the version of VS
Fix the AppxBundle warning.
Sync IsUpdated values
Whitespace

### DIFF
--- a/XamlControlsGallery/ControlPages/ImplicitTransitionPage.xaml
+++ b/XamlControlsGallery/ControlPages/ImplicitTransitionPage.xaml
@@ -13,7 +13,7 @@
         </Style>
     </Page.Resources>
     <StackPanel Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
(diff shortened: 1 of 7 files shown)


XamlControlsGallery/ControlPages/ImplicitTransitionPage.xaml
```diff
@@ -13,7 +13,7 @@
         </Style>
     </Page.Resources>
     <StackPanel Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
-        <local:ControlExample HeaderText="Automatically animate changes to Opacity" >
+        <local:ControlExample HeaderText="Automatically animate changes to Opacity" MinimumUniversalAPIContract="7">
```
chingucoding, line 16: Why do we need this? Isn't the point of WinUI 3 that we are detached from the OS or is this one of the areas where this does not apply?

chingucoding (approved):

PR #946, opened 2022-05-14
CustomTitlebar --> Titlebar

gregwoo-microsoft:
Renaming the titlebar files to be consistent with the naming of other controls

## Types of changes


WinUIGallery/WinUIGallery.DesktopWap.csproj
```diff
@@ -86,5 +86,9 @@
     <Content Remove="@(Content)" />
   </ItemGroup>
   <Import Project="ContentIncludes.props" />
+  <ItemGroup>
```
bpulliam, line 89: Why are you removing these? This shouldn't be necessary.

bpulliam (changes requested):

WinUIGallery/WinUIGallery.DesktopWap.csproj
gregwoo-microsoft: You're right. This got added automatically by Visual Studio. Removing this now.

bpulliam (approved):

PR #953, opened 2022-05-18
Update protocol launcher to work with doc pages

gregwoo-microsoft:
## Types of changes


ocalvo (approved):

ocalvo (changes requested):
Why are we adding back the "3" as version number? Wouldn't this be confusing for customers?

gregwoo-microsoft:
@ocalvo this is because I'm working with the Documentation team to update the names of the gallery apps and they have links to open a specific app on a particular page (e.g. winui2gallery:/item/Expander to open the WinUI 2 Gallery to the expander page). We need to provide a unique protocol for each app for this to happen.

PR #970, opened 2022-05-24
Fixed titlebar buttons when changing theme

gregwoo-microsoft:
Title bar buttons are set to Black when in Light mode and vice versa for dark mode

## Types of changes

### DIFF
--- a/WinUIGallery/Helper/TitlebarHelper.cs
+++ b/WinUIGallery/Helper/TitlebarHelper.cs
@@ -21,17 +21,10 @@ namespace WinUIGallery.DesktopWap.Helper
     internal class TitlebarHelper
     {
 
-        [ComImport, Guid("EECDBF0E-BAE9-4CB6-A68E-9598E1CB57BB"), InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
-        private interface IWindowNative
-        {
-            IntPtr WindowHandle { get; }
-        };
-
         public static void triggerTitleBarRepaint()
         {
             // to trigger repaint tracking task id 38044406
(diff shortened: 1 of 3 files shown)


WinUIGallery/SettingsPage.xaml.cs
```diff
@@ -120,6 +129,22 @@ private void OnThemeRadioButtonChecked(object sender, RoutedEventArgs e)
                     }
                 }
             }
```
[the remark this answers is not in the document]
gregwoo-microsoft: Great suggestion! I just made the change.

WinUIGallery/Helper/TitlebarHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.UI.Xaml.Navigation;
+using WinRT;
+using System.Runtime.InteropServices;
+
+namespace WinUIGallery.DesktopWap.Helper
+{
+    internal class TitlebarHelper
+    {
+
+        [ComImport, Guid("EECDBF0E-BAE9-4CB6-A68E-9598E1CB57BB"), InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
```
marb2000, line 24: You don't need this ComImport. You can use WinRT.Interop.WindowNative.GetWindowHandle(window) to get the hwnd.

marb2000 (changes requested):

WinUIGallery/Helper/TitlebarHelper.cs
gregwoo-microsoft: Thanks for the tip @marb2000! I updated the code.

PR #984, opened 2022-06-02
update windows version list

### LINKED ISSUE #969: GitHub templates out-of-date
**Describe the bug**
Windows versions are missing, doesn't differentiate between Win 10/11.

GitHub Issue templates should be updated and use new [GitHub form feature](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) for easier flow and entry.

bpulliam:
## Description
The bug template was slightly out-of-date on the current in-market Windows versions.  I chose not to update it from markdown to yaml with this change (but anyone is welcome to do that).

## Motivation and Context
It is no longer 2019.
Fixes #969 

## Types of changes

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -28,15 +28,21 @@ Steps to reproduce the behavior:
 Gallery version (found on Settings page): 
 <!-- [e.g. 1.2.3.0] -->
 
-Windows 10 version:
+WinUI version:
+- [ ] WinUI3 / WinAppSDK 1.1
+- [ ] WinUI3 / WinAppSDK 1.0
+- [ ] WinUI2
+
+Windows 11 version:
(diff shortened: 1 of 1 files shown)


gregwoo-microsoft (approved):
Looks great! Should we also include WinAppSDK 0.8?

michael-hawker:
@bpulliam @gregwoo-microsoft I had also suggested in the issue to switch to the new issue form template system which makes things a lot easier for folks to fill out and provide info (think drop-down vs. messing with putting 'x' in the '[ ]' which is error-prone).

I know that's a bit more work, but it's worth it in the long-run (also harder for folks to just delete the whole issue template and provide less detail).

Did you want to track that in a new issue?

gregwoo-microsoft:
That's a great idea @michael-hawker. I've created an issue https://github.com/microsoft/WinUI-Gallery/issues/992 to track this.

PR #989, opened 2022-06-04
Add debug manifest file for development

chingucoding:
Similar to the WinUI 2 gallery, this PR adds a debug manifest that allows you to have the Store version and the development versions installed at the same time instead of the two versions overriding each other all the time.

Question: Right now the app would still display WinUI 3 Gallery when being started, is that good?

### DIFF
--- a/WinUIGallery/Package.WAP.Debug.appxmanifest
+++ b/WinUIGallery/Package.WAP.Debug.appxmanifest
@@ -1,0 +1,55 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Package
+    xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+    xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest"
+    xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+    xmlns:uap3="http://schemas.microsoft.com/appx/manifest/uap/windows10/3"
+    xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+    IgnorableNamespaces="uap mp uap3">
+  <Identity Name="Microsoft.WinUI3ControlsGallery.Debug" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="1.3.13.0" />
+  <mp:PhoneIdentity PhoneProductId="863667e0-667a-4bb4-ac52-c59656c7333a" PhonePublisherId="00000000-0000-0000-0000-000000000000" />
+  <Properties>
+    <DisplayName>WinUI 3 Gallery</DisplayName>
+    <PublisherDisplayName>Microsoft Corporation</PublisherDisplayName>
+    <Logo>Assets\Tiles\StoreLogo-sdk.png</Logo>
+    <uap:SupportedUsers>multiple</uap:SupportedUsers>
+  </Properties>
+  <Dependencies>
+    <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.17134.0" MaxVersionTested="10.0.18362.0" />
+    <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.17134.0" MaxVersionTested="10.0.18362.0" />
+  </Dependencies>
+  <Resources>
+    <Resource Language="x-generate" />
+  </Resources>
+  <Applications>
+    <Application Id="App" Executable="$targetnametoken$.exe" EntryPoint="$targetentrypoint$">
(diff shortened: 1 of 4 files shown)


WinUIGallery/Package.WAP.Debug.appxmanifest
```diff
@@ -0,0 +1,55 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Package
+    xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+    xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest"
+    xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+    xmlns:uap3="http://schemas.microsoft.com/appx/manifest/uap/windows10/3"
+    xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+    IgnorableNamespaces="uap mp uap3">
+  <Identity Name="Microsoft.WinUI3ControlsGallery.Debug" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="1.3.13.0" />
```
bpulliam, line 9: Suggestion: use Dev instead of Debug in both the name and filename.  Following the Terminal model, we can introduce branding (not necessarily in this PR) so that "WinUI 3 Gallery" is always Release, "WinUI 3 Gallery Dev" is the default and can be Debug or Release

bpulliam (approved):

WinUIGallery/Package.WAP.Debug.appxmanifest
chingucoding: I see what you are going after here. So would it be better to rename this to dev and make it the default for both debug and release and keep the "non-dev" version up just for building the store version?

bpulliam: exactly :)

WinUIGallery/Package.WAP.Dev.appxmanifest
```diff
@@ -0,0 +1,55 @@
...
+  <Identity Name="Microsoft.WinUI3ControlsGallery.Debug" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="1.3.13.0" />
+  <mp:PhoneIdentity PhoneProductId="0ba9d13a-a477-463f-955f-91d935fa4cd5" PhonePublisherId="00000000-0000-0000-0000-000000000000" />
+  <Properties>
+    <DisplayName>WinUI 3 Gallery Dev</DisplayName>
+    <PublisherDisplayName>Microsoft Corporation</PublisherDisplayName>
+    <Logo>Assets\Tiles\StoreLogo-sdk.png</Logo>
+    <uap:SupportedUsers>multiple</uap:SupportedUsers>
+  </Properties>
+  <Dependencies>
+    <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.17134.0" MaxVersionTested="10.0.18362.0" />
```
bpulliam, line 18: Can remove Universal (see #1005)

bpulliam (approved):

WinUIGallery/Package.WAP.Dev.appxmanifest
```diff
@@ -0,0 +1,53 @@
...
+  </Properties>
+  <Dependencies>
+    <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.17134.0" MaxVersionTested="10.0.18362.0" />
+  </Dependencies>
+  <Resources>
+    <Resource Language="x-generate" />
+  </Resources>
+  <Applications>
+    <Application Id="App" Executable="$targetnametoken$.exe" EntryPoint="$targetentrypoint$">
+      <uap:VisualElements DisplayName="WinUI 3 Gallery Dev" Square150x150Logo="Assets\Tiles\squaretile-sdk.png" Square44x44Logo="Assets\Tiles\SmallTile-sdk.png" Description="WinUI 3 Gallery Debug" BackgroundColor="transparent">
```
michael-hawker, line 25: Huh, I was able to build from this branch, but I didn't see the 'Dev' title reflect on the running app, was going to come and suggest this, as having a visual distinction is really useful (we did this in the Toolkit as well):

image

michael-hawker: Ah, I see it in the start menu, but that's not helpful if you run the apps side-by-side and are trying to compare (and if you're not running in debug mode or turn the XAML toolbar off, you can lose track of which is which).

image

chingucoding: Very good paint, to be honest forgot a bit about that. But now it should be also display it in the title of the app.

PR #1009, opened 2022-06-12
Fix bug where clicking sounds link on settings page would crash the app

chingucoding:
## Description

The issue was that the navigation parameters where not was what expected.
## How Has This Been Tested?

Tested manually.

## Types of changes

### DIFF
--- a/WinUIGallery/SettingsPage.xaml.cs
+++ b/WinUIGallery/SettingsPage.xaml.cs
@@ -33,6 +33,8 @@ namespace AppUIBasics
     /// </summary>
     public sealed partial class SettingsPage : Page
     {
+        private NavigationRootPage navigationRootPage;
+
         public string Version
         {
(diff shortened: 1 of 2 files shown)


WinUIGallery/SettingsPage.xaml.cs
```diff
@@ -80,7 +83,7 @@ private void OnSettingsPageLoaded(object sender, RoutedEventArgs e)
             NavigationRootPage navigationRootPage = NavigationRootPage.GetForElement(this);
```
bpulliam, line 83: This declaration will cover the newly declared private field. Either use the field or change the variable name. Same for line 139 (use GetForElement(this) or cached variable?)

bpulliam (changes requested):

bpulliam (approved):

Issue #649, opened 2021-03-03
Cannot compile and run Xaml-Controls-Gallery


sketch34:
**Is this a totally new sample, an addition to an existing sample page, or a modification of an existing sample page?**
- [ ] New sample
- [ ] Addition to existing sample
- [x] Modification to existing sample

**If this is an addition/modification to an existing sample page, which one?**
How to correctly compile / run the Xaml-Controls-Gallery solution.

**Describe the sample**
Compile and run the Xaml-Controls-Gallery solution on Windows 10 with a clean install of latest Visual Studio Community 2019 Preview 16.10.0 Preview 1 and required dependencies following the instructions here: https://docs.microsoft.com/en-au/windows/apps/winui/winui3/

I'm on `winui3preview` branch.

**Is your sample request related to a problem? Please describe it**
I'm having problems compiling and running the solution. I would love a from-scratch / clean install tutorial showing step by step how to get this sample project compiling and running. Currently blocked on this.

1. Had to bump the Microsoft.Net.Compilers.Toolset package up to 3.9.0 because "3.8.X.final" was failing the test ">= 3.8.0-4.20472.6". I'm guessing because ".final" is not numeric and ruins the internal version comparison. Couldn't see any other way around this.

![image](https://user-images.githubusercontent.com/1082293/109756143-769f5400-7c3b-11eb-9e56-825c2d5e4bd3.png)

2. "NuGet package restore failed. Please see Error List window for detailed warnings and errors." - I have NuGet.org package source configured. No additional info is in the Error List window.
2. Which configuration is correct for Win 10 desktop?

![image](https://user-images.githubusercontent.com/1082293/109753959-45bd2000-7c37-11eb-9949-1e081c166917.png)

4. Compilation succeeds but deployment fails for x64, x86, Win32. It is correct that it is trying to copy to an ARM64 output folder when I have not chosen the ARM build configuration? Indeed that output older does not exist.
```
Copying file "C:\Users\jarro\.nuget\packages\microsoft.winui\3.0.0-preview4.210210.4\runtimes\win10-arm64\native\Microsoft.UI.Xaml\Assets\NoiseAsset_256x256_PNG.png" to layout "XamlControlsGallery.Desktop\runtimes\win10-arm64\native\Microsoft.UI.Xaml\Assets\NoiseAsset_256x256_PNG.png"...
[...]

PR #1015, opened 2022-06-16
Update BreadcrumbBar samples

### LINKED ISSUE #1000: BreadcrumbBar sample lacking context and navigation
- [ ] New sample

BreadcrumbBar

**Describe the sample**
The current `BreadcrumbBar` sample is fairly lacking in details. There are two examples. In the first, all it shows for 'source code' is XAML without the data to drive it (the list of strings here):

https://github.com/microsoft/WinUI-Gallery/blob/55af91f5993029d073ef5129bbc07b7b908c2ebe/WinUIGallery/ControlPages/BreadcrumbBarPage.xaml.cs#L13

Then the second example at least shows how to work with a custom data type, but then doesn't show how to maintain the same click behavior in order to navigate when a custom item is clicked from the template.

(I've also noticed this with NavigationViewItem, but it seems odd to have to create the container as part of the DataTemplate as done here in the 2nd example as well. Usually things like `DisplayMemberPath` were how these things connected in WPF, so I feel like these two controls are trying to warp a known pattern?)

Hard to understand how BreadcrumbBar works and can be used in an application to provide contextual navigation with custom data objects.

**Additional context**

chingucoding:
## Description

Added the C# sample code to the samples, updated the second sample to show navigation behavior.
## Motivation and Context

Closes #1000 
## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/BreadcrumbBarPage.xaml
+++ b/WinUIGallery/ControlPages/BreadcrumbBarPage.xaml
@@ -26,6 +26,11 @@
 &lt;BreadcrumbBar x:Name="BreadcrumbBar1"/&gt;
                 </x:String>
             </local:ControlExample.Xaml>
+            <local:ControlExample.CSharp>
+                <x:String xml:space="preserve">
+BreadcrumbBar1.ItemsSource = new string[] { "Home", "Documents", "Design", "Northwind", "Images", "Folder1", "Folder2", "Folder3" };
+                </x:String>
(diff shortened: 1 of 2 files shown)


michael-hawker:
Hey @chingucoding I think looks good, but wanted to take it for a spin to see the new behavior, but had issues building the gallery: https://github.com/microsoft/WinUI-Gallery/issues/649#issuecomment-1158242231

gregwoo-microsoft (approved):
Thank you @chingucoding!

PR #1046, opened 2022-07-14
Remove LocalizedControlType from Elements Automation Name

bkudiess:
Removes the control type from the automation name, so narrator doesn't read repetitive and confusing information.
Fixes issue where two Elements of same control type had same automation name.

Fixes internal bug 40234538 and 40234974

### DIFF
--- a/WinUIGallery/ControlPages/CheckBoxPage.xaml
+++ b/WinUIGallery/ControlPages/CheckBoxPage.xaml
@@ -18,7 +18,7 @@
             <local:ControlExample.Example>
                 <StackPanel Orientation="Horizontal">
                     <!-- A two-state CheckBox. -->
-                    <CheckBox Content="Two-state CheckBox" Checked="Control1_Checked" Unchecked="Control1_Unchecked" />
+                    <CheckBox Content="Two-state CheckBox" Checked="Control1_Checked" Unchecked="Control1_Unchecked" AutomationProperties.Name="Two-state" />
                 </StackPanel>
             </local:ControlExample.Example>
             <local:ControlExample.Output>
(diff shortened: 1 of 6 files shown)


WinUIGallery/ControlPages/CheckBoxPage.xaml
```diff
@@ -18,7 +18,7 @@
             <local:ControlExample.Example>
                 <StackPanel Orientation="Horizontal">
                     <!-- A two-state CheckBox. -->
-                    <CheckBox Content="Two-state CheckBox" Checked="Control1_Checked" Unchecked="Control1_Unchecked" />
+                    <CheckBox Content="Two-state CheckBox" Checked="Control1_Checked" Unchecked="Control1_Unchecked" AutomationProperties.Name="Two-state" />
```
bpulliam, line 21: We should either use Two-state or Control1 consistently.  Pick one :)

WinUIGallery/ControlPages/CheckBoxPage.xaml
```diff
@@ -36,7 +36,8 @@
             <local:ControlExample.Example>
                 <StackPanel Orientation="Horizontal">
                     <CheckBox Content="Three-state CheckBox" IsThreeState="True" Indeterminate="Control2_Indeterminate"
-                            Checked="Control2_Checked" Unchecked="Control2_Unchecked" />
+                              Checked="Control2_Checked" Unchecked="Control2_Unchecked"
```
bpulliam, line 39: We should either use Three-state or Control2 consistently.  Pick one :)

WinUIGallery/ControlPages/RevealFocusPage.xaml
```diff
@@ -140,7 +140,7 @@
                          />
 
                 <ComboBox MinWidth="300" Margin="0,16,0,0" 
-                          AutomationProperties.Name="Sample ComboBox"
+                          AutomationProperties.Name="Sample"
```
bpulliam, line 143: shouldn't the names be unique? (lines 143 & 165)

bkudiess: In this case as they are different control types, narrator will read: "Sample combobox" and "Sample slider", so users won't be confused.

PR #1070, opened 2022-09-06
Add Support MicaAlt

ghost1372:
Add Support MicaAlt Backdrop

## Description
more info here

https://docs.microsoft.com/en-us/windows/apps/design/style/mica

image

image

## Types of changes

### DIFF
--- a/WinUIGallery/SamplePages/SampleSystemBackdropsWindow.xaml.cs
+++ b/WinUIGallery/SamplePages/SampleSystemBackdropsWindow.xaml.cs
@@ -1,4 +1,4 @@
-﻿using Windows.Foundation.Metadata;
+using Windows.Foundation.Metadata;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media.Animation;
@@ -100,7 +100,7 @@ namespace AppUIBasics.SamplePages
 
             if (type == BackdropType.Mica)
             {
-                if (TrySetMicaBackdrop())
+                if (TrySetMicaBackdrop(false))
                 {
                     tbCurrentBackdrop.Text = "Mica";
                     m_currentBackdrop = type;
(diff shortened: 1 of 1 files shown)


WinUIGallery/SamplePages/SampleSystemBackdropsWindow.xaml.cs
```diff
@@ -152,6 +176,16 @@ bool TrySetMicaBackdrop()
...
 
+        bool TrySetMicaAltBackdrop()
+        {
+            return TrySetMicaBackdrop(true)
+        }
+        
+        bool TrySetMicaBackdrop()
+        {
+            return TrySetMicaBackdrop(false)
+        }
```
chingucoding, line 187: I'm inclined to just inline those method calls instead of having those two functions (which don't even compile right now).

WinUIGallery/SamplePages/SampleSystemBackdropsWindow.xaml.cs
```diff
@@ -126,7 +141,7 @@ public void SetBackdrop(BackdropType type)
             }
         }
 
-        bool TrySetMicaBackdrop()
+        bool TrySetMicaBackdrop(bool isMicaAlt)
```
chingucoding, line 144: nit: Rename to `useMicaAlt`

WinUIGallery/SamplePages/SampleSystemBackdropsWindow.xaml.cs
ghost1372: Tnx, Done

WinUIGallery/SamplePages/SampleSystemBackdropsWindow.xaml.cs
ghost1372: Done

chingucoding (approved):
Looks good to me, @gregwoo-microsoft what are your thoughts?

PR #1106, opened 2022-10-19
Update Gallery glyphs to current standard

bpulliam:
## Description
Changes:
- Update Delete from E107 to E74D
- Update Accept from E10B to E8FB
- Update Flag from E129 to E7C1
- In sample 4, fixed the Lock (was using Permissions instead) to E72E
- Fixed the sample code to match actual markup

## Motivation and Context
As mentioned in microsoft/microsoft-ui-xaml#7850, there is a note in the documentation to use Unicode points E700+ when referencing glyphs.  This change only changes the specific references in the WinUI Gallery code itself.

## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/SwipeControlPage.xaml
+++ b/WinUIGallery/ControlPages/SwipeControlPage.xaml
@@ -26,8 +26,8 @@
                 <local:ControlExample.Example>
                     <Border>
                         <Border.Resources>
-                            <FontIconSource x:Key="AcceptIcon" Glyph="&#xE10B;"/>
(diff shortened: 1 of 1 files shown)


Jay-o-Way:
Nice. Are these files used for WinUI2 or Winui3 Gallery - or both?

chingucoding (approved):

chingucoding:
@Jay-O-Way These changes are targeting the WinUI 3 Gallery, only PRs targeting the winui2 branch are used by the WinUI 2 Gallery.

Issue #121, opened 2019-07-23
ScrollViewer page is confusing


chigy:
**Describe the bug**

ScrollViewer sample is confusing because the page doesn't even show scroll indicator or scroll bar unless I zooomed out the image. Perhaps better representation of scroll viewer is better here?

**Steps to reproduce the bug**

**Expected behavior**

**Actual behavior**

**Screenshots**

**Version Info**

Gallery version (found on Settings page): 

Windows 10 version:
- [ ] Insider Build (xxxxx)
- [x] May 2019 Update (18362)
- [ ] October 2018 Update (17763)
- [ ] April 2018 Update (17134)
- [ ] Fall Creators Update (16299)
- [ ] Creators Update (15063)
- [ ] Anniversary Update (14393)

Device form factor:
- [ ] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

chingucoding:
I agree that the "normal" ScrollViewer page is confusing. Since there is the "ScrollViewer (Preview)" page which has more and better examples.
It is pretty confusing to see that there are 2 Pages for the same control, so maybe we should delete the normal page and leave the preview page in?

chingucoding:
Given that the "ScrollViewer (Preview)" page no longer exists in the WinUI 3 gallery, I think we should update the ScrollViewer page. @gregwoo-microsoft what are your thoughts on this?

gregwoo-microsoft:
I agree, this page is not intuitive when you first launch it because you have to zoom in to see the scrollviewer. A simple fix could be to start the page with a Zoom = 2. What do you think @chingucoding ?

chingucoding:
Seems like a good idea, that should fix the sample.

PR #1113, opened 2022-11-09
Scroll viewer zoom to 2.0 on startup and rename control x:Name

baileydauterman:
## Description
Updated the scroll viewer starting zoom to be 2.0 so when users select the control it is more clear what the scroll viewer does. Also renamed the x:name for the scrollviewer form `Control1` to `ScrollViewerControl` to make the name more meaningful.

## Motivation and Context
This is a PR to fix issue #121. I picked this because it was the only issue marked as `good first issue` and was a relatively easy fix.

## How Has This Been Tested?
I believe this issue is isolated to the scroll viewer page. If there are tests that need updated I can do this.

## Types of changes


bpulliam:
Nice first fix!

PR #1116, opened 2022-11-10
Merge Feature/winappsdk 1.2 into main

bpulliam:
## Description
This merges the feature branch that has captured the road to 1.2 release.
Now that 1.2 stable has been released, we can bring this into main.

## Types of changes

### DIFF
--- a/WinUIGallery/App.xaml
+++ b/WinUIGallery/App.xaml
@@ -86,7 +86,7 @@
 
             <x:String x:Key="ControlsName">All controls</x:String>
             <x:String x:Key="AppTitleName">WinUI 3 Gallery</x:String>
-            <x:String x:Key="WinUIVersion">WinAppSDK 1.1</x:String>
+            <x:String x:Key="WinUIVersion">WinAppSDK 1.2</x:String>
 
             <Style x:Key="OutputTextBlockStyle" BasedOn="{StaticResource BodyTextBlockStyle}" TargetType="TextBlock">
                 <Setter Property="Margin" Value="8,8,0,0" />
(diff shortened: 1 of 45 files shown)


gregwoo-microsoft (approved):
Works great! The only thing that I would change is adding the MediaPlayerElement control to the "recently added samples" list.

gregwoo-microsoft:
In fact, here is the list of all of the controls that were updated since the last update:
AnimatedIcon
AutoSuggestBox
Button
CheckBox
ListView
RevealFocus
ScrollViewer
Slider
SplitButton
SplitView
Swipe
TitleBar
WebView2
SystemBackdrops

chingucoding (commented):
What's up with the files in the `Helpers\Colorcode.internal` folder? Visual Studio does not seem to show them and they don't seem to be referenced (unless I'm missing something here).

bpulliam:
> What's up with the files in the `Helpers\Colorcode.internal` folder? Visual Studio does not seem to show them and they don't seem to be referenced (unless I'm missing something here).

Colorcode.WinUI is built against the shipping version of WinUI and causes conflicts when WinUI Gallery is built against internal builds of WinUI. One of the tasks we have been working on is eliminating copies of the Gallery code and the cost of keeping them in sync. As such, we use the public WinUIGallery code to validate and develop WinUI itself. The Colorcode.Internal allows us to use Colorcode.Core directly when building WinUI Gallery internally, but it is not used in external builds. The switch is in the csproj file by changing the reference and using compile removes.

chingucoding:
Ah I see, thanks for the clarification @bpulliam!

PR #1131, opened 2022-12-02
New tile carousel on home page + tiles for community & accessibility

gregwoo-microsoft:
- Adds a reference to the Community Toolkit
- Creates better integration between our various dev tools
- Created a folder for all of the header tiles
image

## Types of changes


gregwoo-microsoft:
FYI @niels9001

WinUIGallery/NewControlsPage.xaml
```diff
@@ -114,10 +114,11 @@
...
-                            <controls:HeaderTile Title="Template Studio" Source="/Assets/template_studio.png" Link="https://marketplace.visualstudio.com/items?itemName=TemplateStudio.TemplateStudioForWinUICs"/>
+                            <controls:HeaderTile Title="Getting Started" Source="/Assets/HomeHeaderTiles/gettingStarted.png" Link="https://docs.microsoft.com/windows/apps/get-started/"/>
+                            <controls:HeaderTile Title="Windows Design" Source="/Assets/HomeHeaderTiles/style-1x.png" Link="https://docs.microsoft.com/windows/apps/design/"/>
+                            <controls:HeaderTile Title="WinUI Repo" Source="/Assets/HomeHeaderTiles/githubTile.png" Link="https://github.com/microsoft/microsoft-ui-xaml"/>
+                            <controls:HeaderTile Title="Template Studio" Source="/Assets/HomeHeaderTiles/template_studio.png" Link="https://marketplace.visualstudio.com/items?itemName=TemplateStudio.TemplateStudioForWinUICs"/>
+                            <controls:HeaderTile Title="Community Toolkit" Source="/Assets/HomeHeaderTiles/toolkit.png" Link="https://apps.microsoft.com/store/detail/windows-community-toolkit-sample-app/9NBLGGH4TLCQ"/>
```
niels9001, line 121: You can use this link: https://aka.ms/toolkit/windows

This brings the user directly to the repo, which has a link to the Gallery as well.

chingucoding (approved):

gregwoo-microsoft:
@niels9001 @chingucoding  let me know what you think, I added a new flipview/carousel functionality + a new accessibility tile

chingucoding:
I think we should recalculate the number of items inside the FlipView based on the width of the app and maybe also use a PipsPager at the bottom? Let me know what you think @gregwoo-microsoft.

gregwoo-microsoft:
Yeah that is a great idea @chingucoding. I can look to add a PipsPager tomorrow. Do you know how to adjust the FlipView based on the app's width?

chingucoding:
I think the best way would be to listen to the FlipViews SizeChanged event and recalculate how many items to show based on that.

niels9001:
@gregwoo-microsoft @chingucoding I've created the following: https://github.com/gregwoo-microsoft/WinUI-Gallery/pull/2

This uses a scrollviewer and will show the scroll buttons whenever the is more content to scroll to. This is also how Store handles its screenshots section and is maybe easier than moving items around in a FlipView😊? (Would be nice to have some sort of adaptive FlipView that can stretch to show more content vs. flipping if not).


Tiles gallery

chingucoding:
That would also work but I think having a PipsPager and FlipView for this would look better even if it is more complicated to set up. Maybe we could even let it switch the view every x seconds?  What are your thoughts @gregwoo-microsoft @niels9001 ?

niels9001:
> That would also work but I think having a PipsPager and FlipView for this would look better even if it is more complicated to set up. Maybe we could even let it switch the view every x seconds? What are your thoughts @gregwoo-microsoft @niels9001 ?

Hmm.. looking at the Store homepage carousel and screenshots gallery on the PDP, they seem to favor a ScrollViewer vs. a FlipView. My worry with a FlipView is that you might end up with e.g. 4 items on the first item and 2 on the second item? Not sure if that'd look great. Also discoverability might not be the greatest as the PipsPager might be missed?

I'd say, in the end this might be an interesting opportunity to recreate something like the Store carousel - this would also allow the landingspage to be more engaging by providing big images with additional information :).

chingucoding:
Alright, lets do that then. Not sure about the big pictures in the future but I think we'll figure that out then.

gregwoo-microsoft:
Love the change @niels9001. Thanks for implementing that. I just merged your PR.

niels9001:
@gregwoo-microsoft Can we get this in?

PR #1152, opened 2023-01-05
add a readme to packagestore to ensure its existence

bpulliam:
## Description
With the recent changes to nuget.config, I added a directory that can be used to easily test nuget packages locally without having to configure a feed.  That change did assume that the directory existed.  Now it does :)
Thanks for VinJa for pointing this out.

## Types of changes

### DIFF
--- a/packagestore/readme.md
+++ b/packagestore/readme.md
@@ -1,0 +1,1 @@
+Local source for packages. Places nupkg files in this directory to test them out before uploading.


chingucoding (approved):

chingucoding:
Alternatively, we could also put a .gitignore in that directory so you also don't need to think about accidently checking in a NuGet package.

gregwoo-microsoft (approved):

PR #1163, opened 2023-01-10
Added the Background property to the ProgressRing page

gregwoo-microsoft:
## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/ProgressRingPage.xaml
+++ b/WinUIGallery/ControlPages/ProgressRingPage.xaml
@@ -25,7 +25,7 @@
             <local:ControlExample.Options>
                 <StackPanel>
                     <ToggleSwitch x:Name="ProgressToggle" AutomationProperties.Name="Progress options" Header="Toggle work" OffContent="Do work" OnContent="Working" IsOn="True" />
-                    <ComboBox x:Name="BackgroundComboBox1" SelectionChanged="Background_SelectionChanged1"  Header="Background color" PlaceholderText="Pick a color" Width="200">
+                    <ComboBox x:Name="BackgroundComboBox1" SelectionChanged="Background_SelectionChanged"  Header="Background color" PlaceholderText="Pick a color" Width="200">
                         <x:String>Transparent</x:String>
                         <x:String>LightGray</x:String>
                     </ComboBox>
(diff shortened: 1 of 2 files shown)


WinUIGallery/ControlPages/ProgressRingPage.xaml
```diff
@@ -46,13 +55,22 @@
...
+                    <ComboBox x:Name="BackgroundComboBox2" SelectionChanged="Background_SelectionChanged2"  Header="Background color" PlaceholderText="Pick a color" Width="200">
+                        <x:String>Transparent</x:String>
+                        <x:String>LightGray</x:String>
+                    </ComboBox>
+                </StackPanel>
+            </local:ControlExample.Options>
             <local:ControlExample.Xaml>
                 <x:String>
-                    &lt;ProgressRing Width="60" Height="60" Value="$(DeterminateProgressValue)" IsIndeterminate="False"/&gt;
+                    &lt;ProgressRing Width="60" Height="60" Value="$(DeterminateProgressValue)" IsIndeterminate="False" $(Background)/&gt;
```
chingucoding, line 68: I think a line break for this would be good so that we don't have vertical scrolling.

WinUIGallery/ControlPages/ProgressRingPage.xaml.cs
```diff
@@ -43,5 +44,45 @@ private void ProgressValue_ValueChanged(Microsoft.UI.Xaml.Controls.NumberBox sen
...
+                    showBackgroundProperty = true;
+                    break;
+
+                default:
+                    throw new Exception($"Invalid argument: {colorName}");
+            }
+            RevealBackgroundProperty1.IsEnabled = showBackgroundProperty;
+        }
+
+        private void Background_SelectionChanged2(object sender, SelectionChangedEventArgs e)
```
chingucoding, line 67: Could we refactor this to determine only have one handler that selects the correct progressring based on the sender?

gregwoo-microsoft: I'm not sure how we can get that information from the sender. Any idea?

WinUIGallery/ControlPages/ProgressRingPage.xaml
gregwoo-microsoft: Fixed! :)

WinUIGallery/ControlPages/ProgressRingPage.xaml.cs
gregwoo-microsoft: but I agree, let's figure this out before merging it

chingucoding: We can either set the tag property on the ComboBox and retrieve that through the sender property or doing something like this:

```c#
private void Background_SelectionChanged2(object sender, SelectionChangedEventArgs e)
    var progressRing = sender == BackgroundComboBox1 ? ProgressRing1 : ProgressRing2;
    // Color progress ring accordingly
}
```

gregwoo-microsoft: That's perfect! Thank you @chingucoding. I just pushed the change.

chingucoding (approved):

PR #1167, opened 2023-01-13
Design Guidance section [Work In Progress]

gregwoo-microsoft:
## Types of changes


WinUIGallery/ControlPages/Typography.xaml
```diff
@@ -0,0 +1,116 @@
+<!--
```
chingucoding, line 1: Would it be better to move the Design guidance pages into a separate folder?

gregwoo-microsoft: Yes I agree, this would make more sense.

WinUIGallery/DataModel/ControlInfoData.json
```diff
@@ -21,7 +21,7 @@
           "Content": "",
           "IsNew": false,
           "IsUpdated": false,
-          "HideSourceCodeLinks": true,
```
niels9001: https://github.com/microsoft/WinUI-Gallery/pull/1184 already addresses this by hiding the Related Controls and Docs buttons when their arrays are empty - so then we wouldn't need this bool?

chingucoding: Well we are also hiding the source code links with this option since they don't make much sense for these sections. I don't think there is an option to do this without a special flag on the control page.

niels9001: Yes true, but I wonder if we could do the same.

Thinking about it.. are there particular pages where hiding the source code links? Seems to be pretty useful always right?

niels9001:
Awesome work on the colors section @chingucoding :)! Really cool!

One concern I do have is the stretching of the ColorTiles. Ideally, the tiles would reflow underneath each other on resize (like what a GridView would do), right?

image

WinUIGallery/Controls/DesignGuidance/ColorTile.xaml
```diff
@@ -0,0 +1,44 @@
...
+            Text="{x:Bind ColorName, Mode=OneWay}" Style="{ThemeResource BodyStrongTextBlockStyle}"/>
+        <Button Grid.Column="1" AutomationProperties.Name="Copy brush name"
+                Background="Transparent" BorderBrush="Transparent"
+                Foreground="{x:Bind Foreground, Mode=OneWay}"
+                ToolTipService.ToolTip="Copy brush name" VerticalAlignment="Top"
+                HorizontalAlignment="Right" Click="CopyBrushNameButton_Click" Padding="4">
+            <FontIcon Glyph="&#xE8C8;" FontSize="16"/>
+        </Button>
+
+        <TextBlock Text="{x:Bind ColorExplanation, Mode=OneWay}" Grid.Row="1"
```
niels9001, line 37: This needs to CaptionTextBlockStyle as per design spec.. will save some space as well :)!

WinUIGallery/Controls/DesignGuidance/ColorTile.xaml
```diff
@@ -0,0 +1,44 @@
...
+            TextWrapping="WrapWholeWords"/>
+
+        <TextBlock Text="{x:Bind ColorBrushName, Mode=OneWay}" Grid.Row="3" Grid.ColumnSpan="2"
```
niels9001, line 40: CaptionTextBlockStyle

WinUIGallery/Controls/DesignGuidance/ColorTile.xaml
```diff
@@ -0,0 +1,44 @@
...
+            TextWrapping="Wrap" IsTextSelectionEnabled="True"/>
+        <TextBlock Text="{x:Bind ColorValue, Mode=OneWay}" Grid.Row="4" Grid.ColumnSpan="2" TextWrapping="WrapWholeWords"/>
```
niels9001, line 42: CaptionTextBlockStyle

chingucoding:
I was afraid that someone would ask about that. With the different number of elements in the samples, I don't think there is an easy way for that. A control like GridView or ItemsRepeater will always create rows and fill them but I don't think would make the last elements stretch the rest of the row like we currently have with the hardcoded layouts. I'm really open for suggestions here since the text spanning multiple lines is not ideal.

gregwoo-microsoft:
I think that it's fine to keep it like this for V1 and eventually look to improve it in the future. Thoughts?

niels9001:
> I think that it's fine to keep it like this for V1 and eventually look to improve it in the future. Thoughts?

Yep agree.. @chingucoding Since the WinUI brush name is the longest, I could imagine we could maybe hide that on small screensizes (under an i button with a tooltip)? And maybe the Toolkit's UniformGrid could be helpful as well.

The challenging bit is the many different layouts that we want to support.. but yeah, I agree it's fine for V1.. once we get it in I'm happy to try some stuff out :)

niels9001:
@gregwoo-microsoft @chingucoding The HEX code would be different for Light vs. Dark theme.. We could do something fancy by swapping out the hex codes based on the selected theme - but maybe better to leave this out completely for V1?

Especially since it's information not super relevant for devs - we want them to use the WinUI brush and NOT hardcode colors using HEX values. And if we link to the Figma file at the top of the page (that has the extensive overview of all colors, with HEX values included) they could always look it up if needed? Thoughts?

image

chingucoding:
You are raising a very good point there @nielslaute, especially the fact that developers should use the brushes and not have to resort to the colors. In my opinion, I think we can get rid of them. If people want the colors, they can still get that information from Figma.

WinUIGallery/DataModel/ControlInfoData.json
chingucoding: I don't think there are special pages beside the design guidance where it would make a lot of sense. Thinking about it, the a11y section might actually also benefit from that. The reason for those links is that users often wandered where the source code is for the samples and the sample code below the samples often cannot capture the entire context.

chingucoding:
@gregwoo-microsoft @niels9001 I implemented the changes we discussed:
- Separator between colors
- Remove sample code
- Hide color hex codes
- Show copy button on hove
- Add small notice about using brushes (needs improvement)

What is currently missing is the sticky header and adding missing brushes. I'm not sure how easy the sticky header will be since we are inside an ItemsPage which also provides scrolling of our content. Regarding missing brushes, what do you both think about adding a small "i" button that users can click to know why the brush is red?

niels9001:
> @gregwoo-microsoft @niels9001 I implemented the changes we discussed:
> - Separator between colors
> - Remove sample code
> - Hide color hex codes
> - Show copy button on hove
> - Add small notice about using brushes (needs improvement)
> 
> What is currently missing is the sticky header and adding missing brushes. I'm not sure how easy the sticky header will be since we are inside an ItemsPage which also provides scrolling of our content. Regarding missing brushes, what do you both think about adding a small "i" button that users can click to know why the brush is red?
> 

Yes, makes sense. I'd leave the i and sticky header stuff for now and tackle those in seperate PRs once this gets merged maybe?

chingucoding:
> Yes, makes sense. I'd leave the i and sticky header stuff for now and tackle those in seperate PRs once this gets merged maybe?

Sure, works for me. After all, don't want to drag out PRs waiting on this even longer.

niels9001 (approved):

PR #1172, opened 2023-01-22
Replace ProjectReunion version variables with WindowsAppSdk ones

bpulliam:
## Description
This change retires the ProjectReunionPackageVersion and ProjectReunionWinUIPackageVersion variables that date back to the early days of the project.  The variable WindowsAppSdkPackageVersion is more correct and intuitive.

## Motivation and Context
One less special case to remember


chingucoding (commented):
Shouldn't we also rename the package/project references for those variables? Or at least remove them?

Scottj1s (approved):

chingucoding (approved):

PR #1174, opened 2023-01-22
Accessibility fixes from internal repo

bpulliam:
## Description
Fixes the following:

Changes Theme on the settings page to use radiobuttons
Removes Automation Name from System Backdrops field so Narrator will use Content
Restores LabeledBy in actual Xaml that was still present in sample text
Adds Level to NewControls Groups


chingucoding (approved):
LGTM

PR #1184, opened 2023-01-30
[UX] UX improvements and fixes

niels9001:
@gregwoo-microsoft @chingucoding This turned out to be a larger PR than I anticipated due to the many things that were somehow linked, and a lot of legacy code.

Major changes:
- New design and cleanup of the ItemsPage.
- A ton of design improvements, so all pages now renderly nicely on every viewpoint.
- UI controls will now be hidden, e.g.: whenever there are no 'related controls' the related controls button wouldn't show.
- Images now bleed from edge-to-edge, based on new design specs.
- Removing PageHeader usercontrol (because it was only used on the ItemsPage itself).

Tour:

https://user-images.githubusercontent.com/9866362/215463444-8aa42f77-108a-4617-a94e-0fd8ada15119.mp4


gregwoo-microsoft:
Thank you so much for implementing this Niels. This is great. I listed my comments below:

* The new top buttons are great, they make it easier to access documentation links. Should we remove the related controls button? I don't think that it brings a ton of value (since related controls are already visible in the navView panel on the left). Removing it would leave more space for other buttons (like a Figma button).
* The style icon doesn't appear like a button on my end, I don't see borders around it. Is that normal?
image

  * It would nice to increase the margin below the buttons so that there is a bit of space between the buttons and the content when scrolling
image

* The settings page needs to have more space at the top (this might be linked to the other settings page issues)
image

* Same thing for the search results page, we should increase the top margin (maybe not as much as the current WinUI 3 Gallery but just a bit to have some extra space)
image

* Teaching tip sample: I think that we should change the sentences used in that sample so that it doesn't refer to the app's theme anymore (since it's no longer being targeted at the Theme button) --> for instance change the name of the sample to just "Show a targeted TeachingTip with hero content"
image

Let me know what you think!

niels9001:
@gregwoo-microsoft Thanks, addressed the feedback!

Except for the Settingspage.. I'm leaving that for now since we'll use the new SettingsControls and that requires a rewrite of the page anyways (that'll be my next PR :))

WinUIGallery/AllControlsPage.xaml
```diff
@@ -14,100 +14,71 @@
...
+            Loaded="OnItemGridViewLoaded"
+            SelectionMode="None" />
         <VisualStateManager.VisualStateGroups>
             <!--  Visual states reflect the application's window size  -->
             <VisualStateGroup x:Name="LayoutVisualStates">
                 <VisualState x:Name="WideLayout">
                     <VisualState.StateTriggers>
                         <AdaptiveTrigger MinWindowWidth="{StaticResource Breakpoint640Plus}" />
                     </VisualState.StateTriggers>
-                    <!-- XAML_BEHAVIORS_REMOVED
```
chingucoding, line 31: I think XAML Behaviors supports WinUI now so maybe we want to keep that?

WinUIGallery/AllControlsPage.xaml
```diff
@@ -14,100 +14,71 @@
...
+            IsItemClickEnabled="True"
+            IsSwipeEnabled="False"
+            ItemClick="OnItemGridViewItemClick"
+            ItemContainerStyle="{StaticResource IndentedGridViewItemStyle}"
+            ItemTemplate="{StaticResource ControlItemTemplate}"
+            ItemsSource="{x:Bind Items, Mode=OneWay}"
+            KeyDown="OnItemGridViewKeyDown"
+            Loaded="OnItemGridViewLoaded"
+            SelectionMode="None" />
         <VisualStateManager.VisualStateGroups>
```
chingucoding, line 61: I would prefer having VisualState stuff before the actual content of the control/page. This is also how its done everywhere in the framework.

WinUIGallery/App.xaml
```diff
@@ -24,115 +24,141 @@
...
...12789917,5.34375619888306 17.0145378112793,5.71729135513306 18.596851348877,6.62109994888306 21.799976348877,6.19062519073486 25.004674911499,6.62265014648438 26.5845413208008,5.71818733215332 27.6791000366211,5.34472513198853 28.315746307373,5.28210020065308 28.5218753814697,5.31015014648438 28.8556652069092,6.53784370422363 28.8976573944092,7.5438346862793 28.6499996185303,8.69764995574951 29.6154251098633,10.2268533706665 29.9656257629395,12.1328001022339 29.453296661377,15.1497011184692 28.1123065948486,17.0164012908936 26.2366523742676,18.020601272583 24.120325088501,18.4500007629395 24.7275562286377,19.3355484008789 24.9890747070313,20.8187503814697 24.9804744720459,23.0584030151367 24.9718742370605,24.3312511444092 25.1693305969238,24.8128852844238 25.8531246185303,24.9453010559082 29.3641395568848,23.1273632049561 32.1326217651367,20.3568344116211 33.948070526123,16.8442134857178 34.5999984741211,12.8000001907349 34.3399276733398,10.2204961776733 33.5940399169922,7.81787109375 32.4138298034668,5.64361572265625 30.8507804870605,3.74921894073486 28.9563827514648,2.18616962432861 26.7821273803711,1.00595712661743 24.3795032501221,0.26007080078125 21.7999992370605,0z</x:String>
```
chingucoding, line 147: Is that really the best way to do that? Wouldn't having a GitHub Icon Asset in the assets folder be easier maintenance wise?

WinUIGallery/ControlPages/AppBarPage.xaml
```diff
@@ -1,4 +1,4 @@
-﻿<!--
+<!--
```
chingucoding, line 1: I don't think users can actually visit this page anymore.

WinUIGallery/ControlPages/DesignGuidance/IconsPage.xaml
```diff
@@ -92,235 +100,134 @@
...
-                <ColumnDefinition Width="*" MaxWidth="800"/>
+                <ColumnDefinition Width="*" />
+                <ColumnDefinition Width="296" />
             </Grid.ColumnDefinitions>
-            <Grid.RowDefinitions>
-                <RowDefinition Height="Auto" />
-                <RowDefinition Height="Auto" />
-            </Grid.RowDefinitions>
+            <ScrollViewer
+                Margin="-4,0,12,0"
```
chingucoding, line 138: Please don't use negative margins.

WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -0,0 +1,165 @@
...
+                        </Flyout>
+                    </DropDownButton.Flyout>
+                </DropDownButton>
+                <DropDownButton AutomationProperties.Name="Source code" ToolTipService.ToolTip="Source code">
+                    <DropDownButton.Content>
+                        <StackPanel Orientation="Horizontal" Spacing="8">
+                            <PathIcon
+                                Margin="-3,-3,-16,-3"
+                                VerticalAlignment="Center"
+                                Data="{StaticResource GitHubIconPath}">
```
chingucoding, line 82: Is there really no better way to do this?

WinUIGallery/ItemTemplates.xaml
```diff
@@ -50,33 +31,54 @@
...
+                    Margin="4"
                     VerticalAlignment="Top"
-                    CornerRadius="{StaticResource ControlCornerRadius}"
-                    Visibility="{x:Bind BadgeString, Converter={StaticResource nullToVisibilityConverter}}" >
+                    Style="{StaticResource AttentionIconInfoBadgeStyle}"
+                    Visibility="{x:Bind BadgeString, Converter={StaticResource nullToVisibilityConverter}}" />
 
-                    <Ellipse Fill="{ThemeResource SystemAccentColorDark1}" Width="12" Height="12"/>
-                </Grid>
+                <VisualStateManager.VisualStateGroups>
```
chingucoding, line 63: Same here, please put VSM stuff before the actual content.

WinUIGallery/ItemsPageBase.cs
```diff
@@ -76,14 +76,14 @@ protected void OnItemGridViewKeyDown(object sender, KeyRoutedEventArgs e)
...
-                    FocusManager.TryMoveFocus(FocusNavigationDirection.Up);
-                }
+                //if (nextElement?.GetType() == typeof(Microsoft.UI.Xaml.Controls.NavigationViewItem))
+                //{
+                //    NavigationRootPage.GetForElement(this).PageHeader.Focus(FocusState.Programmatic);
+                //}
+                //else
+                //{
+                //    FocusManager.TryMoveFocus(FocusNavigationDirection.Up);
+                //}
```
chingucoding, line 86: Intentional?

WinUIGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -16,139 +16,220 @@
...
+                <NavigationViewItemHeader Content="Controls" />
+                <NavigationViewItem
+                    x:Name="AllControlsItem"
+                    Content="All"
+                    Icon="AllApps" />
                 <NavigationViewItemSeparator />
             </NavigationView.MenuItems>
+
         </NavigationView>
+        <VisualStateManager.VisualStateGroups>
```
chingucoding, line 216: Same here.

WinUIGallery/NewControlsPage.xaml
```diff
@@ -75,79 +38,76 @@
...
-            Subtitle="{StaticResource WinUIVersion}"
-            Margin="44,92"
-            VerticalAlignment="Top"
-            Foreground="Black"
-            Background="{ThemeResource HomePageBackgroundBrush}"
-            FontSize="28"
-            Padding="{ThemeResource NavigationViewHeaderMargin}"
-            IsHitTestVisible="False"
-            ShadowOpacity="0"/>
+        <VisualStateManager.VisualStateGroups>
```
chingucoding, line 92: Same here.

WinUIGallery/SearchResultsPage.xaml.cs
```diff
@@ -116,8 +114,8 @@ private void BuildFilterList(string queryText)
                 {
                     // Display informational text when there are no search results.
                     VisualStateManager.GoToState(this, "NoResultsFound", false);
-                    var textbox = NavigationRootPage.GetForElement(this)?.PageHeader?.GetDescendantsOfType<AutoSuggestBox>().FirstOrDefault();
-                    textbox?.Focus(FocusState.Programmatic);
+                    //var textbox = NavigationRootPage.GetForElement(this)?.PageHeader?.GetDescendantsOfType<AutoSuggestBox>().FirstOrDefault();
+                    //textbox?.Focus(FocusState.Programmatic);
```
chingucoding, line 118: Intentional?

WinUIGallery/WinUIGallery.DesktopWap.csproj
```diff
@@ -28,6 +28,7 @@
        in the WinUI repo, or the next ItemGroup below when standalone. -->
   <ItemGroup>
     <PackageReference Include="ColorCode.Core" />
+    <PackageReference Include="CommunityToolkit.WinUI.UI" Version="7.1.2" />
```
chingucoding, line 31: If we are currently only using this for converters, it's worth checking if we can remove converters we implemented ourselVes (though I would probably just copied the converters from the toolkit or build them ourselves).

chingucoding (changes requested):
Opening any page seems to crash now:

```
Exception thrown: 'System.InvalidCastException' in System.Private.CoreLib.dll
Exception thrown at 0x00007FF8982C06BC (KernelBase.dll) in WinUIGallery.DesktopWap.exe: WinRT originate error - 0x80004002 : 'Unable to cast object of type 'Microsoft.UI.Xaml.Controls.NavigationViewItemSeparator' to type 'Microsoft.UI.Xaml.Controls.NavigationViewItem'.'.
onecore\com\combase\winrt\error\restrictederror.cpp(1017)\combase.dll!00007FF89A7CB92A: (caller: 00007FF89A7D0E9F) ReturnHr(274) tid(3fc0) 8007007E The specified module could not be found.

```
Further thoughts:
- Clicking on the "All controls" section seems to move selection to the design guidance section.
- Not sure if we need a separator inside the design guidance section, I would probably remove it
- Mayber reworking the individual samples is better done in a separate PR, it's already exhausting to review as is. Refactoring individual samples makes this even harder.
- Adding the "Controls" NVI Header is a very good idea
- When using touchpad, you can scroll the home page into the following temporary situation (because scrollview allows scrolling outside) 
Screenshot of weird homepage

WinUIGallery/Controls/PageHeader.xaml
niels9001: This was something I had laying around for the Toolkit Gallery :).. 

SVG is probably best? We do want to set the Foreground so it renders correctly in light/dark theme

WinUIGallery/ControlPages/DesignGuidance/IconsPage.xaml
niels9001: Have to.. to visually align the ItemsRepeater items with the rest of the content

WinUIGallery/ControlPages/AppBarPage.xaml
niels9001: Removed!

WinUIGallery/AllControlsPage.xaml
niels9001: Ah yes.. XAML Styler messed up. Fixed!

WinUIGallery/AllControlsPage.xaml
niels9001: Yeah, but no longer using pageheader here.. I'd remove it for now and rethink if (based on design review) and how to do this (we could use a Toolkit helper to do this nicely)?

niels9001:
> Opening any page seems to crash now:
> 
> ```
> Exception thrown: 'System.InvalidCastException' in System.Private.CoreLib.dll
> Exception thrown at 0x00007FF8982C06BC (KernelBase.dll) in WinUIGallery.DesktopWap.exe: WinRT originate error - 0x80004002 : 'Unable to cast object of type 'Microsoft.UI.Xaml.Controls.NavigationViewItemSeparator' to type 'Microsoft.UI.Xaml.Controls.NavigationViewItem'.'.
> onecore\com\combase\winrt\error\restrictederror.cpp(1017)\combase.dll!00007FF89A7CB92A: (caller: 00007FF89A7D0E9F) ReturnHr(274) tid(3fc0) 8007007E The specified module could not be found.
> ```

Uh oh.. Do you have repro steps? Seems to work fine on my machine..

> Further thoughts:
> 
> * Clicking on the "All controls" section seems to move selection to the design guidance section.

I think this is due to #1167 -  do you see why this is happening?

> * Not sure if we need a separator inside the design guidance section, I would probably remove it

Removed

> * Mayber reworking the individual samples is better done in a separate PR, it's already exhausting to review as is. Refactoring individual samples makes this even harder.

Yeah you're right.. this thing is becoming too big. I'll leave things in as they are now.

WinUIGallery/ItemsPageBase.cs
niels9001: The PageHeader was not in use.. any idea what the intended behavior was here?

niels9001: Oh nvm.. was thinking about the landings page.

niels9001: Oh no.. wait.. this was only used on the landings page and controls overviews where the PageHeader was no longer in use. So this can be removed all together?

chingucoding:
> Uh oh.. Do you have repro steps? Seems to work fine on my machine..

This only seems to occur with debugger attached. In my case, I was searching for "AppBarButton" and clicked on the respective search result in the dropdown. Next thing is debugger halting because of unhandled exception.

> I think this is due to https://github.com/microsoft/WinUI-Gallery/pull/1167 - do you see why this is happening?

Hmm, not sure yet. I can take a look at this though if you like since I was the one to introduce this behavior.

WinUIGallery/SearchResultsPage.xaml.cs
niels9001: image

Seems to be dead code.. for whenever this was still a thing.

niels9001:
> > Uh oh.. Do you have repro steps? Seems to work fine on my machine..
> 
> This only seems to occur with debugger attached. In my case, I was searching for "AppBarButton" and clicked on the respective search result in the dropdown. Next thing is debugger halting because of unhandled exception.

Are you seeing the same on Main?

chingucoding:
> Are you seeing the same on Main?

I am not able to reproduce it on main and funnily, also not anymore on your branch so I guess it's fixed now.

>
Uh oh.. Do you have repro steps? Seems to work fine on my machine..


> > Hmm, not sure yet. I can take a look at this though if you like since I was the one to introduce this behavior.
> Yes please!

I found the issue, it's [here](https://github.com/microsoft/WinUI-Gallery/blob/2bf63f0697d3684ded6e66d4e48dfe68c3e8ba36/WinUIGallery/AllControlsPage.xaml.cs#L30-L31) not sure why these lines are needed or if we can remove them, however since this PR moves the AllControl page from index 1 to index 3, it now selects the design guidance page when navigating to the all controls page.

niels9001:
@chingucoding Hmm, might have been a faulty commit.. that was a bit all over the place.

I have made all the changes for now.. feel free to do a final pass and approve. Feels after this PR we are in a place where we can target smaller bugs/fixes in dedicated PRs :).

WinUIGallery/Controls/DesignGuidance/ColorSections/SignalSection.xaml
```diff
@@ -0,0 +1,170 @@
...
+                <ColumnDefinition Width="*" />
+            </Grid.ColumnDefinitions>
+            <designguidance:ColorTile
+                Background="{ThemeResource SystemFillColorAttentionBrush}"
+                ColorBrushName="SystemFillColorAttentionBrush"
+                ColorExplanation="Badge"
+                ColorName="System / Attention"
+                ColorValue="#60CDFF (100%)"
+                Foreground="{ThemeResource TextFillColorInverseBrush}"
+                ShowSeparator="False" />
```
chingucoding, line 40: Is there a specific reason to stick with the order provided in the Figma File compared to arranging them in a way that the Success, Caution and Critical files are next to each other and their respective background colors beneath them? I originally did that in #1167 since it looks more pleasing in my opinion.

WinUIGallery/ControlPages/DesignGuidance/IconsPage.xaml
```diff
@@ -92,235 +100,134 @@
         </DataTemplate>
     </Page.Resources>
 
-    <StackPanel Spacing="10" HorizontalAlignment="Stretch">
+    <Grid HorizontalAlignment="Stretch">
```
chingucoding, line 103: The spacing at the bottom looks quite odd now:

Screenshot of IconsPage

WinUIGallery/ControlPages/DesignGuidance/TypographyPage.xaml
```diff
@@ -14,151 +14,218 @@
...
+    xmlns:controls="using:WinUIGallery.DesktopWap.Controls"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:local="using:AppUIBasics"
-    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:controls="using:WinUIGallery.DesktopWap.Controls"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     x:Name="compactPage"
     mc:Ignorable="d">
 
-    <StackPanel Orientation="Vertical">
+    <Grid>
```
chingucoding, line 24: Spacing between header line and actual content seems a bit tight:

Screenshot of Typographypage

WinUIGallery/ItemPage.xaml
```diff
@@ -13,37 +13,25 @@
     x:Class="AppUIBasics.ItemPage"
```
chingucoding, line 13: I think the spacing at the bottom (and maybe right) is a bit tight now:

Screenshot of StandardUICommand page

WinUIGallery/ControlPages/DesignGuidance/AccessibilityScreenReaderPage.xaml
```diff
@@ -28,50 +25,47 @@
...
-                <LineBreak /> Screen readers, such as
-                <Hyperlink NavigateUri="https://support.microsoft.com/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1">Narrator</Hyperlink> , convert text into spoken words to help blind or low vision users.
-                Screen readers use the UI Automation (UIA) names of each control to report their name, role and content.
-                <LineBreak />
-                <LineBreak /> If the control's content is a string, an accessible name is automatically determined from the visible
-                text. However, elements such as images or input fields need to have a custom accessible name. Here
-                are a few examples of how to add UIA names:
-            </Paragraph>
-        </RichTextBlock>
+    <ScrollViewer>
```
chingucoding, line 30: Scrollbar now overlaps content: 
Screenshot of Accessibility page

chingucoding: Same applies for all other a11y pages.

WinUIGallery/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
chingucoding, line 1: Is this file actually needed?

chingucoding (commented):
What do you think about new icons for the categories? Right now it seems a bit out of place having these monocolor icons next to the colorful ones:

Screenshot of navigation

Also, do we not want users to toggle the sample theme on the design guidance pages anymore? Or was that broken, in which case we can leave it as is.

WinUIGallery/Controls/PageHeader.xaml
chingucoding: I think we can do the fill stuff with an SVG too. I also still got https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/4494/files open which touches on the topic of it not being properly centered in the WCT Gallery.

WinUIGallery/ItemsPageBase.cs
chingucoding: Yes I think we can remove that.

WinUIGallery/AllControlsPage.xaml
chingucoding: Maybe this got reverted somehow?

gregwoo-microsoft:
Thanks a lot @niels9001 for working on this!

> What do you think about new icons for the categories? Right now it seems a bit out of place having these monocolor icons next to the colorful ones:
> 
> Screenshot of navigation
> 

Yeah we had looked into adding those icons in the past but the design team had advised against it for that reason. Having a mix of colored + non-colored icons feels inconsistent. 

Other comments:
- Can we slightly increase the window's launch width? Right now when you open the app, you only see 1 column of controls which looks odd. 
image

- I don't think that we should have sticky headers for the typography & colors page --> other control pages don't have this behavior so it seems weird to have it there + the header of the colors page is too big so it takes most of the window's height. I'm in favor of removing the sticky headers for now.
- Can we please revert the changes made to the Icons page? I can't scroll on the page and the layout is no longer consistent with the other pages.

niels9001:
> Thanks a lot @niels9001 for working on this!
> 
> > What do you think about new icons for the categories? Right now it seems a bit out of place having these monocolor icons next to the colorful ones:
> > Screenshot of navigation
> 
> Yeah we had looked into adding those icons in the past but the design team had advised against it for that reason. Having a mix of colored + non-colored icons feels inconsistent.

Yeah this was more of a trial haha - especially since the old icons are still using the W8/W10 style. Will revert!


> Other comments:
> 
> * Can we slightly increase the window's launch width? Right now when you open the app, you only see 1 column of controls which looks odd.
> 
> image

Yeah this happened because the design spec wanted the cards to have the same width as 2 tiles + spacing.. which makes sense visually. Changing the window size is tricky and might bring us more, as WASDK apps due this based on screen ratio

> * I don't think that we should have sticky headers for the typography & colors page --> other control pages don't have this behavior so it seems weird to have it there + the header of the colors page is too big so it takes most of the window's height. I'm in favor of removing the sticky headers for now.

Ooooh.. I figured the sticky header was intended in the design mockup 😅. My bad! The thought was that you could always switch between brushes vs. scrolling all the way up and then switch. If we decide not to go with the sticky headers for the colors page we can remove some other stuff here as well for the design guidance stuff. I'll revert!

> * Can we please revert the changes made to the Icons page? I can't scroll on the page and the layout is no longer consistent with the other pages.

The previous version only shows 9 or so icons :(.. this version takes advantage of all screen real estate but see your point. I'll have a look what we can optimize still to have something in between :)!

niels9001:
@gregwoo-microsoft @chingucoding We're using ItemsRepeater for the icon tool - is there a specific reason why we do this vs. GridView? ItemsRepeater doesn't support keyboard navigation out of the box where GridView does.. not sure if it's a huge priority though. Let me know!

WinUIGallery/Controls/DesignGuidance/ColorSections/SignalSection.xaml
niels9001: image

Yeah, there is a connection between those colors (one is the icon and other is the background of the same style so it makes sense to keep them together?). Moved the attention below similar to what you had so the green/orange/red is visually more pleasing :)

niels9001:
@gregwoo-microsoft @chingucoding Addressed feedback, reverted some changes made, added missing Elevation brushes and warning tooltip:

image

chingucoding:
Regarding ItemsRepeater, we opted to use that since styling GridViewItems can be quite difficult while ItemsRepeater is very unopinionated in that regard.

Further feedback, the links seem to be stuck to text on A11y pages
Screen reader support page with faulty text

gregwoo-microsoft:
Thanks a lot Niels for implementing those changes.

I've tried compiling the last few bits we need to merge this PR. Let me know what you think @niels9001 and @chingucoding.
- Add new home banner (I know Niels managed to make it work but listing it for completeness)
	
- Move the 'external link' icon to the bottom right of the card instead of the top right (the design team gave this feedback. They think that it is best to have the logo and title be at the top of the hierarchy)
	
- Remove the black and white control icons from the navView (since the icons are in color, the design team had suggested only having icons for the categories (i.e. basic input, collections, etc.) but no icons for the individual controls --> basically how it is in the current version of the Gallery). I think it's fine to leave the icons for the design guidance pages since we don't have color icons for those pages. Any thoughts?
	
- [Stretch] Decrease the size (both width and length) of the home tiles? They seem really big right now but let me know what you think is best.

- Move the "Controls" header in the navView below the separator? or remove the separator under the "All controls" page. (with the separator, it looks at first glance like the "Basic input" section is not part of the "Controls" section. Any thoughts?)

- The hyperlinks on the typography and accessibility pages are missing a space at the beginning

- Love what you did with the Icons page, you can now see more icons at a time. The only thing I would add for consistency with the other samples is the grey outline around both the icons library and the details pane on the right.

- On the stroke page, the first example (control elevation) has an extra/empty space on the right. If there are only 3 brushes, can we remove this empty space and use the 3 tile design (like for the second row of the control stroke example)?

- The toggle light/dark mode doesn't work on the colors page. Let me know if this should be a separate PR.

- [Stretch] Thanks for removing the sticky headers. I agree that ideally we would want to make the segmented control sticky but not sure if we can do it since there text over it. Let me know if this is possible (you can scroll down the page past the intro text and then only the segmented control sticks to the top of the screen) and if we should do it in another PR.

gregwoo-microsoft (approved):
Thanks a lot Niels for making those changes!

gregwoo-microsoft:
@chingucoding I assume that you're no longer experiencing any crashes but didn't want to dismiss your change request just in case. Feel free to merge this PR if you're good.

gregwoo-microsoft:
Including here a quick list of the remaining fixes we can address in separate PRs:
- Add new home banner
- Move the 'external link' icon to the bottom right of the card instead of top right
- Make home link tiles a bit smaller?
- Add github + template studio icon assets
- Remove "all controls" page and the separator below? (@chingucoding let me know your thoughts on this, Niels and I were chatting and we feel like this page doesn't add any value)
- Toggle light/dark mode doesn't work on the Colors page

chingucoding (approved):

chingucoding:
A few months ago, I would have probably said "sure let's remove the all controls page" but recently I often used it to see if there is any control that could possibly suite a use case. I'll leave that decision to you though I would argue that it doesn't cost us much to maintain does it?

gregwoo-microsoft:
Ok sounds good, let's not remove it then. The reason why I brought it up is because I felt like it was weird visually to have the "Controls" header in the navView, then the "All controls" page, then a separator, and then all of the controls. Having the separator feels like the controls underneath are not part of the "Controls" header. A solution to that was to remove the "All controls" page along with the separator. Perhaps the best solution is to simply remove the separator or move it above the new "Controls" header.

Scottj1s:
@niels9001 looks like a [bad merge](https://github.com/microsoft/WinUI-Gallery/blob/7f4d33a1b6a9b725902d6da245f0e8d6163ceafa/WinUIGallery/Navigation/NavigationRootPage.xaml.cs#L118). can you take a look?

niels9001:
> @niels9001 looks like a [bad merge](https://github.com/microsoft/WinUI-Gallery/blob/7f4d33a1b6a9b725902d6da245f0e8d6163ceafa/WinUIGallery/Navigation/NavigationRootPage.xaml.cs#L118). can you take a look?  I would just open a PR, but want to make sure nothing else is amiss.

https://github.com/microsoft/WinUI-Gallery/pull/1193 should fix that :)

PR #1192, opened 2023-02-10
[UX] New landingspage experience

niels9001:
New landingspage UX

https://user-images.githubusercontent.com/9866362/218108274-f845fa71-24e9-4c76-b49d-79ba7a0d826d.mp4


WinUIGallery/Controls/HomePageHeaderImage.xaml.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        private Compositor _compositor;
+        private CompositionLinearGradientBrush _imageGridBottomGradientBrush;
+        private CompositionEffectBrush _imageGridEffectBrush;
+        private ExpressionAnimation _imageGridSizeAnimation;
+        private ExpressionAnimation _bottomGradientStartPointAnimation;
+        private SpriteVisual _imageGridSpriteVisual;
+        private CompositionSurfaceBrush _imageGridSurfaceBrush;
+        private Visual _imageGridVisual;
+        private CompositionVisualSurface _imageGridVisualSurface;
```
chingucoding, line 32: This seems quite a lot, especially the fact that some of those are animations objects. Is that really the best way to solve the problem of fading out an image?

WinUIGallery/Controls/TileGallery.xaml
```diff
@@ -164,36 +178,31 @@
...
+                                FontSize="44"
                                 Foreground="{ThemeResource TextFillColorPrimaryBrush}"
                                 Glyph="&#xE943;" />
                         </local:HeaderTile.Source>
                     </local:HeaderTile>
                     <local:HeaderTile
                         Title="Partner Center"
                         Description="Upload your app to the Store."
-                        Link="https://docs.microsoft.com/windows/apps/get-started/samples">
+                        Link=" https://partner.microsoft.com/en-us/dashboard/home">
```
chingucoding, line 203: Maybe it would be better to point to the landing page or similar of the partner center? I don't think that it's a good experience if the first thing people are required to do when opening this link is signing in to their MSFT account. 

Also, trailing space.

WinUIGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -14,7 +14,6 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:data="using:AppUIBasics.Data"
-    xmlns:local="using:AppUIBasics"
```
chingucoding, line 17: Good catch!

WinUIGallery/WinUIGallery.DesktopWap.csproj
```diff
@@ -29,6 +29,7 @@
   <ItemGroup>
     <PackageReference Include="ColorCode.Core" />
     <PackageReference Include="CommunityToolkit.WinUI.UI" Version="7.1.2" />
+    <PackageReference Include="CommunityToolkit.WinUI.UI.Animations" Version="7.1.2" />
```
chingucoding, line 32: Is this because of the home screen fade out?

chingucoding (changes requested):

WinUIGallery/Controls/HomePageHeaderImage.xaml.cs
niels9001: Ah, good catch - I forgot to push the zoom in animation on app load.

For the gradient: this is actually trickier than expected and I couldn't find a lot online.. having the transparency-fade in the image wouldn't work on resizing the window, using a LinearGradientBrush wouldn't work either when Mica is enabled. This allows us to put in any image and have a nice fade effect based on Composition. But obviously open to other code suggestions :)

WinUIGallery/Controls/TileGallery.xaml
niels9001: Hmm good point.. reverted this back to: https://developer.microsoft.com/windows/ which has a "Dashboard" link at the top.

WinUIGallery/WinUIGallery.DesktopWap.csproj
niels9001: Yeah, and the onload animation mentioned in the reply above.

WinUIGallery/Controls/HomePageHeaderImage.xaml.cs
chingucoding: It seems it's really not that trivial. I would probably not have implemented that animation due to the complexity of it but I don't have a strong opinion on that.

chingucoding (approved):

PR #1212, opened 2023-03-10
Switch to new theme changing

chingucoding:
This PR fixes the design guidance sections not fully supporting theme changes. The icons page just changes the samples theme, for the colors page the most easiest way to implement the theme change currently would be to switch the entire page's theme.


niels9001 (approved):
LGTM!

PR #1236, opened 2023-03-29
Removing unused assets

niels9001:
This PR removes a lot of unused assets.


WinUIGallery/CompactSizePage.xaml
```diff
@@ -1,15 +0,0 @@
-﻿<Page
-    x:Class="AppUIBasics.CompactSizePage"
```
chingucoding, line 2: I assume compact sizing is not something WinUI 3 is encouraging to use?

chingucoding (approved):

WinUIGallery/CompactSizePage.xaml
Jay-o-Way: Or because it's also under `WinUIGallery\ControlPages` ?

Jay-o-Way:
Spring cleanup 🧹🧽🪣😋

PR #1239, opened 2023-03-30
Fix for bug when user changes Windows System theme when winui gallery is up and running, caption buttons don't pick the correct theme color. 

pratikone:
When you open winui gallery and then goto Windows System setting and change its theme, winui gallery dynamically updates to new theme but not caption buttons. 

This happens because we are using WinUI 3 custom titlebar for titlebar and caption controls. The caption buttons need to be colored by our code. In order for this code to dynamically update the caption button foreground color, it needs to know when the system setttings have changed. Windows.UI.ViewManagement.Settings provide that solution. Though for reasons unknown to me, its lambda is called off-thread so a dispatcher queue is needed to call caption controls update function in app's UI thread. 

(I referenced the implementation Windows Community Toolkit's [ThemeListener ](https://learn.microsoft.com/en-us/windows/communitytoolkit/helpers/themelistener)helper class)

Line endings have also been updated for some files to CRLF which Windows uses.

## Types of changes

### DIFF
--- a/WinUIGallery/Helper/TitleBarHelper.cs
+++ b/WinUIGallery/Helper/TitleBarHelper.cs
@@ -21,7 +21,7 @@ namespace WinUIGallery.DesktopWap.Helper
     internal class TitleBarHelper
     {
 
-        public static void triggerTitleBarRepaint(Window window)
+        private static void triggerTitleBarRepaint(Window window)
         {
             // to trigger repaint tracking task id 38044406
             var hwnd = WinRT.Interop.WindowNative.GetWindowHandle(window);


chingucoding (changes requested):
Thanks for submitting a PR for this! There are a few things I noticed:
1. We should reuse the existing mechanism we already have, namely update the `TitleBarHelper.ApplySystemThemeToCaptionButtons();` to properly update the color of the buttons. This function is already being triggered on theme change by a colors changed listener in the NavigationRootPage.
2. I was able to get dark caption buttons in dark theme, is that intentional?

pratikone:
Thanks for the suggestion. Yeah,I could definitely refactor more so that Titlebarhelper functions are used more effectively and functionality is not repeated. However, a lot of this code will go away soon once I provide appwindow + winui 3 custom titlebar merger, hence, i am not spending more effort in refactoring this code.

For 2. Can you provide more details ?  did you visit Titlebar page ? It sets the foreground color to something custom. The default custom color is black. I could set it that default is different for light and dark colors but that's for another PR 🙂

chingucoding:
Regarding 1, I'm strongly against pulling in a dependency for a functionality/pattern we already have. If you plan on refactoring AppWindow and custom titlebar anyway, I think this can also wait.

Regarding 2, I opened the app on the settings page. I switched the app theme to dark and then changed the system theme to light.

pratikone:
@chingucoding  thanks for letting me know of that bug issue. It should be fixed now. I am using `FrameworkElement.ActualTheme` instead of `Application.RequestedTheme`. 
For the first part, the merger code will take months to come. Given winui gallery is a demo app for WinUI functionality, I would prefer it to have correct functionality, even if it comes with cost of pulling a new dependency.

WinUIGallery/Helper/TitleBarHelper.cs
```diff
@@ -1,43 +1,63 @@
...
+using Microsoft.UI.Xaml.Navigation;
+using WinRT;
+using System.Runtime.InteropServices;
+
+namespace WinUIGallery.DesktopWap.Helper
+{
+    internal class TitleBarHelper
+    {
+
+        public static void triggerTitleBarRepaint(Window window)
```
bpulliam, line 24: I would make this private.  The other methods should give all required functionality and this is more of an implementation detail.

bpulliam (approved):
approved with one suggestion.

chingucoding (approved):
LGTM

PR #1240, opened 2023-03-30
tweak fontsizes

Jay-o-Way:
## Description

Just small changes

## Motivation and Context

- Do not to use font-size 14 on icons, they are designed with 16px in mind.
  - only change properties when needed, but at the very minimum use the "rule-of-4"
- Quote from the app: "The minimum values should be 12px Regular, 14px Semibold"

WinUI3Gallery icons

## Types of changes


karenbtlai:
Hey Jay-o-Way! Thanks for your contribution. Do you mind adding more detailed before and after changes photos in your PR description?

Jay-o-Way:
Hi @karenbtlai I would love to. But I have issues actually running/debugging both the projects.
WinUIGallery.DesktopWap gives me:
image

WinUIGallery.DesktopWap.Package gives me:
image

Jay-o-Way:
@karenbtlai done. Visually, it's absolutely minute. More a matter of principle and design/code habits.

WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -48,7 +48,7 @@
                     Visibility="{x:Bind Item.Docs, Converter={StaticResource collectionConverter}}">
                     <DropDownButton.Content>
                         <StackPanel Orientation="Horizontal" Spacing="8">
-                            <FontIcon FontSize="14" Glyph="&#xE8A5;" />
+                            <FontIcon Glyph="&#xE8A5;" />
```
niels9001, line 51:
```suggestion
                            <FontIcon FontSize="16" Glyph="&#xE8A5;" />
```

WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -116,7 +116,7 @@
                     ToolTipService.ToolTip="Toggle theme"
                     Visibility="{x:Bind ThemeButtonVisibility, Mode=OneWay}">
                     <Button.Content>
-                        <FontIcon FontSize="14" Glyph="&#xE793;" />
+                        <FontIcon Glyph="&#xE793;" />
```
niels9001, line 119:
```suggestion
                        <FontIcon FontSize="16" Glyph="&#xE793;" />
```

WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -155,7 +154,7 @@
                     Click="OnFeedBackButtonClick"
                     ToolTipService.ToolTip="Send feedback">
                     <Button.Content>
-                        <FontIcon FontSize="14" Glyph="&#xED15;" />
+                        <FontIcon Glyph="&#xED15;" />
```
niels9001, line 157:
```suggestion
                        <FontIcon FontSize="16" Glyph="&#xED15;" />
```

niels9001 (changes requested):

niels9001:
@jay-o-way @karenbtlai 

Synced with the design team, and by default we should adhere to using 16/24/32/48 as @Jay-o-Way pointed out.

However, FontIcon/SymbolIcon uses 20 by default (I always was under the impression it would default to 16, like normal text..). So we still have to manually set these to "16".

Chart for comparison (as @Jay-o-Way pointed out, 14 causes blurriness on higher resolutions screens):
image

This is also inline with icons in e.g. the Start Menu or systemtray:
  
image

Jay-o-Way:
> FontIcon/SymbolIcon uses 20 by default

Could that be, because of a style somewhere, perhaps?

niels9001:
> > FontIcon/SymbolIcon uses 20 by default
> 
> Could that be, because of a style somewhere, perhaps?

Yeah, I'm guessing this might be from the pre-WinUI era where in general icons and fonts were bigger.

WinUIGallery/ControlPages/ListViewPage.xaml
```diff
@@ -429,7 +429,7 @@ sent or received, and those values are bound in the DataTemplate.--&gt;
             &lt;StackPanel Orientation="Horizontal"&gt;
                 &lt;TextBlock Text="{x:Bind Views}" HorizontalAlignment="Left" Style="{ThemeResource CaptionTextBlockStyle}" Margin="0,0,0,0"/&gt;
                 &lt;TextBlock Text=" Views " HorizontalAlignment="Left" Style="{ThemeResource CaptionTextBlockStyle}"/&gt;
-                &lt;TextBlock Text=" &#x22C5; " FontFamily="Segoe UI" FontSize="12" FontWeight="Bold" 
+                &lt;TextBlock Text=" &#x22C5; " FontFamily="Segoe UI" Style="{ThemeResource CaptionTextBlockStyle}"
```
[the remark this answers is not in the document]
Jay-o-Way: Done

Jay-o-Way:
So, pr looks good now? :)

niels9001:
@jay-o-way

Icons seem to be vertically off center now..
image

WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -48,7 +48,7 @@
                     Visibility="{x:Bind Item.Docs, Converter={StaticResource collectionConverter}}">
                     <DropDownButton.Content>
                         <StackPanel Orientation="Horizontal" Spacing="8">
-                            <FontIcon FontSize="14" Glyph="&#xE8A5;" />
+                            <FontIcon FontSize="16" Glyph="&#xE8A5;" />
                             <TextBlock Text="Documentation" />
```
niels9001, line 52: You might want to add a bottom margin (`Margin="0,0,0,2") to the `TextBlock` so it's centered correctly.

WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -129,7 +129,6 @@
                     <Button.Content>
                         <FontIcon
                             x:Name="CopyLinkButtonIcon"
```
niels9001, line 131: This would need `FontSize="16"` as well?

niels9001 (changes requested):

WinUIGallery/Controls/PageHeader.xaml
Jay-o-Way: > looks like it's the link icon

@niels9001 You've got sharp eyes!
I took the image from https://learn.microsoft.com/nl-nl/windows/apps/design/style/segoe-fluent-icons-font#:~:text=Stop-,e71b,-Link
and see that this is by design. If Microsoft designed it like that, I'm nog touching it :)
image

WinUIGallery/Controls/PageHeader.xaml
Jay-o-Way: "need" seems to be a difficult word here. If it defaults to 16, I don't think it **needs** to. Same discussion for the FontFamily.

WinUIGallery/Controls/PageHeader.xaml
niels9001: It still looks out of place now. So we'd still need to fix the offset.

Jay-o-Way: > this is by design. I'm not changing it :)

WinUIGallery/Controls/PageHeader.xaml
niels9001: As mentioned above, `FontIcon` defaults to 20. That's what's happening now:

image

When `FontSize` is set to 16 it renders the correct size (with the offset as mentioned earlier):
image

So we'd need to explicitly set it to make sure all icons render the same size.

WinUIGallery/Controls/PageHeader.xaml
Jay-o-Way: > IIRC Fluent Icons use a 1 pixel thickness for the lines

Try to think of it as **4px** thickness in a **16x16** icon. Then it all adds up. How else would an icon like image be made?

niels9001: Right, what I meant was that these icons were probably designed with a 1px grid. Then you end up with 3 pixels above or below in some edge cases, so it's not perfectly aligned.

image

I think with a simple margin tweak we can visually fix that in context with the other icons?

Jay-o-Way: I respectfully refuse. If you insist, you're free to do so in a pr of your own :)

WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -48,7 +48,7 @@
                     Visibility="{x:Bind Item.Docs, Converter={StaticResource collectionConverter}}">
                     <DropDownButton.Content>
                         <StackPanel Orientation="Horizontal" Spacing="8">
-                            <FontIcon FontSize="14" Glyph="&#xE8A5;" />
```
niels9001, line 51: image

This is still a visible regression over the current version as the text label is not vertically aligned wrt the icon and glyph.

FYI, it's very common to add a specific margin to center align text within a control. (e.g. see the default button padding that has a different top/bottom margin to center align the label)

New vs. old:
image

niels9001 (changes requested):

WinUIGallery/Controls/PageHeader.xaml
Jay-o-Way: Then I would still prefer `VerticalAlignment="Center"`

Jay-o-Way: You're comparing the baseline of the icon content with the baseline of text. They don't need to be the same thing.

Jay-o-Way:
@niels9001 can we wrap this up?

niels9001 (approved):
Thanks

PR #1243, opened 2023-04-03
With retarget of app to 19041, wapproj also needs update to prevent nuget errors

Scottj1s:
With the UI refresh, the app was retargeted to 19041, but the wapproj was not.  This was causing nuget restore errors (though the build would still succeed), and the potential for runtime versioning issues.  The wapproj needs to specify a TargetPlatformMinVersion >= the app's TargetFrameworkVersion.


chingucoding (approved):
LGTM

Issue #1252, opened 2023-04-05
XAML code still declares Font Family "MDL2"


Jay-o-Way:
### Which version of the app?

WinUI 3 Gallery

### Description

`FontFamily="Segoe MDL2 Assets"` shouldn't be used anymore

### Screenshots

![image](https://user-images.githubusercontent.com/65828559/230216337-68ea733e-08bb-4e8d-ba83-debb70f79233.png)

Jay-o-Way:
Seems simple. Should I wait for a "green light" or may I?

chingucoding:
I think you can go ahead on this. I assume you would replace it with `SymbolThemeFontFamily` then?

Jay-o-Way:
Thank you. It'll be replaced by either \<nothing> or Symboltheme, indeed.

Jay-o-Way:
I have a problem in order to actually test this. See, the computer I build/run this on is on Windows 10. Because of this issue https://github.com/microsoft/WinUI-Gallery/discussions/1246 I can't actually run any of the exe's and therefore I can't copy that folder to my Windows 11 laptop... Can I solve this? Or shall I just create the pr and let people like you or Niels check this on their own device?

chingucoding:
Hmm that's strange. I'm curious, is there a reason you want to test this on a Window 11 device? After all we also should support Windows 10 as WinUI 3 does. Regarding #1246, I'll take a look what the correct way would be to get an exe file working.

Jay-o-Way:
> is there a reason you want to test this on a Window 11 device?

Of course, Windows 10 doesn't have the Fluent font (well, unless you installed it manually). The icons _show_ without a problem, just wanna double-check the font.

Jay-o-Way:
P.S. I found a way to use Fluent Icons on Windows 10. Just install the font, that's it! 🤣

PR #1260, opened 2023-04-10
Remove MDL2 from code

Jay-o-Way:
## Description

Let's not use MDL2 :)

## Motivation and Context

Closes: #1252

## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/IconElementPage.xaml
+++ b/WinUIGallery/ControlPages/IconElementPage.xaml
@@ -32,7 +32,7 @@
                 <StackPanel>
                     <TextBlock Text="Use FontIcon as the icon for a control if you want to specify a Glyph value from a FontFamily. Windows 10 uses the Segoe MDL2 Assets FontFamily and that is what this example is showing." Style="{ThemeResource BodyTextBlockStyle}" Margin="0,0,0,12"  />
                     <Button Name="ExampleButton1">
-                        <FontIcon Glyph="&#xE790;"/>
+                        <FontIcon FontFamily="Segoe MDL2 Assets" Glyph="&#xE790;"/>
                     </Button>
                 </StackPanel>
             </local:ControlExample.Example>
--- a/WinUIGallery/ControlPages/IconElementPage.xaml
+++ b/WinUIGallery/ControlPages/IconElementPage.xaml
@@ -32,7 +32,7 @@
                 <StackPanel>
                     <TextBlock Text="Use FontIcon as the icon for a control if you want to specify a Glyph value from a FontFamily. Windows 10 uses the Segoe MDL2 Assets FontFamily and that is what this example is showing." Style="{ThemeResource BodyTextBlockStyle}" Margin="0,0,0,12"  />
                     <Button Name="ExampleButton1">
-                        <FontIcon FontFamily="Segoe MDL2 Assets" Glyph="&#xE790;"/>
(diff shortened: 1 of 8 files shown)


WinUIGallery/ControlPages/IconElementPage.xaml
```diff
@@ -32,7 +32,7 @@
                 <StackPanel>
                     <TextBlock Text="Use FontIcon as the icon for a control if you want to specify a Glyph value from a FontFamily. Windows 10 uses the Segoe MDL2 Assets FontFamily and that is what this example is showing." Style="{ThemeResource BodyTextBlockStyle}" Margin="0,0,0,12"  />
```
chingucoding, line 33: I think we need to update this info here. Maybe just remove it? @niels9001 @karenbtlai thoughts?

niels9001: Maybe update this with the explaination about Segoe MDL2 Assets being the default on W10 and Segoe Fluent Icons on W11?

chingucoding: Sounds good to me. @jay-o-way do you want to include this in your PR?

Jay-o-Way: Hm, actually, the header **explicitly** says "...a glyph from a specific font family..." so in that context I should be reverting this specific example.

tashatitova:
Not sure if we can just remove it like that. We use SymbolThemeFontFamily that redirects to appropriate font for controls that use icons

tashatitova (changes requested):
We can't just remove it like that. We use SymbolThemeFontFamily that redirects to appropriate font for controls that use icons. Otherwise it will fall back to Xaml Auto Font Family which is a text font. Also, all these instances would need to be tested, for example, AnimatedIcon

Jay-o-Way:
@tashatitova that's highly peculiar because this very app shows examples just like this. \<Fonticon glyph="">

Jay-o-Way:
Can you show how it looks on your device?

Jay-o-Way:
Actually, all that hard-coding is bad practice. The variable(s) exist(s) for a reason and, as you say, Fonticon defaults to that, i see no reason to hardcode it. Only adding ore possilities for failures.

Jay-o-Way:
@tashatitova 

> FontIcon defaults to SymbolThemeFontFamily

This speaks for itself.

> isn't something recommended, at least as far as I've seen

Have a look at WinUI 3 Gallery, it's a good source:

```xaml
<DropDownButton AutomationProperties.Name="Email">
    <DropDownButton.Content>
        <FontIcon Glyph="&#xE715;"/>
    </DropDownButton.Content>
    <DropDownButton.Flyout>
        <MenuFlyout Placement="Bottom">
            <MenuFlyoutItem Text="Send">
                <MenuFlyoutItem.Icon>
                    <FontIcon Glyph="&#xE725;"/>
                </MenuFlyoutItem.Icon>
            </MenuFlyoutItem>
            <MenuFlyoutItem Text="Reply">
                <MenuFlyoutItem.Icon>
                    <FontIcon Glyph="&#xE8CA;"/>
                </MenuFlyoutItem.Icon>
```

Jay-o-Way:
@niels9001 I'm ready

niels9001 (approved):
Thanks, changes look good.

PR #1276, opened 2023-04-20
Using JsonSerializer instead of JsonObject

ghost1372:
Using System.Text.Json.JsonSerializer instead of JsonObject for deserializing ControlInfoData.json file.
By using JsonSerializer, the codes become simpler and less, and we can use Async/Await
https://github.com/microsoft/WinUI-Gallery/discussions/1270

## Description

- Using JsonSerializer
- Remove unnecessary items from ControlInfoData.json
- Bring Back Icons to NavigationViewItem  (Why has no one noticed it?)

## How Has This Been Tested?

Tested On my Laptop

## Types of changes

### DIFF
--- a/WinUIGallery/Common/FileLoader.cs
+++ b/WinUIGallery/Common/FileLoader.cs
@@ -1,9 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using System.Reflection;
-using System.Text;
 using System.Threading.Tasks;
 using Windows.Storage;
 
@@ -11,7 +8,7 @@ namespace AppUIBasics.Common
 {
     internal class FileLoader
     {
-        public static async Task<string> LoadText(string relativeFilePath)
+        public static async Task<StorageFile> GetJsonFile(string relativeFilePath)
         {
 #if UNPACKAGED
             var sourcePath = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), relativeFilePath));
@@ -20,6 +17,11 @@ namespace AppUIBasics.Common
             Uri sourceUri = new Uri("ms-appx:///" + relativeFilePath);
             var file = await StorageFile.GetFileFromApplicationUriAsync(sourceUri);
 #endif
+            return file;
+        }
+        public static async Task<string> LoadText(string relativeFilePath)
+        {
+            var file = await GetJsonFile(relativeFilePath);
 
             return await FileIO.ReadTextAsync(file);
         }
--- a/WinUIGallery/Common/FileLoader.cs
+++ b/WinUIGallery/Common/FileLoader.cs
@@ -8,7 +8,7 @@ namespace AppUIBasics.Common
 {
     internal class FileLoader
     {
-        public static async Task<StorageFile> GetJsonFile(string relativeFilePath)
+        public static async Task<string> LoadText(string relativeFilePath)
         {
 #if UNPACKAGED
             var sourcePath = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), relativeFilePath));
@@ -17,12 +17,6 @@ namespace AppUIBasics.Common
             Uri sourceUri = new Uri("ms-appx:///" + relativeFilePath);
             var file = await StorageFile.GetFileFromApplicationUriAsync(sourceUri);
 #endif
-            return file;
-        }
-        public static async Task<string> LoadText(string relativeFilePath)
-        {
-            var file = await GetJsonFile(relativeFilePath);
-
             return await FileIO.ReadTextAsync(file);
         }
 
--- a/WinUIGallery/DataModel/ControlInfoDataSource.cs
+++ b/WinUIGallery/DataModel/ControlInfoDataSource.cs
@@ -12,8 +12,9 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading.Tasks;
-using Windows.Data.Json;
 using AppUIBasics.Common;
+using System.IO;
+using System.Text.Json;
 
 // The data model defined by this file serves as a representative example of a strongly-typed
 // model.  The property names chosen coincide with data bindings in the standard item templates.
@@ -25,12 +26,16 @@ using AppUIBasics.Common;
 
 namespace AppUIBasics.Data
 {
+    public class Root
+    {
+        public ObservableCollection<ControlInfoDataGroup> Groups { get; set; }
+    }
     /// <summary>
     /// Generic item data model.
     /// </summary>
     public class ControlInfoDataItem
     {
-        public ControlInfoDataItem(string uniqueId, string title, string apiNamespace, string subtitle, string imagePath, string imageIconPath, string badgeString, string description, string content, bool isNew, bool isUpdated, bool isPreview, bool hideSourceCodeAndRelatedControls)
+        public ControlInfoDataItem(string uniqueId, string title, string apiNamespace, string subtitle, string imagePath, string imageIconPath, string badgeString, string description, string content, bool isNew, bool isUpdated, bool isPreview, bool hideSourceCodeAndRelatedControls, ObservableCollection<ControlInfoDocLink> docs, ObservableCollection<string> relatedControls)
         {
             this.UniqueId = uniqueId;
             this.Title = title;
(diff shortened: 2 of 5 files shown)


WinUIGallery/Navigation/NavigationRootPage.xaml.cs
```diff
@@ -201,7 +196,7 @@ private void AddNavigationMenuItems()
 
                 foreach (var item in group.Items)
                 {
-                    var itemInGroup = new Microsoft.UI.Xaml.Controls.NavigationViewItem() { IsEnabled = item.IncludedInBuild, Content = item.Title, Tag = item.UniqueId, DataContext = item };
```
chingucoding, line 204: This change was intentional, see #1006. @niels9001, @karkarl thoughts?

WinUIGallery/DataModel/ControlInfoDataSource.cs
```diff
@@ -194,105 +197,70 @@ public async Task<ControlInfoDataGroup> GetGroupFromItemAsync(string uniqueId)
...
+            {
+                WriteIndented = true,
+                PropertyNameCaseInsensitive = true
+            });
+            string pageRoot = "AppUIBasics.ControlPages.";
+            foreach (var groupItem in controlInfoDataGroup.Groups)
             {
-                string pageRoot = "AppUIBasics.ControlPages.";
-                foreach (JsonValue groupValue in jsonArray)
+                ControlInfoDataGroup group = new ControlInfoDataGroup(groupItem.UniqueId,
```
chingucoding, line 215: Why do we still need to this, doesn JSONSerializer support parsing nested trees?

WinUIGallery/DataModel/ControlInfoDataSource.cs
```diff
@@ -45,26 +50,26 @@ public ControlInfoDataItem(string uniqueId, string title, string apiNamespace, s
...
-        public string ImageIconPath { get; private set; }
-        public string BadgeString { get; private set; }
-        public string Content { get; private set; }
-        public bool IsNew { get; private set; }
-        public bool IsUpdated { get; private set; }
-        public bool IsPreview { get; private set; }
-        public bool HideSourceCodeAndRelatedControls { get; private set; }
-        public ObservableCollection<ControlInfoDocLink> Docs { get; private set; }
-        public ObservableCollection<string> RelatedControls { get; private set; }
+        public string UniqueId { get; set; }
```
chingucoding, line 58: Is the setter public so the JSONSerializer can set them?

WinUIGallery/Common/FileLoader.cs
```diff
@@ -1,17 +1,14 @@
...
-using System.Text;
 using System.Threading.Tasks;
 using Windows.Storage;
 
 namespace AppUIBasics.Common
 {
     internal class FileLoader
     {
-        public static async Task<string> LoadText(string relativeFilePath)
+        public static async Task<StorageFile> GetJsonFile(string relativeFilePath)
```
chingucoding, line 11: Why this split in loading the file and reading it as string? Shouldn't it be fine to just load the json as string?

WinUIGallery/DataModel/ControlInfoDataSource.cs
```diff
@@ -194,105 +197,70 @@ public async Task<ControlInfoDataGroup> GetGroupFromItemAsync(string uniqueId)
...
-
-            JsonObject jsonObject = JsonObject.Parse(jsonText);
-            JsonArray jsonArray = jsonObject["Groups"].GetArray();
-
-            lock (_lock)
+            var jsonFile = await FileLoader.GetJsonFile("DataModel/ControlInfoData.json");
+            using FileStream openStream = File.OpenRead(jsonFile.Path);
+            var controlInfoDataGroup = await JsonSerializer.DeserializeAsync<Root>(openStream, new JsonSerializerOptions
+            {
+                WriteIndented = true,
```
chingucoding, line 209: Intentional?

WinUIGallery/DataModel/ControlInfoDataSource.cs
```diff
@@ -194,105 +197,70 @@ public async Task<ControlInfoDataGroup> GetGroupFromItemAsync(string uniqueId)
...
 
-            string jsonText = await FileLoader.LoadText("DataModel/ControlInfoData.json");
-
-            JsonObject jsonObject = JsonObject.Parse(jsonText);
-            JsonArray jsonArray = jsonObject["Groups"].GetArray();
-
-            lock (_lock)
+            var jsonFile = await FileLoader.GetJsonFile("DataModel/ControlInfoData.json");
+            using FileStream openStream = File.OpenRead(jsonFile.Path);
+            var controlInfoDataGroup = await JsonSerializer.DeserializeAsync<Root>(openStream, new JsonSerializerOptions
```
chingucoding, line 207: Why use a stream and not just the text?

chingucoding (changes requested):

WinUIGallery/Navigation/NavigationRootPage.xaml.cs
ghost1372: oh, I didn't know, should I return it?

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: I tried not to touch the original code, I will check

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: To use Async we need to use Stream. Text is not part of the input parameters in `DeserializeAsync` method, If there is no need for Async, we can use Text in `Deserialize` method.
so should we keep async or sync?

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/how-to?pivots=dotnet-7-0

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: for using async we need to use stream

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: JsonSerializer will be able to set its value during deserialization. In `private` mode, the number of items in group is 0 (group.Items = 0)

WinUIGallery/DataModel/ControlInfoDataSource.cs
chingucoding: If you need a reference for this, you can look at the IconsDataSource which just lets JSONSerializer handle everything.

WinUIGallery/DataModel/ControlInfoDataSource.cs
chingucoding: I assume protected or internal doesn't work either right?

WinUIGallery/DataModel/ControlInfoDataSource.cs
chingucoding: There really is no need to make the parsing be async. The app cannot start when we don't have this information so we need to wait for this. And the fact that both DeserializeAsync and the function calling DeserializeAsync are both awaited indicates we can take use of the synchronous function.

WinUIGallery/Navigation/NavigationRootPage.xaml.cs
chingucoding: Let's wait for Niels or Karens answer on this but I would probably say yes.

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: yes not working

WinUIGallery/Common/FileLoader.cs
ghost1372: you are using `LoadText` inside `LoadLines` and `LoadLines` inside `ListViewPage` so I didn't want to make too many changes.
If we don't use async, all these things will return to the previous state

WinUIGallery/DataModel/IconsDataSource.cs
```diff
@@ -27,26 +27,24 @@ internal class IconsDataSource
...
         {
-            lock (_lock)
+            if (icons.Count != 0)
             {
-                if (icons.Count != 0)
-                {
-                    return icons;
-                }
+                return icons;
             }
```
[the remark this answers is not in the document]
ghost1372: done

WinUIGallery/Common/FileLoader.cs
ghost1372: done, i changed async to sync and we dont need to split methods

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: done, i changed async to sync and we are using Text now.

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: Did it get better?

WinUIGallery/DataModel/ControlInfoDataSource.cs
chingucoding: But do we need `WriteIndented` ?

chingucoding (approved):

WinUIGallery/DataModel/ControlInfoDataSource.cs
ghost1372: No we dont need for deserializing, Done.

chingucoding (approved):

WinUIGallery/Navigation/NavigationRootPage.xaml.cs
karkarl: Yes, the icons in the submenu level were removed since they are not fluent icons, and the design team does not have capacity to create fluent icons for each of the controls. We will have to revert this change unfortunately.

karkarl (changes requested):
Will have to revert the changes for the sublevel NavView icons.

ghost1372:
> Will have to revert the changes for the sublevel NavView icons.

done

karkarl (approved):
Looks great, thank you!

PR #1282, opened 2023-05-08
Display Windows App SDK and Runtime version info

Scottj1s:
Uses auto-init bootstrapper version info contained in WindowsAppSDK-VersionInfo.cs to generate version strings for Home and Settings pages.  This also demonstrates how app developers can use version info for other purposes.


bpulliam (approved):
While you're here, we can remove line 155 from Settingspage.xaml: <HyperlinkButton Content="Xaml Behaviors" NavigateUri="https://github.com/Microsoft/XamlBehaviors" />.  We don't use XamlBehaviors anywhere.

PR #1289, opened 2023-05-15
Add initial CI pipeline with UI tests and axe scanning

### LINKED ISSUE #1024: Create CI pipeline
### Issue type

enhancement

### Which version of the app?

WinUI 3 Gallery

### Description

Use this pipeline to do test automation, performance testing, and accessibility testing. 

Start with simple navigation test in this pipeline.

### Screenshots

_No response_

### Windows version

_No response_

### Additional context

_No response_

chingucoding:
This PR adds the initial CI pipeline with azure devops, re-enable the UI tests and add Axe Accessibility scanning.

Closes #1024


kmahone:
We should avoid adding a .pfx key to the repo. This can cause issues, since they expire and also get flagged by scanning tools as being 'secrets' even in cases such as this where it is just for testing.

In WinUI, we use a script to generate the .pfx key at build time  [GenerateTestPfx.ps1](https://github.com/microsoft/microsoft-ui-xaml/blob/main/GenerateTestPfx.ps1)
A similar approach might work well here.

WinUIGallery/Package.WAP.Dev.appxmanifest
```diff
@@ -7,7 +7,7 @@
     xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
     IgnorableNamespaces="uap mp uap3">
   <!--Note: package version and assembly version must be kept in sync-->
-  <Identity Name="Microsoft.WinUI3ControlsGallery.Debug" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="2.1.0.0" />
+  <Identity Name="Microsoft.WinUI3ControlsGallery.Debug" Publisher="CN=Microsoft Corporation" Version="2.1.0.0" />
```
kmahone, line 10: >Publisher="CN=Microsoft Corporation" [](http://example.com/codeflow?start=57&length=36)

If this is being used for testing purposes, we could update the Publisher to something like "CN=WinUI-Gallery-Test" so it is clear that it is not an official build of something from Microsoft.

UITests/TestBase.cs
```diff
@@ -0,0 +1,50 @@
...
+
+		public static void OpenControlPage(string name)
+		{
+			var search = Session.FindElementByName("Search");
+			search.Clear();
+			Thread.Sleep(1_000);
+			search.SendKeys(name);
+			Thread.Sleep(1_000);
+			Session.FindElementByName(name).Click();
+			Thread.Sleep(5_000);
```
kmahone, line 34: Can we do something better than thread sleeps here? This seems unreliable.

E.g. can we wait for a certain piece of UI to update instead?

kmahone:
> global using OpenQA.Selenium;

I don't see the advantage of doing this. It makes more sense to include the using directives in the files that require them.

---
Refers to: UITests/GlobalUsings.cs:1 in 07c17dd. [](commit_id = 07c17dd72cc19bf9d05796e75f5d9e6a14e767b9, deletion_comment = False)

UITests/TestBase.cs
```diff
@@ -0,0 +1,50 @@
...
+			AxeHelper.AssertNoAccessibilityErrors();
```
kmahone, line 35: We should add explicit Axe test methods, instead of having this validation piggy-back on other tests via this helper method.

kmahone: E.g. Tests\Button.cs should be updated to have
```
[TestMethod]
public void ValidateAccessibilityWithAxe()
```
or something like that.

UITests/TestBase.cs
```diff
@@ -0,0 +1,50 @@
...
+
+		public static void TypeText(string text)
+		{
+			var actions = new Actions(Session);
+			actions.SendKeys(text).Perform();
+		}
+
+		[ClassCleanup]
+		public static void Cleanup()
+		{
```
kmahone, line 46: What is the purpose of this Cleanup step?

kmahone (changes requested):
:clock1:

UITests/TestBase.cs
chingucoding: I'm not really sure how you would do that with selenium, if you have any pointers I would love to do that.

chingucoding:
I think I addressed most of your feedback @kmahone, the global usings file is gone. accessibility testing is an explicit test now. Regarding waiting for an element, I would appreciate some pointers on how to set this up.

kmahone:
> I think I addressed most of your feedback @kmahone, the global usings file is gone. accessibility testing is an explicit test now. Regarding waiting for an element, I would appreciate some pointers on how to set this up.

Thanks. I've updated the comment with a suggested approach for this.

UITests/TestBase.cs
chingucoding: I also thought about that API, but its quite sophisticated. IIRC, part of that infra was also moved into a package right? Or was that something else.

Regarding the polling approach, I'm not sure if there is something sensible we want to wait for. The idea was to wait for the app to navigate to the page but I don't think we can "poll" for this without introducing UIA elements.

kmahone: In the case of a page navigate you could wait until a known element on the page is available in the UIA tree.

UITests/TestBase.cs
```diff
@@ -0,0 +1,73 @@
...
+			for (int i = 0; i < 100; i++)
+			{
+				var header = Session.FindElementsByName(name).Where(x => x.GetProperty("AutomationId") == "PageHeader");
+				if (header != null)
+				{
+					break;
+				}
+				Thread.Sleep(50);
+			}
```
kmahone, line 64: >} [](http://example.com/codeflow?start=3&length=1)

We should throw in the case where we never find the element.

kmahone (approved):

kmahone:
Looks good to me. @karkarl should take a look and approve also before merge.

UITests/TestBase.cs
chingucoding: Gread idea, fixed that.

chingucoding:
After/before merging, of course some Azure DevOps project also needs to be configured which is not something that I can do.

kmahone:
> After/before merging, of course some Azure DevOps project also needs to be configured which is not something that I can do.

Yes. Karen is going to set that up.

karkarl (approved):

Issue #1265, opened 2023-04-12
[Color contrast tool] TextBoxes should be wider


niels9001:
### Which version of the app?

WinUI 3 Gallery

### Description

When the color code textboxes have focus, the clear button appears and the text is cut off.. which is annoying as changing the cursor position will make the text jump around.

Increasing the width to support a hex code + clear button should resolve that.

<img width="358" alt="image" src="https://user-images.githubusercontent.com/9866362/231422332-0342807b-c256-4ab9-9ee6-30ac132cc71a.png">

PR #1294, opened 2023-05-17
[A11y] Increasing MinWidth of color textbox

niels9001:
This PR increases the MinWidth of the color textbox so that when selected the user can read the entire HEX value without the clear button getting in the way. Addressing: https://github.com/microsoft/WinUI-Gallery/issues/1265

Before:
image

After:
image

### DIFF
--- a/WinUIGallery/ControlPages/DesignGuidance/AccessibilityColorContrastPage.xaml
+++ b/WinUIGallery/ControlPages/DesignGuidance/AccessibilityColorContrastPage.xaml
@@ -38,7 +38,7 @@
         <RichTextBlock>
             <Paragraph>
                 Accessibility is about building experiences that make your Windows application usable by people of
-                all abilities. For more information about designing accessible apps: <Hyperlink NavigateUri="https://learn.microsoft.com/windows/apps/design/accessibility/accessibility-overview">Accessibility overview</Hyperlink>
+                all abilities. For more information about designing accessible apps:<Hyperlink NavigateUri="https://learn.microsoft.com/windows/apps/design/accessibility/accessibility-overview">Accessibility overview</Hyperlink>
                 .<LineBreak />
                 <LineBreak />
                 To ensure optimal accessibility and usability, apps should strive to use high-contrast and easy-to-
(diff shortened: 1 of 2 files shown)


chingucoding (changes requested):
The textboxes now are cut off and you cannot click the "clear text" button anymore. I think we also need to increase the width specified on the color contrast page for the control to fix this.

niels9001:
> The textboxes now are cut off and you cannot click the "clear text" button anymore. I think we also need to increase the width specified on the color contrast page for the control to fix this.

Whoops, missed that. Thanks! Fixed now:

image

WinUIGallery/ControlPages/DesignGuidance/AccessibilityColorContrastPage.xaml
```diff
@@ -38,15 +38,15 @@
...
+                all abilities. For more information about designing accessible apps:<Hyperlink NavigateUri="https://learn.microsoft.com/windows/apps/design/accessibility/accessibility-overview">Accessibility overview</Hyperlink>
                 .<LineBreak />
                 <LineBreak />
                 To ensure optimal accessibility and usability, apps should strive to use high-contrast and easy-to-
                 read color combinationto for text and its background. Not only will this benefit users with lower
                 visual acuity, but this will also ensure visibility and legibility under a wide range of lighting
                 conditions, screens, and device settings.<LineBreak />
                 <LineBreak />
-                Check out the <Hyperlink NavigateUri="https://accessibilityinsights.io/">Accessibility Insights</Hyperlink>
+                Check out the<Hyperlink NavigateUri="https://accessibilityinsights.io/">Accessibility Insights</Hyperlink>
```
chingucoding, line 49: This makes the links stick to the word beforehand, we should probably revert that: 
Screenshot of text with links not being properly spaced to the text before it

chingucoding (changes requested):

WinUIGallery/ControlPages/DesignGuidance/AccessibilityColorContrastPage.xaml
niels9001: Ugh, XAML styler removing the spaces. Fixed now.

chingucoding (approved):

Issue #1247, opened 2023-04-04
Typograpy page: accessibility for Copy button


Jay-o-Way:
### Which version of the app?

WinUI 3 Gallery

### Description

On the Typography page, there are hidden <kbd>Copy</kbd> buttons to the right of the style names. Only on hover, they will appear. Problem is that some of them disappear as soon as you move way from the style name. Must be because of the left margin, added to a (totally unnecessary) Grid element. At the minimum, the margin could be removed, also I suggest to align with the **Icons** page: leave it visible.

Also note that **none** of these (hidden) Copy buttons can be accessed via the **keyboard**.

Note that the first two Grids could probably be merged too.

```xaml
    <Grid
        MinHeight="68"
        PointerEntered="Grid_PointerEntered"
        PointerExited="Grid_PointerExited">
        <Grid
            Padding="{x:Bind Padding, Mode=OneWay}"
            HorizontalAlignment="Stretch"
            Background="{x:Bind Background, Mode=OneWay}"
            CornerRadius="{ThemeResource ControlCornerRadius}">
            <Grid.ColumnDefinitions>
                <ColumnDefinition Width="286" />
                <ColumnDefinition Width="156" />
                <ColumnDefinition Width="140" />
                <ColumnDefinition Width="*" />
            </Grid.ColumnDefinitions>
```

### Screenshots

![Accessibility](https://user-images.githubusercontent.com/65828559/229718815-faab8460-322d-4195-9ae3-65d505304554.gif)

### Windows version

> Editie	Windows 10 Pro
> Versie	**22H2**
> Installatiedatum	‎25-‎8-‎2022
> Build van besturingssysteem	**19045.2788**
> Ervaring	Windows Feature Experience Pack 120.2212.4190.0

Jay-o-Way:
@niels9001

PR #1295, opened 2023-05-17
Adding CopyButton

niels9001:
This PR solves the current issues:

- We had multiple buttons to copy content to with different styles.
- The teaching tip didn't auto-hide correctly after clicking the button.
- There's now an easy to use `CopyButton` that shows the same animation as Snipping Tool and handles the narrator announcement.
- Addressing: #1247

Animated button


WinUIGallery/Controls/DesignGuidance/TypographyControl.xaml
```diff
@@ -137,22 +138,13 @@
...
-                        VerticalAlignment="Stretch"
-                        AutomationProperties.Name="Copy Style to clipboard"
                         Click="CopyToClipboardButton_Click"
                         Content="&#xE8C8;"
-                        FontFamily="{ThemeResource SymbolThemeFontFamily}"
-                        Style="{StaticResource SubtleButtonStyle}"
-                        ToolTipService.ToolTip="Copy Style to clipboard"
+                        Margin="4,2,8,0"
+                        ToolTipService.ToolTip="Copy to clipboard"
                         Visibility="Collapsed" />
```
Jay-o-Way, line 147: (correct me if i'm mistaken, but) This property makes it so that it still is hidden by default. That isn't consistent across all the examples: usually it's always visible.

Jay-o-Way (commented):
HorizontalAlignment (right) and Visibility are not yet consistent across the app (at least for **Typography** and **Colors** page)

Issue #1315, opened 2023-07-07
Content Dialog backdrop not displaying over custom TitleBar


BrancheyJuhas:
### Which version of the app?

WinUI 3 Gallery

### Description

The backdrop that shows over the window does not extend into the non-system title bar when the ContentDialog is displayed. This is causing an inconsistent design and can lead to confusion and unwanted result, because the problem remains still, even if the sample code is copied by the user.

### Screenshots

![Dialog Box Bug](https://github.com/microsoft/WinUI-Gallery/assets/98485942/23839ffa-e432-441a-9c92-7579c1a45d92)

### Windows version

Windows 11 22H2 (22621)

Jay-o-Way:
External issue: https://github.com/microsoft/microsoft-ui-xaml/issues/6534

PR #1316, opened 2023-07-07
ContentDialog backdrop fix

BrancheyJuhas:
This code applies overlay to the title bar of a ContentDialogExample instance.

## Description
Using VisualTreeHelper, which allows for navigation of the visual tree in Windows UI XAML, the parent and child instances of the dialog object are accessed. The first child is a rectangle that represents the title bar of the window frame. The rectangle margin is set to zero, which is used to make the overlay fill the title bar. A callback is also registered to handle a change of margin property in the rectangle. When the margin property changes to its original value, the overlay is removed by the sender.ClearValue(dp) method. 

## Motivation and Context
Extending the backdrop into the titlebar can provide a more seamless visual experience and better integration with the overall theme of the application, not resulting in inconsistant results provided to user.  https://github.com/microsoft/WinUI-Gallery/issues/1315

## How Has This Been Tested?
The Code was thoroughly user tested in both dark and light mode. In addition, it was executed successfully with changes to titlebar settings. It should not result in any additional changes or conflicts in other existing parts of the project.

Original ContentDialog
Dialog Box Bug
Fixed ContentDialog
Dialog Box Bug Fixed

## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/ContentDialogPage.xaml.cs
+++ b/WinUIGallery/ControlPages/ContentDialogPage.xaml.cs
@@ -37,6 +37,22 @@ namespace AppUIBasics.ControlPages
             dialog.Content = new ContentDialogContent();
             dialog.RequestedTheme = (VisualTreeHelper.GetParent(sender as Button) as StackPanel).ActualTheme;
 
+            dialog.Loaded += (s, args) =>
+            {
+                // Apply a dark overlay to the title bar of the window while the dialog is open
+                var parent = VisualTreeHelper.GetParent((DependencyObject)dialog);
+                var child = VisualTreeHelper.GetChild(parent, 0);
+                var frame = (Microsoft.UI.Xaml.Shapes.Rectangle)child;
+                frame.Margin = new Thickness(0);
+                frame.RegisterPropertyChangedCallback(
+                    FrameworkElement.MarginProperty,
+                    (DependencyObject sender, DependencyProperty dp) =>
+                    {
+                        if (dp == FrameworkElement.MarginProperty)
+                            sender.ClearValue(dp);
+                    });
+            };
+
             var result = await dialog.ShowAsync();
 
             if (result == ContentDialogResult.Primary)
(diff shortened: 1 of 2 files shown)


BrancheyJuhas:
> 

@microsoft-github-policy-service agree

BrancheyJuhas:
@microsoft-github-policy-service agree

Jay-o-Way:
Can you still use the title bar? E.G. drag the window?

Jay-o-Way:
Also, if this a real bug, it's actually a WinUI issue. I've seen an issue {id coming} just this week. Might be better to follow that.

BrancheyJuhas:
> Can you still use the title bar? E.G. drag the window?

Yes, it is still possible to drag the window after this bugfix.

BrancheyJuhas:
> Also, if this a real bug, it's actually a WinUI issue. I've seen [microsoft/microsoft-ui-xaml#6534](https://github.com/microsoft/microsoft-ui-xaml/issues/6534) just this week. Might be better to follow that.

I would agree, but if there is a way to fix the bug from the frontend, while not fixed in WinUI source, developers user should be aware of the solution. After bugfix in WinUI codebase, it is of course possible to edit ContentDialog page.

WinUIGallery/ControlPages/ContentDialogPage.xaml.cs
```diff
@@ -37,6 +37,22 @@ private async void ShowDialog_Click(object sender, RoutedEventArgs e)
             dialog.Content = new ContentDialogContent();
             dialog.RequestedTheme = (VisualTreeHelper.GetParent(sender as Button) as StackPanel).ActualTheme;
 
+            dialog.Loaded += (s, args) =>
+            {
+                // Apply a dark overlay to the title bar of the window while the dialog is open
```
karkarl, line 42: Can you update the comment to include that this workaround for the custom titlebar? In the meantime, I'll open a bug internally on WinUI.

karkarl:
Closing PR as the issue is fixed for WASDK 1.4 🎉

Issue #1026, opened 2022-06-17
Support for unpackaged version


gregwoo-microsoft:
### Issue type

enhancement

### Which version of the app?

WinUI 3 Gallery

### Description

Simplify solution & make it easy to build package vs unpackaged.

pratikone:
closing it as it is now complete.

Issue #993, opened 2022-06-07
App crashing


liesauer:
![image](https://user-images.githubusercontent.com/8676741/172276006-77767827-1abf-4f06-ad2d-a8082ed95910.png)

![image](https://user-images.githubusercontent.com/8676741/172276099-92acefff-d94e-4178-a710-7b36246da9aa.png)

![image](https://user-images.githubusercontent.com/8676741/172276290-f61fafda-177e-4658-8b6e-c611af9fcc44.png)

you didn't even define `UNPACKAGED` constant, i am wondering how can the unpackaged app run?

chingucoding:
The `UNPACKEGED` constant is being set by the framework IIRC. Can you elaborate what steps you did to get the app to crash?

liesauer:
i did nothing, after temporary fixing the #983 issue by copy tons of wasdk look like dlls from other winui app, relaunch `WinUIGallery.DesktopWap` and you will facing this issue. the app tries to load the bundle resources which would never success since we are launching the unpackaged `WinUIGallery.DesktopWap`.

and there is a another issue
if i launch `WinUIGallery.DesktopWap.Package` in VS, the app launches as expected. but if launch it in explorer, the app crashes as this https://github.com/microsoft/WinUI-Gallery/issues/983#issuecomment-1151788456 descripted.

chingucoding:
Are you still able to reproduce this issue on main? 😅

pratikone:
this is now fixed with support for unpackaged app

PR #1319, opened 2023-07-17
WinUI gallery unpackaged mode working + single project config for both packaged and unpackaged + namespace change

pratikone:
EDIT : will clean it up and start another PR with base to unpkg-submodule.

HOW TO REVIEW THIS PR  : 
This PR has changed namespace from WinUIGallery.DesktopWap to just WinUIGallery resulting in large number of files changes. They are trivial 1-line changes because of copy and replace.
Actual content of PR is in .csproj file and wherever `GetCurrentPackageId` is used. Rest can be skipped.

### Unpackaged mode no longer crashes on launch 
The WinUI gallery unpacakged mode was broken and crashed on launch - no matter whether building from msbuild cmd line or VS. It was happening because at some places in the code resource was getting accessed through URI which only makes sense for packaged app and not unpackaged. To remedy this, other parts of the code had separate code under `UNPACKAGED` which would build a non-URI path for unpackaged app config. This `UNPACKAGED` constant was not defined anywhere and as a result, that alternative code was not getting accessed. Some places had missing alternative code entirely. 

~This PR solves it by correctly identifying when app is built using unpackaged app and set the `UNPACKAGED` constant. It required separate changes for both VS build and msbuild command prompt.~
This PR side steps compile time variable for a easier to maintain runtime function which uses `Appmodel/GetCurrentPackageId` system api to determine whether an app is running in packaged mode or unpackaged mode.

### Single project configuration
As a need arose for cleaning up the project files for unpackaged winui gallery to work, I used this opportunity to build upon @Scottj1s 's work in `unpkg-module` branch and get rid of WAP configuration, have single project configuration for building any flavor of WinUI gallery.
As a result, using launch configurations, you can build WinUI gallery in either packaged or unpackaged mode.
I have introduced more solution configurations to aid in that.
image

image

* If you want to build a Packaged winui gallery,
select Packaged launch configuration and choose between Debug or Release solution configs.

* If you want to build a Unpackaged winui gallery,
select Unpackaged launch configuration and choose between Debug-Unpackaged or Release-Unpackaged solution configs.

A mix and match config will result in error because that's how Visual Studio works.

### Namespace change
Since WAP project is gone, the namespace Winuigallery.DesktopWap didn't make sense. I changed it to Winuigallery only. A lot of files have this change only. 

## Motivation and Context
Motivation was frustration 😁

## How Has This Been Tested?
For both VS build and msbuild scenarios, it is tested manually for both configs packaged and unpackaged.  2x2

## Types of changes


pratikone:
solves #993  and #1026

WinUIGallery/Helper/WindowHelper.cs
```diff
@@ -65,8 +67,26 @@ static public Window GetWindowForElement(UIElement element)
...
+        {
+            get
+            {
+                int bufferSize = 0;
+                byte byteBuffer = 0;
+                // Try to grab the ID of the Package of the current process (if it exists)
+                uint lastError = NativeHelper.GetCurrentPackageId(ref bufferSize, out byteBuffer);
+                bool isPacakged = true;
+
+                if (lastError == NativeHelper.APPMODEL_ERROR_NO_PACKAGE)
```
alexlamtest, line 84: What if you are unlucky and the call to GetCurrentPackageId() failed due to transient errors, such as "out of memory" or else? Currently you default to isPackage=true in that situation, can it cause problems downstream?

WinUIGallery/Helper/ProtocolActivationClipboardHelper.cs
```diff
@@ -12,33 +12,40 @@ namespace AppUIBasics.Helper
...
+                        ApplicationData.Current.LocalSettings.Values[ShowCopyLinkTeachingTipKey] = true;
+                        valueFromSettings = true;
+                    }
+                    return (bool)valueFromSettings;
+                }
+                else
                 {
-                    ApplicationData.Current.LocalSettings.Values[ShowCopyLinkTeachingTipKey] = true;
-                    valueFromSettings = true;
+                    return _showCopyLinkTeachingTip;
```
Scottj1s, line 33: don't need field - just return true

WinUIGallery/Helper/WindowHelper.cs
```diff
@@ -65,8 +67,26 @@ static public Window GetWindowForElement(UIElement element)
...
+
+        public static bool IsAppPackaged
+        {
+            get
+            {
+                int bufferSize = 0;
+                byte byteBuffer = 0;
+                // Try to grab the ID of the Package of the current process (if it exists)
+                uint lastError = NativeHelper.GetCurrentPackageId(ref bufferSize, out byteBuffer);
+                bool isPacakged = true;
```
Scottj1s, line 82: typo

WinUIGallery/WinUIGallery.DesktopWap.Package.wapproj
```diff
@@ -44,8 +44,8 @@
     <AppxPackageSigningEnabled>True</AppxPackageSigningEnabled>
     <EntryPointProjectUniqueName>WinUIGallery.DesktopWap.csproj</EntryPointProjectUniqueName>
     <AppxBundleNameForOutput>WinUIGallery.DesktopWap</AppxBundleNameForOutput>
-    <WindowsPackageType>Desktop</WindowsPackageType>
-    <!-- it will create packaged desktop app -->
+    <WindowsPackageType>Desktop</WindowsPackageType> <!-- it will create packaged desktop app -->
+    <WindowsAppSdkSelfContained>true</WindowsAppSdkSelfContained>
```
Scottj1s, line 48: always?

WinUIGallery/WinUIGallery.DesktopWap.csproj
```diff
@@ -13,7 +13,7 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
     <EnableMsixTooling>false</EnableMsixTooling>
-    <DefineConstants Condition="'$(WindowsPackageType)' == 'None'">$(DefineConstants);UNPACKAGED</DefineConstants> <!-- will execute unpackaged app specific code -->
+    <WindowsAppSdkSelfContained>true</WindowsAppSdkSelfContained>
```
Scottj1s, line 16: always?

WinUIGallery/Helper/WindowHelper.cs
Scottj1s: Looking at the implementation of GetCurrentPackageId, the first check it makes is whether the app is packaged, short-circuiting out with APPMODEL_ERROR_NO_PACKAGE if not.  This approach appears safe.

WinUIGallery/Helper/WindowHelper.cs
```diff
@@ -65,8 +67,26 @@ static public Window GetWindowForElement(UIElement element)
...
 
         static private List<Window> _activeWindows = new List<Window>();
+
+        public static bool IsAppPackaged
+        {
+            get
+            {
+                int bufferSize = 0;
+                byte byteBuffer = 0;
+                // Try to grab the ID of the Package of the current process (if it exists)
```
Scottj1s, line 80: comment is inaccurate - grabbing the package ID is precisely what you're NOT doing here.

WinUIGallery/WinUIGallery.DesktopWap.Package.wapproj
pratikone: my bad. this got slipped in. I will remove it

WinUIGallery/WinUIGallery.DesktopWap.csproj
pratikone: accidentally committed

bpulliam (changes requested):

chingucoding:
Since the WinUI Gallery is also a resource folks use for guidance on how to implement stuff: Is using a runtime check the recommended way if you ever need to target both packaged and unpackaged at the same time and need to differentiate?

Scottj1s:
> Since the WinUI Gallery is also a resource folks use for guidance on how to implement stuff: Is using a runtime check the recommended way if you ever need to target both packaged and unpackaged at the same time and need to differentiate?

@chingucoding thanks for raising this.  Certainly, if you want to produce a library that can support both packaged/unpackaged client apps, the runtime check is essential.

PR #1325, opened 2023-08-03
Merges unpkg-submodule branch into main, make the repo submodule ready

pratikone:
This change merges changes made by @Scottj1s in unpkg-submodule branch to have WinUIGallery repo act better as a submodule.

## Motivation and Context
For better maintenance of winuigallery repo when it is being maintained and referenced by winui repo internally as well as a separate OSS project. 

## How Has This Been Tested?
Manually launching the app in packaged mode (unpackaged mode is broken which my other PR will fix)
## Types of changes


nuget.config
```diff
@@ -1,15 +1,13 @@
...
   <config>
-    <clear />
     <add key="repositoryPath" value="$\..\packages" />
   </config>
   <disabledPackageSources>
     <clear />
   </disabledPackageSources>
   <packageSources>
     <!-- For more info, see https://docs.nuget.org/consume/nuget-config-file -->
-    <clear />
```
evelynwu-msft, line 12: ><clear /> [](http://example.com/codeflow?start=4&length=9)

Is this necessary? To the best of my knowledge WinUI Gallery doesn't depend on any of our internal feeds, and when this is being used as a submodule it's too deep in the file directory to influence the WinUI build.

nuget.config
```diff
@@ -1,15 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <config>
-    <clear />
```
evelynwu-msft, line 4: > <clear /> [](http://example.com/codeflow?start=3&length=10)

Same comment as below.

nuget.config
Scottj1s: this is necessary to allow a single feed configuration (best practice) so WinUI Gallery picks up its nuget.org dependencies indirectly via the single internal WinUI feed upstream.

evelynwu-msft: According to the **NuGet and MyGet multi-feed detection** page of the site you linked, this `nuget.config` would fail the check. (There's actually no way it can pass the check whether or not the `<clear />` is present.) Can we clobber/delete this file when initing the submodule so the root WinUI `nuget.config` is used?

Scottj1s: this nuget.config is currently being built without error.  we can experiment with removing it, but I made the change to enable the build.

evelynwu-msft (approved):
:shipit:

Scottj1s:
@pratikone have we confirmed that unpkg-main still builds as a submodule?

pratikone:
> @pratikone have we confirmed that unpkg-main still builds as a submodule?

Can confirm. A submodule tracking this branch's commit builds well as a submodule with internal mux repo

Scottj1s (approved):

PR #1329, opened 2023-08-12
WinUI gallery unpackaged mode working + single project config for both packaged and unpackaged

pratikone:
Unpackaged mode no longer crashes on launch
The WinUI gallery unpacakged mode was broken and crashed on launch - no matter whether building from msbuild cmd line or VS. It was happening because at some places in the code resource was getting accessed through URI which only makes sense for packaged app and not unpackaged. To remedy this, other parts of the code had separate code under UNPACKAGED which would build a non-URI path for unpackaged app config. This UNPACKAGED constant was not defined anywhere and as a result, that alternative code was not getting accessed. Some places had missing alternative code entirely.

This PR removes dependence on UNPACKAGED compile time variable entirely.
This PR side steps compile time variable for a easier to maintain runtime function which uses `Appmodel/GetCurrentPackageId()` system api to determine whether an app is running in packaged mode or unpackaged mode during runtime.

Single project configuration
As a need arose for cleaning up the project files for unpackaged winui gallery to work, I used this opportunity to build upon @Scottj1s 's work in unpkg-module branch and get rid of WAP configuration, have single project configuration for building any flavor of WinUI gallery.
As a result, using launch configurations, you can build WinUI gallery in either packaged or unpackaged mode.
I have introduced more solution configurations to aid in that.
image

257696564-a68328a2-d611-45cb-b217-1f3577d653c6 1
257696536-e3d92c55-39e2-48b4-a49c-8a7016bc67bd 1

If you want to build a Packaged winui gallery,
select Packaged launch configuration and choose between Debug or Release solution configs.

If you want to build a Unpackaged winui gallery,
select Unpackaged launch configuration and choose between Debug-Unpackaged or Release-Unpackaged solution configs.

A mix and match config will result in error because that's how Visual Studio works.


WinUIGallery/Package.appxmanifest
```diff
@@ -10,20 +10,20 @@
   <Identity Name="Microsoft.WinUI3ControlsGallery" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="2.1.0.0" />
   <mp:PhoneIdentity PhoneProductId="863667e0-667a-4bb4-ac52-c59656c7333a" PhonePublisherId="00000000-0000-0000-0000-000000000000" />
   <Properties>
-    <DisplayName>WinUI 3 Gallery</DisplayName>
+    <DisplayName>WinUI 3 Gallery Dev</DisplayName>
```
bpulliam, line 13: revert this line.  Debug builds should use the Package.Dev manifest and the dev identity.  Release should use the Package manifest and the non-dev identity so we can publish that build to the store.

WinUIGallery/Package.appxmanifest
```diff
@@ -10,20 +10,20 @@
...
-    <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.17134.0" MaxVersionTested="10.0.18362.0" />
+      <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.17763.0" MaxVersionTested="10.0.19041.0" />
   </Dependencies>
   <Resources>
     <Resource Language="x-generate" />
   </Resources>
   <Applications>
     <Application Id="App" Executable="$targetnametoken$.exe" EntryPoint="$targetentrypoint$">
-      <uap:VisualElements DisplayName="WinUI 3 Gallery" Square150x150Logo="Assets\Tiles\squaretile-sdk.png" Square44x44Logo="Assets\Tiles\SmallTile-sdk.png" Description="WinUI 3 Gallery" BackgroundColor="transparent">
+      <uap:VisualElements DisplayName="WinUI 3 Gallery Dev" Square150x150Logo="Assets\Tiles\squaretile-sdk.png" Square44x44Logo="Assets\Tiles\SmallTile-sdk.png" Description="WinUI 3 Gallery Debug" BackgroundColor="transparent">
```
bpulliam, line 26: revert

WinUIGallery/Package.appxmanifest
```diff
@@ -41,8 +41,8 @@
           </uap3:AppUriHandler>
         </uap3:Extension>
         <uap:Extension Category="windows.protocol">
-          <uap:Protocol Name="winui3gallery">
-            <uap:DisplayName>WinUI 3 Gallery</uap:DisplayName>
+          <uap:Protocol Name="winui3gallerydev">
```
bpulliam, line 44: revert

bpulliam (changes requested):

WinUIGallery/Package.appxmanifest
Scottj1s: there's also a dependency on DisplayName in WinUIGalleryDesktopTests.cs

bpulliam (approved):

PR #1332, opened 2023-08-24
Build Errors and Warning Fixes

karkarl:
## Description

- Add missing namespace in BreadCrumBar.xaml.cs
- Add `#Nullable enable` to disable build warning
- Move compile include line to more appropriate location.


Scottj1s (approved):
thanks!

PR #1343, opened 2023-08-28
[Design guidance] Automatically expand the "Design guidance" section on load

citelao:
Today, our design guidance section is hard to discover. This change auto-expands the **Design guidance** section in the navigation view.

## Description

* When the UI loads, expand the Design guidance section

Unfortunately, NavigationView items cannot be expanded before the control renders (or at least, until some unknown point after `Loaded`), so the expansion is a visible animation immediately after load.

## Motivation and Context

We worked hard on our design guidance section, but it is not very discoverable. How many of our users know we have an accessibility section? We should expand it by default.

## How Has This Been Tested?

Deployed locally.

* [x] Design guidance is expanded by default.
* [x] Design guidance can be collapsed.
* [ ] ~~Design guidance remembers state after expanding/collapsing sidebar~~ It reverts to closed :(

## Screenshots (if appropriate)

Expanded design guidance section

## Types of changes


karkarl:
NavView sub items should not be expanded by default. The gallery should be showcasing proper UX for windows design.

citelao:
That somewhat explains why all the APIs to expand items did not work on load 😛

Issue #1344, opened 2023-08-29
Fix documentation links to UWP content


bpulliam:
### Which version of the app?

WinUI 3 Gallery

### Description

We have a few links that point to UWP docs instead of WinAppSDK.

- [x] 1. https://learn.microsoft.com/uwp/api/windows.ui.xaml.controls.fonticon should be https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.fonticon
- [ ] 2. https://docs.microsoft.com/en-us/windows/uwp/controls-and-patterns should be https://learn.microsoft.com/windows/apps/design/controls
- [x] 3. https://learn.microsoft.com/windows/uwp/debug-test-perf/version-adaptive-apps
- [x] 4. https://learn.microsoft.com/windows/uwp/composition/composition-animation
- [ ] 5. https://learn.microsoft.com/windows/uwp/composition/using-the-visual-layer-with-xaml
- [ ] 6. https://learn.microsoft.com/uwp/api/windows.applicationmodel.datatransfer.clipboard
- [ ] 7. https://learn.microsoft.com/uwp/api/Windows.Storage.Pickers
- [ ] 8. https://learn.microsoft.com/windows/uwp/contacts-and-calendar/images/shoulder-tap-static-

payload.png

chingucoding:
@karkarl Would you mind if I create a PR for this?

karkarl:
> @karkarl Would you mind if I create a PR for this?

Go for it!

Jay-o-Way:
Speaking of links... The specific parts of the URLs for point 2 all start with a `/` which gives the concatenated URL a double `...//...`. Functionally no problem, so no priority, just wanted to mention it.

PR #1347, opened 2023-08-30
First round of UWP link replacement

chingucoding:
## Motivation and Context

This PR updates links to point to non UWP documentation as per #1344

However, I was unable to find suitable replacements for the following APIs, there only seems to exist UWP documentation for these classes:
- https://learn.microsoft.com/uwp/api/Windows.Storage.Pickers
- https://learn.microsoft.com/uwp/api/windows.applicationmodel.datatransfer.clipboard

For the following link, I'm not really sure what to do. There is no WinUI version of this but I'm also not sure why this was used in the first place (currently only referenced here: https://github.com/microsoft/WinUI-Gallery/blob/main/WinUIGallery/ControlPages/PersonPicturePage.xaml)
- https://learn.microsoft.com/windows/uwp/contacts-and-calendar/images/shoulder-tap-static-payload.png 

## Types of changes


WinUIGallery/Controls/ControlExample.xaml
```diff
@@ -80,7 +80,7 @@
             IsTextSelectionEnabled="True"
             Visibility="Collapsed">
             <Paragraph>
-                This sample requires a later version of Windows to be fully functional. Learn more about version adaptive apps:<Hyperlink NavigateUri="https://learn.microsoft.com/windows/uwp/debug-test-perf/version-adaptive-apps">https://learn.microsoft.com/windows/uwp/debug-test-perf/version-adaptive-apps</Hyperlink>
+                This sample requires a later version of Windows to be fully functional.
```
chingucoding, line 83: Do we still need this or can we remove this feature? WinUI 3 isn't really platform agnostic anymore and I don't think we actually need to show a warning for samples that only work on specific OS versions.

bpulliam: Your change looks good.  We should keep this check, which is based on the ApiContract versions.

bpulliam (approved):

WinUIGallery/Controls/ControlExample.xaml
chingucoding: Sure let's keep it though I don't think we actually still use this. Do you happen to know what the replacement link would be or do we keep the changes as is and remove the info on how to do such a check?

PR #1339, opened 2023-08-28
Upgrading to Toolkit 8.x

niels9001:
Upgrading to our new Toolkit 8.x packages.

This PR:

- Moves Toolkit packages to 8.0
- Removes the Labs preview feed, as packages are now available on the public nuget feed.
- Updating a bunch of Toolkit related namespaces
- XAML styling kicked in when editing a few .XAML packages

PR #1349, opened 2023-09-01
Restore WAP solution and project for bundle publishing 

Scottj1s:
(and remove win32)


Scottj1s:
@niels9001, @karkarl I had to revert https://github.com/microsoft/WinUI-Gallery/pull/1339 because it introduced a new dependency on WinUI 2.7.  The Microsoft.UI.Xaml.dll in WinUI 2.7 then overrides the one from WinAppSDK and does not export XamlCheckProcessRequirements, causing a crash at startup.

michael-hawker:
@Scottj1s there shouldn't be any change to the WinUI dependency with the package changes... We've been referencing 2.7 for months:

https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/main/MultiTarget/PackageReferences/Uwp.props

The app should override this right? An app using WinUI 2.8 should be able to use a library built against 2.7?

Oh, this is the WinUI 3 gallery, so this is the bug again about multi-targeting for UWP and Windows App SDK in the same package? That was fixed in WASDK 1.3 I thought? [This was the internal link for this before](https://dev.azure.com/microsoft/WinUI/_git/microsoft-ui-xaml-lift/pullrequest/8490329#) related to [this issue we had](https://github.com/CommunityToolkit/Labs-Windows/issues/262). Maybe this regressed in 1.4, though we just tried a 1.4 app today and didn't have issues.

Since this was fixed in 1.3, I re-enabled multi-targeting and we hadn't seen issues elsewhere yet.

Scottj1s:
> @Scottj1s there shouldn't be any change to the WinUI dependency with the package changes... We've been referencing 2.7 for months:

The last commit to this PR is the revert, to make it easy to compare.  If you 'nuget restore -force WinUIGallery.DesktopWap.sln' with and without that commit, you see in .\WinUIGallery\obj\WinUIGallery.DesktopWap.Package\project.assets.json that all the new WCT package references now bring in "Microsoft.UI.Xaml": "2.7.0".

michael-hawker:
@Scottj1s I'm aware that's what happening, I'm saying is that we saw that before pre-1.3 too. It was because our package multitargeted `uap` and `net6.0-windows` that pulled both Microsoft.UI.Xaml versions in and caused a conflict because something in WASDK was pulling in the UAP targets.

That was supposedly fixed in 1.3, so we added the multi-targeting to our package back in which the latest rc has but not the older labs packages. We haven't seen this issue elsewhere though with the WASDK again with the new packages...

michael-hawker:
The TFMs look good on the new packages. @niels9001 I think I forgot to ask if you tested them in your wapproj test you had. Think I confused it with the other test we were doing. `MainLatest` feed should have from this build when it's done: https://github.com/CommunityToolkit/Windows/actions/runs/6102571346

@Scottj1s would it be best for us to open an issue in the WASDK GitHub for the wapproj problem in general for the future?

Also, we decided we should release in the PT AM, so polishing up everything this afternoon and packages should be up on Nuget.org tomorrow with build number `8.0.230907` if all goes well (lets me sync with Niels for final validation too 😋). 🤞

niels9001:
> The TFMs look good on the new packages. @niels9001 I think I forgot to ask if you tested them in your wapproj test you had. Think I confused it with the other test we were doing. `MainLatest` feed should have from this build when it's done: https://github.com/CommunityToolkit/Windows/actions/runs/6102571346
> 
> @Scottj1s would it be best for us to open an issue in the WASDK GitHub for the wapproj problem in general for the future?
> 
> Also, we decided we should release in the PT AM, so polishing up everything this afternoon and packages should be up on Nuget.org tomorrow with build number `8.0.230907` if all goes well (lets me sync with Niels for final validation too 😋). 🤞

Tested in a standalone WAP proj, and the changes seem to work :-)! I'll add a PR to re-revert the SettingsCard changes and test it once we have everything on Nuget.org!

Scottj1s:
> > @Scottj1s would it be best for us to open an issue in the WASDK GitHub for the wapproj problem in general for the future?
> 
> Tested in a standalone WAP proj, and the changes seem to work :-)! I'll add a PR to re-revert the SettingsCard changes and test it once we have everything on Nuget.org!

I tested in WinUI 3 Gallery, and the changes work there as well.

Yes, let's open an issue in WASDK github to protect against inadvertently combining WinUI 2.7 and WinAppSDK.

Scottj1s:
@niels9001 thanks for the fix!
@karkarl I've updated to CommunityToolkitWinUIVersion 8.0.230907

bpulliam (approved):

PR #1357, opened 2023-09-15
Added a "Capture element / camera preview" page in the Media section.

codendone:
## Description
`CaptureElement` from UWP XAML does not exist in WinAppSDK, but `MediaPlayerElement` can be used in its place. The change adds a new sample page to demonstrate how this can be done.

Note: Currently this page just reuses the same icon as the MediaPlayerElement page. Ideally at some point we'll get a camera icon which could be used here instead.

## Motivation and Context
Demonstrate the couple of calls to get from a `MediaCapture` to an `IMediaPlaybackSource` necessary to display the capture in a MediaPlayerElement. This also demonstrates basic creation of a `MediaCapture` to provide an end-to-end sample. This sample should help address questions in microsoft/microsoft-ui-xaml#4710 and microsoft/microsoft-ui-xaml#8214.

## How Has This Been Tested?
Ad-hoc testing and testing with Accessibility Insights for Windows tool.

## Types of changes


WinUIGallery/DataModel/ControlInfoData.json
```diff
@@ -2009,6 +2009,24 @@
             }
           ]
         },
+        {
+          "UniqueId": "CaptureElementPreview",
+          "Title": "Capture element / camera preview",
```
karkarl, line 2014: Should the title be MediaCapture to be consistent with the page naming convention of using the control name. 

If not, the title should be capitalized.

codendone: The intent was to have the title more directly mirror "CaptureElement", since this demos the alternative which is available in WinAppSDK without CaptureElement.

Capitalizing makes sense.

karkarl (approved):

PR #1361, opened 2023-09-22
Add Test to Scan All Pages for Axe Issues

karkarl:
## Description

- Add a test that physically navigates to each page via NavView and check for Axe issues.
- Utilizes DynamicData attribute to create grouped tests based on section category (ie.  TreeView parent)
- This also tests for run-time crashes with each page. 
- Miscellaneous Axe fixes.

## Types of changes


UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Text.Json;
+using System.IO;
+using System.Collections.ObjectModel;
+
+namespace UITests.Tests
+{
+    [TestClass]
+    public class AxeScanAll : TestBase
+    {       
+        public static readonly string jsonUri = "ms-appx:///../../../../../../WinUIGallery/DataModel/ControlInfoData.json";
```
chingucoding, line 14: Should we instead add the file to the project to copy it to the output directory?

UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        public static new WindowsDriver<WindowsElement> Session => SessionManager.Session;
+
+        public class ControlInfoData
+        {
+            public ObservableCollection<ControlInfoDataGroup> Groups { get; set; }
+        }
+
+        public class ControlInfoDataGroup
```
chingucoding, line 22: I'm curious what you think about this data duplication, should these classes exist in a shared project and be referenced by both the Gallery and this project?

UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                    var controlItem = Session.FindElementByAccessibilityId(controlName);
+                    controlItem.Click();
+
+                    AxeHelper.AssertNoAccessibilityErrors(controlName);
+                }
+            }
+        }
+        
+        public static void ParseJson()
+        {;
```
chingucoding, line 117: Stray semicolon?

WinUIGallery/Navigation/NavigationRootPage.xaml.cs
```diff
@@ -221,12 +222,13 @@ private void AddNavigationMenuItems()
...
                     itemGroup.MenuItems.Add(itemInGroup);
                     AutomationProperties.SetName(itemInGroup, item.Title);
+                    AutomationProperties.SetAutomationId(itemInGroup, item.UniqueId);
                 }
 
                 NavigationViewControl.MenuItems.Add(itemGroup);
             }
 
-            NewControlsItem.Loaded += OnNewControlsMenuItemLoaded;
+            Home.Loaded += OnNewControlsMenuItemLoaded;
```
chingucoding, line 231: Thank you for renaming this item to what it actually represents now 🙈

UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+            public string Title { get; set; }
+            public string Uri { get; set; }
+        }
+
+        private static ControlInfoData controlInfoData;
+
+        private string[] ExclusionList =
+        {
+            "WebView2" // 46668961: Web contents from WebView2 are throwing null BoundingRectangle errors.
```
chingucoding, line 72: I'm curious if this is an issue with WebView2, the page we show or Axe. In the last case, maybe we should open an issue on their repo?

UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        [ClassInitialize]
+        public static void ClassInitializeAsync(TestContext context)
+        {
+            ParseJson();
+        }
+
+        [TestMethod]
+        [TestProperty("Description", "Scan all controls in the WinUIGallery for accessibility issues.")]
+        public void ValidateAccessibilityWithAxe()
```
kmahone, line 83: This will result in a single test case that covers every page in the app. This has some advantages. One is that it takes a very long time to run. Another is that if it fails on one page, it will stop on failure and the rest of the pages will not execute. And if you want to re-run the test to investigate a failure you have to run the whole thing, even if you only care about a single page.

A better approach would be to use data-driven testing. With this approach you have a single test method, but it is run multiple times with different data each time.

This is actually how we test WinUI Gallery in the WinUI3 repo. Take a look at:
controls\test\MUXControls.Test\ScenarioAppTests\WinUIGalleryDesktopTests.cs

The test method has this attribute:
```
[DataSource("Table:WinUIGalleryTestData.xml#MenusAndToolbars")]
```
Which means that it will pull in data from the file "WinUIGalleryTestData.xml". This is a file that is generated at build time by parsing ControlInfoData.json.

You won't be able to directly re-use this code. The WinUI3 repo uses TAEF, but this repo uses MSTest. But MSTest also supports data-driven testing.

I haven't uses MSTest's data driven features before, so I can't comment on exactly how to use it. But there is documentation here:
[https://learn.microsoft.com/en-us/visualstudio/test/how-to-create-a-data-driven-unit-test?view=vs-2022](https://learn.microsoft.com/en-us/visualstudio/test/how-to-create-a-data-driven-unit-test?view=vs-2022)

These docs seem to imply that only a sql db can be used as a source. But it sounds like it should be possible to use xml and other formats. Have a look around and see what you can find. I came across these articles:
[https://www.meziantou.net/mstest-v2-data-tests.htm](https://www.meziantou.net/mstest-v2-data-tests.htm)
[https://visualstudiomagazine.com/articles/2009/09/15/unit-testing-with-vsts2008-part-3.aspx](https://visualstudiomagazine.com/articles/2009/09/15/unit-testing-with-vsts2008-part-3.aspx)
[https://mariusbancila.ro/blog/2019/11/28/data-driven-unit-tests-for-managed-code/](https://mariusbancila.ro/blog/2019/11/28/data-driven-unit-tests-for-managed-code/)

WinUIGallery/ControlPages/DesignGuidance/IconsPage.xaml
```diff
@@ -110,8 +110,12 @@
             <RowDefinition Height="*" />
         </Grid.RowDefinitions>
         <TextBlock Text="With the release of Windows 11, Segoe Fluent Icons is the recommended icon font." />
+
+        <!-- https://github.com/CommunityToolkit/Windows/issues/240 -->
+        <!-- External toolkit SettingsExpander does not pass Axe testing-->
         <toolkit:SettingsExpander
             Grid.Row="1"
+            AutomationProperties.AccessibilityView="Raw"
```
kmahone, line 118: >AutomationProperties.AccessibilityView="Raw" [](http://example.com/codeflow?start=12&length=44)

This seems like the wrong approach. We should update the test to skip this control instead of moving it to the Raw view.

Moving this control to the Raw view will result in the app not working correctly in Narrator and other Assistive Technologies.

kmahone (changes requested):
:clock1:

chingucoding:
Given that we have quite a lot of tests now (assuming we split as @kmahone suggested), would it be worth thinking about splitting up the tests into groups for the CI, i.e. have the CI use multiple runners and split tests based on some criteria?

UITests/Tests/AxeScanAllTests.cs
karkarl: The issue is already open on their repo :) I should follow up with them actually.

UITests/Tests/AxeScanAllTests.cs
karkarl: Thanks for the info. I've updated the tests do be data driven testing. Since `TestContext.DataRow` isn't supported in .net, I'm using the DynamicData attribute as recommended in the docs.

karkarl:
> Given that we have quite a lot of tests now (assuming we split as @kmahone suggested), would it be worth thinking about splitting up the tests into groups for the CI, i.e. have the CI use multiple runners and split tests based on some criteria?

Yes that will be very helpful, and will be the next on the backlog.

UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    [TestClass]
+    public class AxeScanAll : TestBase
+    {       
+        public static readonly string xmlUri = "WinUIGalleryTestData.xml";
+        public static new WindowsDriver<WindowsElement> Session => SessionManager.Session;
+
+        private static IEnumerable<object[]> TestData()
+        {
+            var testCases = new List<object[]>();
```
kmahone, line 24: Since you are doing your own parsing of the data, would it be simpler to directly parse the ControlInfoData.json file instead of having a separate script to convert ControlInfoData.json into WinUIGalleryTestData.xml and then parsing that xml file?

For the Lifted Xaml repo, I needed to convert from json into an xml format that taef understood. But since you re doing your own parsing here, you don't have to use xml at all.

UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        [ClassInitialize]
+        public static void ClassInitialize(TestContext context)
+        {
+        }
+
+        [TestMethod]
+        [DynamicData(nameof(TestData), DynamicDataSourceType.Method, DynamicDataDisplayName = nameof(GetCustomDynamicDataDisplayName))]
+        [TestProperty("Description", "Scan pages in the WinUIGallery for accessibility issues.")]
+        public void ValidatePageAccessibilityWithAxe(string tableId, List<string> rowNames)
```
kmahone, line 60: >string tableId, List<string> rowNames [](http://example.com/codeflow?start=53&length=37)

Consider giving clearer names for these parameters.

I think tableId corresponds to the sections of the app (e.g. Basic Input, Layout, etc.) and rowNames is the pages in that section? Is that right?

UITests/Tests/AxeScanAllTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            // Expand tree view and check for page accessibility.
+            var page = Session.FindElementByAccessibilityId(tableId);
+            page.Click();
+
+            AxeHelper.AssertNoAccessibilityErrors();
+
+            // Click into each page and check for accessibility issues.
+            foreach (var rowName in rowNames)
```
kmahone, line 69: >foreach (var rowName in rowNames) [](http://example.com/codeflow?start=12&length=33)

Instead of looping over rowNames, could we have a unique test for each page? 

i.e.
ValidatePageAccessibilityWithAxe(string section, string page)

which gets called for each page in the app?

This results in smaller tests, which has the advantages mentioned in my previous comment about data-driven testing.

UITests/Tests/AxeScanAllTests.cs
karkarl: I believe it's easier to parse the .xml file instead of the .json. The json file has quite a lot of info that will need to be parsed, but is not relevant to the tests. This method adds an extra file, but results in cleaner test code.

chingucoding: What about using the same parsing code from the WinUI 3 Gallery? Since we are using the same .NET SDK, we could move them in a shared project so both projects can reference the same data and parsing.

kmahone: You are already parsing the json file. You are just doing it in powershell as part of the build. You can move that same logic out of powershell and into the test code.

karkarl: @kmahone Done.
@chingucoding That can definitely be a future refactor, but too big of a change for this PR in my opinion.

PR #1377, opened 2023-10-10
Turn off trimming (until it works with Gallery). Turn off Self-Contai…

bpulliam:
- Turn off trimming (until it works with Gallery)
- Turn off Self-Contained for public repo

## Description
Trimming does not yet work and was causing crashes in Release builds.
This will be re-enabled once the issues are resolved.


Scottj1s (approved):

Scottj1s (changes requested):
holding PR while investigating trimming issue (since this worked with .NET 6)

Scottj1s:
see https://github.com/microsoft/WinUI-Gallery/pull/1378

PR #1396, opened 2023-11-02
Remove WindowHelper from FilePicker sample code

AndrewKeepCoding:
## Description
Removed the WinUI3Helper class from the `FilePicker` sample code and showed how to get the `MainWindow` from the `App` class instead.

## Motivation and Context
The sample code for the `FilePicker` controls uses the WindowHelper class which is implemented as a helper class in this project. This is very confusing because it looks like the WindowHelper class belongs to WinUI 3 and when one tries to use that code in their apps it just doesn't work. I find several questions in StackOverflow related to this and I think it's better not to use the WindowHelper in sample code.

## How Has This Been Tested?
Run the app and confirmed its content.

## Types of changes


AndrewKeepCoding:
@microsoft-github-policy-service agree

karkarl (commented):
Thanks for the fix!

karkarl (approved):

MartinZikmund:
@AndrewKeepCoding shouldn't the code snippets say App.MainWindow instead 👀?

AndrewKeepCoding:
@MartinZikmund Thanks for pointing that out. 😵

PR #1397, opened 2023-11-02
Fix Listbox text truncation when text is scaled up

christineyan4:
## Description
This change updates the Width set on ListBox to be MinWidth instead.

## Motivation and Context
This fixes an accessibility issue in WinUI3 Gallery, where ListBox would truncate longer text when text scaling is sized up. 

## How Has This Been Tested?
Ad-hoc testing

## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/ListBoxPage.xaml
+++ b/WinUIGallery/ControlPages/ListBoxPage.xaml
@@ -22,7 +22,6 @@
                         <x:String>Green</x:String>
                         <x:String>Red</x:String>
                         <x:String>Yellow</x:String>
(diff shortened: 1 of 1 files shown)


WinUIGallery/ControlPages/ListBoxPage.xaml
```diff
@@ -17,7 +17,7 @@
         <local:ControlExample x:Name="Example1" HeaderText="A ListBox with items defined inline and its width set.">
```
RBrid, line 17: its minimum width set

RBrid (approved):

PR #1405, opened 2023-11-17
[Cleanup] Remove unused controlicons

niels9001:
This PR removes a bunch of unused control icon assets and the links in the `ControlDataInfo.json`. It also renames `ImageIconPath` to `IconGlyph` as we are using those for the navview and are actual icons vs. image :).


chingucoding (approved):
LGTM, thanks for cleaning that up. Is it correct that some controls/sections still don't have an icon (e.g. the text section)?

WinUIGallery/ControlPages/SemanticZoomPage.xaml
```diff
@@ -25,7 +25,6 @@
 
         <DataTemplate x:Key="ZoomedInTemplate" x:DataType="data:ControlInfoDataItem">
             <StackPanel Orientation="Horizontal" MinWidth="200" Margin="12,6,12,6">
```
Jay-o-Way, line 27: This stackpanel now has only one child, which makes it useless :⁠-⁠)

Jay-o-Way (changes requested):

PR #1406, opened 2023-11-17
change Windows App Runtime version to actually reflect runtime, not SDK

Scottj1s:
Note: Microsoft.WindowsAppRuntime.Insights.Resource.dll is the only module that carries complete version info.

### DIFF
--- a/WinUIGallery/App.xaml.cs
+++ b/WinUIGallery/App.xaml.cs
@@ -7,26 +7,20 @@
 // PURPOSE, MERCHANTABILITY, OR NON-INFRINGEMENT.
 //
 //*********************************************************
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
 using AppUIBasics.Common;
 using AppUIBasics.Data;
 using AppUIBasics.Helper;
-using System;
-using System.Linq;
-using System.Reflection;
-using System.Text.RegularExpressions;
-using System.Threading.Tasks;
-using Windows.ApplicationModel;
-using Windows.ApplicationModel.Activation;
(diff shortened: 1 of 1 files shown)


WinUIGallery/App.xaml.cs
```diff
@@ -42,15 +36,20 @@ public static string WinAppSdkDetails
...
+        }
+
         public static string WinAppSdkRuntimeDetails
         {
             get
-            {
-                var details = WinAppSdkDetails;
+            {
+                var details = WinAppSdkDetails;
 #if WindowsAppSdkRuntimeDependent
```
codendone, line 46: >WindowsAppSdkRuntimeDependent [](http://example.com/codeflow?start=4&length=29)

With the new implementation, is there any need to condition this?

Scottj1s: well no, come to think of it

codendone (approved):
:shipit:

Jay-o-Way:
Question: what if there is an issue with something like `windowsAppRuntimeVersion.First()`? Like, empty collection or something like that...

Scottj1s:
> Question: what if there is an issue with something like `windowsAppRuntimeVersion.First()`? Like, empty collection or something like that...

First() will throw InvalidOperationException if the collection is empty.  This is not expected here.

Jay-o-Way:
@Scottj1s I was looking at #1528 at the time. #1480 holds the solution.

Scottj1s:
ah, thanks for the explanation.  I'll follow up on https://github.com/microsoft/WinUI-Gallery/pull/1480.

PR #1426, opened 2024-01-12
[proof of concept] List lightweight styling resources

citelao:
Today, it is very difficult to get a list of [lightweight styles](https://learn.microsoft.com/en-us/windows/apps/design/style/xaml-styles#lightweight-styling) for a given control.

This change introduces a **proof of concept** that can automatically fetch a list of lightweight styles for a given control. It does this by reading the Application `ResourceDictionary`s for brushes that have the correct prefixes.

## Thoughts

* How do we order the styles?
* How about non-brush styles?
* Are there some styles that are colors instead?
* How do we generalize this for all controls?

## Screenshots

a simple list of brushes

## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/ButtonPage.xaml
+++ b/WinUIGallery/ControlPages/ButtonPage.xaml
@@ -12,7 +12,8 @@
 <Page x:Class="AppUIBasics.ControlPages.ButtonPage" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
       xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:local="using:AppUIBasics"
       xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      xmlns:pages="using:AppUIBasics.ControlPages"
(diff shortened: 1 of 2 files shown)


niels9001:
@citelao Whoa, this is pretty amazing and super useful!

Just wondering, should this live in a component page? Or should there be a dedicated 'lightweight styling' page where you could select a UI control and then see its styles?

karkarl:
Thanks so much for this @citelao, I'm really looking forward for this feature!

In response to @niels9001 suggestions, here are my two cents: I think the lightweight styling resources should live with the control, since most developers only need this information for the one control they need. I don't think a dialog or flyout will be helpful since it should stay static to allow for easy copy and pasting of the resources. That said, I would recommend it to live in a expander, with copy buttons for each resource name.

citelao:
I strongly agree with @karkarl here: I think this UI should live in the control, since that's where developers will be looking when they lookup "button" etc.

Also:

* +1: I agree that an expander would be nice.
* +1: I agree we need copy-paste.
* I want to get the underlying swatch name if we can (eg if `ButtonFooBrush` maps to `SystemSomethingBrush`, it'd be nice to display that name).

---

I mocked up what I hope might happen (minus copy-paste, which we should add):

an expander with a list of swatches and example controls

When you hover a brush, it's "highlighted" in the sample (by actually replacing the brush in the sample!):

hovered brush highlighted in pink

marcelwgn:
I think if we could get the actual XAML source code of WinUI 3 during runtime and go from that information. That way we could also try to get more brushes that might not be named after the control (e.g. resources that apply for multiple controls). Also have you thought about visualizing non color/brush resources?

marcelwgn:
Closing this PR due to staleness. Feel free to reopen this PR if you wish to continue working on this feature 😊

PR #1444, opened 2024-02-06
Add a unit test project

kmahone:
This adds a set of unit tests to WinUIGallery.

The initial set of tests are mostly just to serve as a demonstration of how to write tests in MSTest for WinUI. But as new features are added to WinUI we can optionally write them in a more unit-testable way and add new tests here. Unit tests have the advantage of being quicker and more reliable that UI Interaction tests which is where we are getting most of the coverage now.

I've also updated the Azure Pipeline job to execute the unit tests as part of the build.

This PR updates the version of WinAppSDK from 1.4 to 1.5 Preview1. Once 1.5.0 stable is released we should update to that.


WinUIGalleryUnitTests/UnitTestApp.xaml
```diff
@@ -0,0 +1,52 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:common="using:AppUIBasics.Common"
+    xmlns:local="using:WinUIGalleryUnitTests">
+    <Application.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
+                <!-- Other merged dictionaries here -->
+            </ResourceDictionary.MergedDictionaries>
+            <ResourceDictionary.ThemeDictionaries>
```
karkarl, line 14: Are these resources pasted over from WinUI Gallery? If not all resources are required, may be worth it to clean it up.

azure-pipelines.yml
```diff
@@ -41,6 +41,19 @@ steps:
         zipAfterPublish: false
         modifyOutputPath: false
 
+    - script: |
+        dir /b /s $(Build.ArtifactStagingDirectory)
+        dir /b /s $(System.DefaultWorkingDirectory)
+      displayName: 'List contents of ArtifactStagingDirectory and DefaultWorkingDirectory'
+
+    - task: VSTest@2
+      displayName: 'Run Unit Tests'
```
karkarl, line 50: Might be worth it to label it as "Run Sample Unit Tests" since these mostly serve as a demonstration rather than an actual test.

karkarl (approved):

marcelwgn:
I'm wondering if having unit tests also warrants having a `tests` and `src` folder where the projects are located similar to a lot of other .Net projects. What do you folks think?

WinUIGallery/IdleSynchronizer.cs
```diff
@@ -13,7 +13,7 @@
 
 namespace AppUIBasics
 {
-    class IdleSynchronizer
+    public class IdleSynchronizer
```
marcelwgn, line 16: Is this related to the unit tests?

WinUIGalleryUnitTests/WinUIGalleryUnitTests.csproj
```diff
@@ -0,0 +1,65 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
marcelwgn, line 1: Do we not need a certificate for the unit types project?

WinUIGallery/IdleSynchronizer.cs
kmahone: At one point I was using it from one of the unit tests. I'll leave it as it is.

WinUIGalleryUnitTests/WinUIGalleryUnitTests.csproj
kmahone: We're not creating an msix so we don't need the cert.

Issue #1279, opened 2023-05-03
Refactor Codebase


karkarl:
Refactor WinUI-Gallery codebase:

```[tasklist]
### Tasks
- [x] Remove unused code
- [x] remove outdated ThemeResources
- [x] Proper xaml formatting
- [x] Proper class name / namespaces
- [x] Accessibility pass
- [x] Clean out unnecessary branches
- [x] Re-name `NewControls*` to `Home`
- [ ] https://github.com/microsoft/WinUI-Gallery/issues/1445
- [ ] https://github.com/microsoft/WinUI-Gallery/issues/1542
```

Jay-o-Way:
Spring clean-up 🧹🪣🧽😛

Jay-o-Way:
@niels9001  Would this include - for example - the following?
https://github.com/microsoft/WinUI-Gallery/blob/0a300bf69a51cf7684e93d95d89d9375130738f3/WinUIGallery/SettingsPage.xaml#L12C2-L20

Those numbers are not by the book; the elements (TextBlock) can probably be replaced with more modern implementations via SettingsGroup (Header)

karkarl:
> 

Yes - the idea is to have WinUI Gallery also be a useful repo to refer to for best coding practices. That includes updating redundant xaml.

Jay-o-Way:
Also a nit I found: https://github.com/microsoft/WinUI-Gallery/blob/0a300bf69a51cf7684e93d95d89d9375130738f3/WinUIGallery/SectionPage.xaml#L40
is unneeded - actually conflicting with line 49: `Style="{StaticResource TitleTextBlockStyle}"`

PR #1446, opened 2024-02-13
Replace AppUIBasics to WinUIGallery  Namespace

karkarl:
Replace AppUIBasics to WinUIGallery Namespace

Part of #1279


nuget.config
```diff
@@ -5,6 +5,7 @@
   </config>
   <disabledPackageSources>
     <clear />
+    <add key="Package source" value="true" />
```
marcelwgn, line 8: Is this related to the namespace rename?

marcelwgn (changes requested):
I think the following three pages are still using the AppUIBasics namespace:
- CollectionInterop.cs
- ImageScrollBehavior.cs
- InputValidationPage.xaml.cs

marcelwgn (approved):

PR #1531, opened 2024-05-02
[Design guidance] Additional pages

niels9001:
## Description
- Updated the design guidance menu names to be inline with the docs (e.g. Icons => Iconography, Colors => Color) and sorted alphabetically.
- Moved the Card stroke brush on the Stroke colors page to the top, so it's inline with the Card background brush and the control strokes are grouped.
- Adding a bit more spacing between the color sections
- Left aligned type ramp image on the Typography page
- Added a new Geometry page explaining how to use the CornerRadius resources:
image

- Added a new Spacing page showing a spacing ramp and some common UI examples with redlines:
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #1541, opened 2024-05-19
Design suggestions

Jay-o-Way:
few suggestions, as also commented in  #1531
feel free to adapt where needed

### DIFF
--- a/WinUIGallery/ControlPages/DesignGuidance/SpacingPage.xaml
+++ b/WinUIGallery/ControlPages/DesignGuidance/SpacingPage.xaml
@@ -38,7 +38,16 @@
                 <Style.Setters>
                     <Setter Property="HorizontalAlignment" Value="Stretch" />
                     <Setter Property="MinHeight" Value="68" />
-                    <Setter Property="CornerRadius" Value="{StaticResource OverlayCornerRadius}" />
+                    <Setter Property="CornerRadius" Value="{StaticResource ControlCornerRadius}" />
+                </Style.Setters>
+            </Style>
+
+            <Style
+                x:Key="GridRowLightStyle"
+                BasedOn="{StaticResource GridCardStyle}"
+                TargetType="Grid">
+                <Style.Setters>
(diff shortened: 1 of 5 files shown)


WinUIGallery/Controls/DesignGuidance/TypographyControl.xaml
```diff
@@ -99,50 +98,46 @@
             </Setter>
         </Style>
     </UserControl.Resources>
-    <Grid MinHeight="68" Background="Transparent">
```
niels9001, line 102: I think we need the padding. 

Left (your changes) vs. right (current):

image

WinUIGallery/ControlPages/DesignGuidance/SpacingPage.xaml
```diff
@@ -38,7 +38,16 @@
                 <Style.Setters>
                     <Setter Property="HorizontalAlignment" Value="Stretch" />
                     <Setter Property="MinHeight" Value="68" />
-                    <Setter Property="CornerRadius" Value="{StaticResource OverlayCornerRadius}" />
+                    <Setter Property="CornerRadius" Value="{StaticResource ControlCornerRadius}" />
+                </Style.Setters>
+            </Style>
+
+            <Style
+                x:Key="GridRowLightStyle"
```
niels9001, line 46: Ah yes, this is good to do.

Should we rename these so they are consistent?

"GridRowCardStyle" and "GridRowLightCardStyle" or something like that?

And should we move these to App.xaml.cs so they can be re-used across these 2 pages?

niels9001 (changes requested):
Thanks @jay-o-way, great changes! Some small things to fix, and then I'll merge this in!

WinUIGallery/ControlPages/DesignGuidance/SpacingPage.xaml
Jay-o-Way: Sure, do your thing! 👍🏻

PR #1548, opened 2024-05-21
Added a Scratch Pad page where sample xaml can be loaded

codendone:
## Description
Added a “Scratch Pad” page to the gallery where the user can type in some XAML markup and click the Load button to immediately see the results. This provides a convenient place for experimentation. It can also be used to provide a simple repro for issues which only require loading the right XAML markup.

This change currently adds the “Scratch Pad” page to the “System” group in the navigation pane. This seems like the best fit currently, without a “Tools” section or something else.

## Details about the Scratch Pad
* There is syntax coloring for the XAML in the RichEditBox on the page. The syntax coloring only updates on Load.
* The user can press F5 when focus is in the RichEditBox to trigger a Load.
* The XAML markup is stored in application data on Load. This stored XAML is loaded the next time the user returns to Scratch Pad, even on a fresh launch of WinUI 3 Gallery.
* There is some basic auto-indenting, auto close-tag insertion, and auto property-quotes insertion handling.

## Known issues/limitations
* Syntax coloring only updates on Load.
* The undo stack of the RichEditBox contains an extra undo every time the syntax coloring is applied. (This is the main reason it isn’t applied every time the text changes.)
* `{x:Bind}` is not supported and will cause a load error. A possible future improvement would be to strip out any such properties to load as much as possible. 
* Events (like Button.Click event) appear to not cause errors, but don’t get connected to anything.
* If either the content or the xaml markup textbox get too tall, they expand to take as much space as needed. It would be nicer for individual ScrollViewers to scroll when needed, which would require overriding the standard ScrollViewr on ItemsPage.xml.

## How Has This Been Tested?
Tested loading a variety of markup. Tested its ability to save and load small and large markup to app data. Tested the "Toggle Theme" button to ensure it behaves well. Tested with Accessibility Insights to ensure there weren't any issues.

image

## Types of changes

## Some possible future improvements
* Maybe support choosing from multiple default samples to load.
* Maybe support saving and loading multiple user-named samples.
* Maybe add a way for every control sample in the app to be able to navigate to Scratch Pad with the content of that sample.

## Bonus fix
Removed an unused variable in IconsPage.xaml.cs to remove a compiler warning.


WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
```diff
@@ -0,0 +1,575 @@
...
+
+            // If there is no stored XAML, load the default.
+            if (xamlStr == null || xamlStr.Trim().Length == 0)
+            {
+                xamlStr = GetDefaultScratchXAML();
+            }
+
+            m_oldText = xamlStr;
+            textbox.TextDocument.SetText(Microsoft.UI.Text.TextSetOptions.None, m_oldText);
+            var formatter = new XamlTextFormatter(textbox);
```
Scottj1s, line 44: make formatter a member?

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
```diff
@@ -0,0 +1,575 @@
...
+
+        public string ReadScratchPadXAMLinLocalSettings()
+        {
+            var appData = Windows.Storage.ApplicationData.Current;
+            if (appData.LocalSettings.Containers.ContainsKey("ScratchPad"))
+            {
+                var scratchPadContainer = appData.LocalSettings.CreateContainer("ScratchPad", Windows.Storage.ApplicationDataCreateDisposition.Existing);
+                if (scratchPadContainer != null && scratchPadContainer.Values.ContainsKey("ScratchPadXAML"))
+                {
+                    // String values are limited to to 4K characters. Use a composite value to support a longer string.
```
Scottj1s, line 82: "to to"

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
```diff
@@ -0,0 +1,575 @@
...
+                    var compositeStr = scratchPadContainer.Values["ScratchPadXAML"] as ApplicationDataCompositeValue;
+                    var xamlStr = "";
```
Scottj1s, line 84: consider [StringBuilder](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?view=net-8.0) for performance

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
```diff
@@ -0,0 +1,575 @@
...
+                }
+            }
+            return null;
+        }
+
+        public void SaveScratchPadXAMLinLocalSettings(string xamlStr)
+        {
+            var appData = Windows.Storage.ApplicationData.Current;
+            var scratchPadContainer = appData.LocalSettings.CreateContainer("ScratchPad", Windows.Storage.ApplicationDataCreateDisposition.Always);
+            // String values are limited to to 4K characters. Use a composite value to support a longer string.
```
Scottj1s, line 100: to to

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
```diff
@@ -0,0 +1,575 @@
...
+            }
+            return "";
+        }
+
+        private void HandleEnter()
+        {
+            textbox.TextDocument.BeginUndoGroup();
+            string previousLine = GetTextboxTextPreviousLine();
+            string indentStr = previousLine.Substring(0, previousLine.Length - previousLine.TrimStart().Length);
+            // TODO: Ident further if this is the start of a child tag or property.
```
Scottj1s, line 219: Ident

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
Scottj1s: or better yet, use C#/WinRT string marshaling, which can create a fast-pass HSTRING (reference) for writing:
https://github.com/microsoft/CsWinRT/blob/master/docs/interop.md#strings

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
```diff
@@ -0,0 +1,575 @@
...
+
+        private void textbox_ActualThemeChanged(FrameworkElement sender, object args)
+        {
+            // Updating the formating for theme change
+            var formatter = new XamlTextFormatter(textbox);
+            formatter.ApplyColors();
+        }
+    }
+
+    public class XamlTextFormatter
```
Scottj1s, line 397: Did you consider ICSharpCode's controls?  they have rich support for syntax highlighting, indent, etc.

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
codendone: That might be useful if in the future XamlTextFormatter supports incremental parsing and updates. Today it doesn't have that and I think doesn't really benefit from caching. Won'tFix.

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
codendone: Could use a StringBuilder, but this loop will typically run only once or twice, and only on page load or Reset. I think a StringBuilder isn't necessary in these conditions.

WinUIGallery/ControlPages/ScratchPadPage.xaml.cs
codendone: I didn't want to add any any dependencies just for Scratch Pad. I hadn't intended to add any editor functionality, but ended up deciding that some quality-of-life improvements were needed.

For ICSharpCode, I see a WinForms and maybe WPF version, but I don't see any indicator of a drop-in solution for WinUI3.

ColorCode.WinUI, which Gallery already depends on, could be a future option. Currently it only works with RichTextBlock, which isn't very useful for an edit box scenario. But it may be extendable to RichEditBox. Likely RichEditBox would need to have some added features to address the undo stack issue mentioned in the PR description.

karkarl:
Thanks for this PR, I'm very excited for this! We can create a separate NavView Item and put this page under "Tools". This can be in the next PR.

PR #1554, opened 2024-05-26
Fix: No contrast of the CopyButton in the color guidance page

HO-COOH:
## Description
The `CopyButton`'s `PointerOver` state is bind to `ButtonPressedForeground` brush, and it will be white in a dark theme, making the button invisible when hoveed in that page.
image

It should be bind to its `Foreground` so it remains invisible. So after this fix:
image

## How Has This Been Tested?
Manual

## Types of changes


WinUIGallery/Controls/DesignGuidance/ColorTile.xaml
```diff
@@ -53,7 +53,225 @@
...
+                                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="RootGrid" Storyboard.TargetProperty="BorderBrush">
+                                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ButtonBorderBrushPointerOver}" />
+                                                        </ObjectAnimationUsingKeyFrames>
+                                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="Foreground">
+                                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{Binding Foreground, RelativeSource={RelativeSource Mode=TemplatedParent}}" />
+                                                        </ObjectAnimationUsingKeyFrames>
+                                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="CopySuccessGlyph" Storyboard.TargetProperty="Foreground">
+                                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{Binding Foreground, RelativeSource={RelativeSource Mode=TemplatedParent}}" />
```
karkarl, line 192: You should be able to directly update the copyButton controls under .\controls\CopyButton\CopyButton.xaml

This will avoid the redundancy to re-template here.

karkarl (changes requested):

WinUIGallery/Controls/DesignGuidance/ColorTile.xaml
HO-COOH: The `ButtonForegroundPointerOver` and `ButtonBackgroundPressed` is not purely black and purely white in the original template, and you have set `Foreground` in the original template so `TargetNullValue` cannot be used.

Jay-o-Way:
Hey all, is this issue still valid in the latest build? 'cause I can't reproduce it locally!

WinUIGallery/Controls/DesignGuidance/ColorTile.xaml
Jay-o-Way: I agree this isn't the solution.

Jay-o-Way:
Here's a weird idea? What would happen if we remove **all** of this manual work and reset the button to defaults?! 🤓

Jay-o-Way:
...On-topic...
On rare occasions, the text **foreground** of the **ColorTile** control is overwritten to deal with the background. That is, to prevent white-on-white. Is there a possibility to override the foreground of the **CopyButton** THROUGH the ColorTile control?

https://github.com/microsoft/WinUI-Gallery/blob/a9796b9817bef255e96820ada5d8c3c129b4519a/WinUIGallery/Controls/DesignGuidance/ColorSections/TextSection.xaml#L177-L185

> Background of control (TextOnAccentFillColorSelectedTextBrush) is _always_ white, force foreground to black

HO-COOH:
I figured out a way, just set `Foreground` to `Null` in the original `CopyButton` template, then check whether it is override, if it is, then we use it as the `PointerOver` and `Pressed` foreground color.

Jay-o-Way (approved):
Yay, that works!

karkarl (approved):

PR #1560, opened 2024-06-12
Updated spacing images

niels9001:
Minor fix: @sonnemaf pointed out that, according to our docs, our OK labels should be all capital.


karkarl (approved):

karkarl:
> Minor fix: @sonnemaf pointed out that, according to our docs, our OK labels should be all capital.

TIL :)

PR #1564, opened 2024-06-15
Fix: A potential bug in Win32WindowHelper.cs

Paper-Scallion:
Modify the code to avoid potential problems(in 'WinUIGallery/Helper/Win32WindowHelper.cs')

## Description
max may have been incorrectly written as min in the code

## Motivation and Context
By analyzing the code, it can be inferred that there might be a problem here.

## How Has This Been Tested?
Manual test

## Types of changes


marcelwgn (approved):
Very good catch, thank you for fixing this!

PR #1572, opened 2024-06-20
Remove unused code

### LINKED ISSUE #1569: Remove screenshot mode code
There's still screenshot mode code that we do not enable. I think it's safe to say no-one is using this? So better to remove the XAML and any corresponding code that's still left.

niels9001:
This PR removes screenshot mode and some MinimumContract checks that was UWP specific.

Fixes #1569

### DIFF
--- a/WinUIGallery/ControlPages/ComboBoxPage.xaml
+++ b/WinUIGallery/ControlPages/ComboBoxPage.xaml
@@ -1,4 +1,4 @@
-﻿<!--
+<!--
 //*********************************************************
 //
 // Copyright (c) Microsoft. All rights reserved.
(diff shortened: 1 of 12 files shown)


marcelwgn:
Always great to see code cleaned up! Out of curiosity, why was this feature added in the first place and what makes it become obsolete now? 😅

karkarl:
> Always great to see code cleaned up! Out of curiosity, why was this feature added in the first place and what makes it become obsolete now? 😅

I believe it was a WinUI2 Gallery feature, before snipping tool was at all useful?

karkarl (approved):

PR #1574, opened 2024-06-28
Updating TabView to consume tear-out APIs

llongley:
These changes update WinUI Gallery to consume the new TabView tear-out APIs in WASDK experimental 1, instead of using the WinRT drag-and-drop APIs.  This makes it so that tab tear-out immediately creates a new window that the user starts to drag, rather than waiting for a drop to create the new window.

While I was here, I also added a change to enable running WinUI Gallery as an unpackaged app, which appeared to previously not be properly functional.


karkarl:
Is it intended to merge into Scott's 1.6 Experimental1 branch?

Scottj1s (changes requested):
per Karen, target branch looks like it needs updating

llongley:
Updated to target the main branch instead.

standalone.props
```diff
@@ -1,7 +1,8 @@
 <Project>
   <PropertyGroup>
     <!-- The NuGet versions of dependencies to build against. -->
-    <WindowsAppSdkPackageVersion>1.5.240607001</WindowsAppSdkPackageVersion>
+    <WindowsSdkPackageVersion>10.0.22621.35-preview</WindowsSdkPackageVersion>
```
Scottj1s, line 4: can use non-preview 10.0.22621.42

WinUIGallery/WinUIGallery.csproj
```diff
@@ -311,5 +311,10 @@
                 <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
             </_NoneWithTargetPath>
         </ItemGroup>
-    </Target> 
+    </Target>
+    <!-- Needed in order for running unpackaged to work properly. -->
```
Scottj1s, line 315: shouldn't need this - what assets aren't being copied?

karkarl:
Fixes #1622

WinUIGallery/WinUIGallery.csproj
llongley: It looks like there are other problems that I don't immediately understand causing unpackaged running to not work anyhow, so I'll remove these.

WinUIGallery/AnimatedVisuals/LottieLogo1.cs
```diff
@@ -82,17 +82,17 @@ sealed class AnimatedVisual : IAnimatedVisual
             // Rectangle Path 1
             CompositionColorBrush ColorBrush_AlmostDarkTurquoise_FF00D1C1()
             {
-                return _c.CreateColorBrush(Windows.UI.Color.FromArgb(0xFF, 0x00, 0xD1, 0xC1));
+                return _c.CreateColorBrush(Microsoft.UI.ColorHelper.FromArgb(0xFF, 0x00, 0xD1, 0xC1));
```
llongley, line 85: While I was briefly looking at unpackaged deployment, I happened across this, which was causing Windows.UI.Xaml.dll to be loaded.  We should be using Microsoft.UI color functionality, other than the Windows.UI.Color struct that does not require loading that DLL to be used.

karkarl:
Closing this PR as the changes have been merged in with https://github.com/microsoft/WinUI-Gallery/pull/1629.

PR #1575, opened 2024-07-01
Feat: Better xaml sound demo

HO-COOH:
## Description
See screenshot.

## Motivation and Context
Previously the sound demo requires toggling the ComboBox, but the ComboBox itself will have an element sound too, causing some confusion. Also the ComboBox item ordering is incorrect (MovePrevious -> MoveNext is the correct order)

image

## How Has This Been Tested?
Manual

image

## Types of changes


karkarl:
I worry this UI makes it so that it's not scalable, for example if more System sounds are added. However, I don't foresee more to be added in the near future, so the convenience of individual buttons should be more valuable.

karkarl (approved):

PR #1591, opened 2024-07-22
Show item base classes

AndrewKeepCoding:
## Description

Shows the base classes of each control/element.

## Motivation and Context

This will help devs understand the inheritance structure of WinUI controls/elements.

## How Has This Been Tested?

Tested on VS.

image

## Types of changes

### DIFF
--- a/WinUIGallery/Controls/PageHeader.xaml
+++ b/WinUIGallery/Controls/PageHeader.xaml
@@ -23,9 +23,23 @@
             <RowDefinition Height="Auto" />
             <RowDefinition Height="Auto" />
             <RowDefinition Height="Auto" />
+            <RowDefinition Height="Auto" />
         </Grid.RowDefinitions>
-
+        <BreadcrumbBar
+            Grid.Row="0"
+            IsHitTestVisible="False"
+            Visibility="{x:Bind Item.BaseClasses, Converter={StaticResource collectionConverter}}"
+            ItemsSource="{x:Bind Item.BaseClasses}">
+            <BreadcrumbBar.ItemTemplate>
+                <DataTemplate x:DataType="x:String">
+                    <TextBlock
+                        Style="{StaticResource CaptionTextBlockStyle}"
+                        Text="{x:Bind}" />
+                </DataTemplate>
+            </BreadcrumbBar.ItemTemplate>
+        </BreadcrumbBar>
         <TextBlock
+            Grid.Row="1"
             AutomationProperties.AutomationId="PageHeader"
             Style="{StaticResource TitleTextBlockStyle}"
             Text="{x:Bind Item.Title}"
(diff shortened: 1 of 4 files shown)


karkarl:
Thanks for the PR, this is an awesome addition! 

I think the breadcrumb bar should be below the header, since it's not for navigating and purely for informative purposes. @niels9001

niels9001:
Oooh I like this.. agree with @karkarl that having it below the title makes more sense.

That being said, I think the page header is starting to become pretty info heavy.

image

I wonder if the inheritance and namespace should live under a button+flyout? It's relevant information, but also not critical information. Thoughts @AndrewKeepCoding @karkarl ?

AndrewKeepCoding:
Changed the order and here is how it looks:
image

AndrewKeepCoding:
> I wonder if the inheritance and namespace should live under a button+flyout? It's relevant information, but also not critical information. 

@niels9001 I agree. There's a lot of text around there. But IMHO, I thought it'd be better for devs, especially the new ones in WinUI, that this inheritance info just gets into their eyes.

marcelwgn:
> > I wonder if the inheritance and namespace should live under a button+flyout? It's relevant information, but also not critical information.
> 
> @niels9001 I agree. There's a lot of text around there. But IMHO, I thought it'd be better for devs, especially the new ones in WinUI, that this inheritance info just gets into their eyes.

I might be an outlier with this position, but I don't think new devs really need to have that information often. There are very few situations where one wouldn't derive from UserControl/Control (which the new control templates already do). Out of curiosity, in which situation would this information would have been helpful to you @AndrewKeepCoding? 😃

AndrewKeepCoding:
> Out of curiosity, in which situation would this information would have been helpful to you @AndrewKeepCoding? 😃

@marcelwgn
Now that I've been working with WinUI for a couple of years, I don't find this feature that helpful. But when I started working with WinUI 3, I remember that learning the inheritance of the elements helped me understand the framework better.

Might sound stupid and maybe it's just me, but for example, I remember I was assuming that `RichTextBox`, `AutoSuggestBox` and `NumberBox` derive from `TextBox`. 😅 After I learned that they are totally different controls, I tend to check out the inheritance of controls/elements when I use them by going all the way to the docs.

That's why I thought this feature would be good for devs new to WinUI. 

I totally understand your opinions guys. Let me know if you want to close this or modify it to a less texty way. 🙂

karkarl:
Perhaps it can mirror the msdn format, and be placed under the control description? 

image

AndrewKeepCoding:
It's didn't push the commit yet but looks like this:

image

karkarl:
I think that looks great! Can you push the commit?

AndrewKeepCoding:
@karkarl Done.

karkarl (approved):

PR #1629, opened 2024-10-02
 Updating TabView to consume tear-out APIs + Fix Failing Axe Tests

karkarl:
Clone of `origin/user/llongley/ConsumingTabTearOut` to debug the failing axe tests in CI.

Contains changes to https://github.com/microsoft/WinUI-Gallery/pull/1574


llongley (approved):

WinUIGallery/TabViewPages/TabViewWindowingSamplePage.xaml.cs
```diff
@@ -69,108 +63,67 @@ public void AddTabToTabs(TabViewItem tab)
...
 
-            // ... set the drag data to the tab...
-            args.Data.Properties.Add(DataIdentifier, firstItem);
+            var newPage = (TabViewWindowingSamplePage)tabTearOutWindow.Content;
 
-            // ... and indicate that we can move it
-            args.Data.RequestedOperation = DataPackageOperation.Move;
+            foreach (TabViewItem tab in args.Tabs.Cast<TabViewItem>())
+            {
+                GetParentTabView(tab)?.TabItems.Remove(tab);
```
michael-hawker, line 90: This is just the `sender` in this case

WinUIGallery/TabViewPages/TabViewWindowingSamplePage.xaml.cs
```diff
@@ -69,108 +63,67 @@ public void AddTabToTabs(TabViewItem tab)
...
 
-        private async void Tabs_TabStripDrop(object sender, DragEventArgs e)
+        private void Tabs_ExternalTornOutTabsDropping(TabView sender, TabViewExternalTornOutTabsDroppingEventArgs args)
         {
-            // This event is called when we're dragging between different TabViews
-            // It is responsible for handling the drop of the item into the second TabView
+            args.AllowDrop = true;
+        }
+
+        private void Tabs_ExternalTornOutTabsDropped(TabView sender, TabViewExternalTornOutTabsDroppedEventArgs args)
```
michael-hawker, line 100: @llongley it's a bit odd that it's a collection of UIElement when you'd only be dragging TabViewItems, also the base control doesn't support selecting multiple tabs currently.

Also, could these event args just provide the external parent TabView of the external TabViewItem so the VisualTree isn't needed here? Wouldn't it just be passing an additional reference to `this` as the event creating this is the Drag event on that TabView/Window pair?

https://github.com/microsoft/microsoft-ui-xaml/blob/f31293f4227cd3b4f7c151106528b8e1367bda11/src/controls/dev/TabView/TabView.cpp#L1902-L1935

PR #1630, opened 2024-10-10
Add schema for ControlInfoData.json

HO-COOH:
## Description
I added a schema file for the `ControlInfoData.json` to make adding new control / control group to it simpler. Now you get nice intellisense for all properties and error message for missing required ones. The schema should be automatically picked up by any reasonable editors.

## Motivation and Context
I just happens to know this.

## How Has This Been Tested?
The current file reports no mismatch for this schema.

image

image

## Types of changes

### DIFF
--- a/WinUIGallery/DataModel/ControlInfoDataSchema.json
+++ b/WinUIGallery/DataModel/ControlInfoDataSchema.json
@@ -14,7 +14,7 @@
                     },
                     "Title": {
                         "type": "string",
-                        "description": "Title appears on the control group page"
+                        "description": "The name of the control group. Will be displayed on the control group page and navigation"
                     },
                     "Subtitle": {
                         "type": "string",
@@ -30,7 +30,7 @@
                     },
                     "IconGlyph": {
                         "type": "string",
-                        "description": "Appears as an icon for the NavigationViewItem of this control group"
+                        "description": "Glyph of this control group. Will be displayed as an icon for the NavigationViewItem of this control group"
                     },
                     "ApiNamespace": {
                         "type": "string",
@@ -38,7 +38,7 @@
                     },
                     "IsSpecialSection": {
                         "type": "boolean",
-                        "description": "Special section is hard-coded in NavigationRootPage.xaml",
+                        "description": "Indicates that this section is not a regular control section but is added to the navigation differently",
                         "default": true
                     },
                     "Folder": {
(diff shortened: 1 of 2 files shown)


WinUIGallery/DataModel/ControlInfoDataSchema.json
```diff
@@ -0,0 +1,132 @@
...
+            "items": {
+                "type": "object",
+                "properties": {
+                    "UniqueId": {
+                        "type": "string",
+                        "description": "Globally unique id of this group",
+                        "uniqueItems": true
+                    },
+                    "Title": {
+                        "type": "string",
```
marcelwgn, line 16: I would probably write something like "The name of the control group. Will be displayed on the control group page and navigation".

Same for the other descriptions, I think ideally we should tell the intention of the property and then tell where it is being used.

WinUIGallery/DataModel/ControlInfoDataSchema.json
```diff
@@ -0,0 +1,132 @@
...
+                        "description": "Appears as an icon for the NavigationViewItem of this control group"
+                    },
+                    "ApiNamespace": {
+                        "type": "string",
+                        "description": "ApiNamespace of a control group is not used. Can leave empty"
+                    },
+                    "IsSpecialSection": {
+                        "type": "boolean",
+                        "description": "Special section is hard-coded in NavigationRootPage.xaml",
```
marcelwgn, line 41:
```suggestion
                        "description": "Indicates that this section is not a regular control section but is added to the navigation differently",
```

WinUIGallery/DataModel/ControlInfoDataSchema.json
```diff
@@ -0,0 +1,132 @@
...
+                                        },
+                                        "required": [ "Title", "Uri" ]
+                                    },
+                                    "description": "Add a link to the doc in the Documentation dropdown on the detail page of this control"
+                                },
+                                "RelatedControls": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "string",
+                                        "description": "Name of the related control"
```
marcelwgn, line 119:
```suggestion
                                        "description": "ID of the related control"
```

marcelwgn (changes requested):
Love the idea of adding a schema for this file. One thing about the descriptions, I think ideally we want to first name the intention of the property and then explain where it is used (if it is necessary to do that at all)

HO-COOH:
@marcelwgn Did some changes to it, please review

marcelwgn (approved):
Looks good to me, thank you for these changes!

PR #1641, opened 2024-11-05
Copy "git clone https..." to clipboard when clicked

AndrewKeepCoding:
## Description

Copy "git clone https://github.com/microsoft/WinUI-Gallery" to clipboard when clicked.

## Motivation and Context

This way is easier and less error prone.

## How Has This Been Tested?

Run the app and confirmed that the text is copied to the clipboard.

image

## Types of changes

### DIFF
--- a/WinUIGallery/SettingsPage.xaml
+++ b/WinUIGallery/SettingsPage.xaml
@@ -118,6 +118,7 @@
                             IsClickEnabled="True">
                             <TextBlock
                                 x:Name="gitCloneTextBlock"
+                                FontFamily="Consolas"
                                 Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                                 IsTextSelectionEnabled="True"
                                 Text="git clone https://github.com/microsoft/WinUI-Gallery" />
--- a/WinUIGallery/SettingsPage.xaml
+++ b/WinUIGallery/SettingsPage.xaml
@@ -111,11 +111,19 @@
                         IsTextSelectionEnabled="True"
                         Text="{x:Bind Version}" />
                     <toolkit:SettingsExpander.Items>
-                        <toolkit:SettingsCard Header="To clone this repository">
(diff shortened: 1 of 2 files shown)


AndrewKeepCoding:
To keep consistent alignment with the bug request icon in the `SettingsCard` below, I didn't use the `CopyButton` control here and used a `FontIcon` as `SettingsCard.ActionIcon` thus there's no fancy copy animation the `CopyButton` control has.

WinUIGallery/SettingsPage.xaml
```diff
@@ -111,11 +111,19 @@
...
                         Text="{x:Bind Version}" />
                     <toolkit:SettingsExpander.Items>
-                        <toolkit:SettingsCard Header="To clone this repository">
+                        <toolkit:SettingsCard
+                            x:Name="toCloneRepoCard"
+                            Click="toCloneRepoCard_Click"
+                            Header="To clone this repository"
+                            IsClickEnabled="True">
                             <TextBlock
+                                x:Name="gitCloneTextBlock"
```
marcelwgn, line 120: I think we should probably use a monospaced font and a separate background for this text (though this text was already like this before this PR)

AndrewKeepCoding: I changed the font to `Consolas` just like the other places (Iconography/Geometry) that has a copy button.

image

AndrewKeepCoding: > a separate background for this text

I'm not sure what kind of background you mean. 🤔
Is there any section in the app that I can use as a reference?

marcelwgn: Huh interesting, I thought that we had a specific background for source code but apparently we no longer have that.
Thank you for using Consolas btw!

marcelwgn (approved):

PR #1643, opened 2024-11-08
[DRAFT] Add ComputeSharp animations to page transitions and the ContentDialog page

getrou:
- Add ComputeSharp as a nuget dependency
  - Bump WinAppSDK version to 1.6
- Add Shaders folder
  - Add RippleFade and TwirlDismiss shader effects
  - Add PixelShaderRenderer, ShaderPanel, and OverlayPanel to enable shaders and overlays in WinUI
  - Add CaptureHelper to facilitate taking captures of UI


nuget.config
```diff
@@ -12,6 +12,7 @@
 
     <!-- a local directory to allow testing nupkg files without pushing to a remote feed -->
     <add key="packagestore" value="packagestore" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
getrou, line 15: Going to need someone with permissions to add the ComputeSharp nuget packages to an approved feed.

WinUIGallery/WinUIGallery.csproj
```diff
@@ -115,14 +118,14 @@
     </ItemGroup>
 
     <ItemGroup Condition="'$(IsInWinUIRepo)' != 'true'">
-        <PackageReference Update="Microsoft.WindowsAppSDK" Version="$(WindowsAppSdkPackageVersion)" />
+        <PackageReference Update="Microsoft.WindowsAppSDK" Version="1.6.240923002" />
```
ionvasilian, line 121: Why do we need the bump to 1.6?

getrou: ComputeSharp has a dependency on 1.6

WinUIGallery/Helper/WindowHelper.cs
```diff
@@ -31,7 +31,7 @@ static public Window CreateWindow()
         {
             Window newWindow = new Window
             {
-                SystemBackdrop = new MicaBackdrop()
+                //SystemBackdrop = new MicaBackdrop()
```
jeffstall, line 34: Why?

getrou: The capture API doesn't capture Mica, so there was a visual glitch when it swapped in the capture to run the effect.

PR #1661, opened 2024-11-21
**feat**: Add XAML Fundamentals section with pages and examples

Zakariathr22:
## Description
This pull request introduces the "XAML Fundamentals" section, including pages and examples for Resources, ResourceDictionary, Theme Resources, XAML Styles, Binding, and Templates. It adds relevant sample code files and assets for theme resource examples. Additionally, navigation is updated to include these new sections and ensure proper handling of navigation items, main changes are:
- `Added` sample code files for Resources, ResourceDictionary, Theme Resources, XAML Styles, Binding, and Templates.
- `Added` new assets for theme resource examples and updated control images.
- `Updated` `ControlInfoData.json` to include the new "XAML Fundamentals" section.
- `Added pages for Resources, ResourceDictionary, Theme Resources, XAML Styles, Binding, and Templates, each with practical examples.
- `Updated` navigation to include the new section and handle its items programmatically.

## Motivation and Context
I noticed that many learners, especially those new to WinUI 3, often turn to the WinUI 3 Gallery application as their starting point. While the app provides excellent explanations of controls and code samples, it sometimes introduces concepts that may be unfamiliar to beginners.

To address this, I had created a dedicated section in the application called XAML Fundamentals. This section is organized into key topics, including: Resources, ResourceDictionary, Theme Resources, XAML Styles, Binding, and Templates.

The goal is to help learners better understand these foundational concepts by providing clear explanations, practical examples, and links to relevant documentation.

## How Has This Been Tested?
This code was tested **manually**, and no issues were encountered during the testing. The changes are isolated to the new section and do not affect other areas of the app.

**Resouces Page**
image

**ResouceDictionary**
image

**Theme resources**
image

**XAML Styles**
image

**Binding**
image

**Templates**
image

## Types of changes


Zakariathr22:
> Nit: note the unnecessary use of Capital letters for a number of Words.

Adjusted as suggested.

Jay-o-Way:
In the XAML, I see a couple of different values, based on 5 or 10. This doesn't fit within the general recommendation to use multiples of four. Is there a reason for this? Also, something like a **CornerRadius** can be set with `ControlCornerRadius`, right?
Examples:
```xaml
CornerRadius="5"
Padding="10"
Margin="5"
```

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
```diff
@@ -0,0 +1,394 @@
...
+                        <Run Text=": the data you are binding to, like a property in a class, another control, or a static resource." />
+                    </Paragraph>
+                    <Paragraph>
+                        <Run Text="3. " />
+                        <Bold>Binding modes</Bold>
+                        <Run Text=":" />
+                    </Paragraph>
+                    <Paragraph>
+                        <Run Text="   - " />
+                        <Bold>OneWay</Bold>
```
Jay-o-Way, line 51: I think it's valuable to mention that the **default** modes for `Binding` and `x:Bind` are different (OneWay and OneTime, resp.) A developer converting Binding to x:Bind might have unwanted results.

WinUIGallery/ControlPages/XamlFundamentals/ResourceDictionaryPage.xaml
```diff
@@ -0,0 +1,163 @@
...
+        <ResourceDictionary>
+            <x:Double x:Key="HeaderFontSize">24</x:Double>
+            <Color x:Key="AccentColor">#4CAF50</Color>
+            <SolidColorBrush x:Key="AccentBrush"
+                             Color="{StaticResource AccentColor}" />
+            <x:String x:Key="AppTitle">WinUI 3 Gallery</x:String>
+
+            <x:String x:Key="WelcomeMessage">Welcome to this page</x:String>
+            <SolidColorBrush x:Key="PageBackgroundBrush"
+                             Color="#E8EAF6" />
```
Jay-o-Way, line 20: Be very, very careful to hardcode colors like this. Did you check readability in light and dark mode? Is there no resource to use here?

WinUIGallery/ControlPages/XamlFundamentals/ThemeResourcesPage.xaml
```diff
@@ -0,0 +1,149 @@
...
+                         IsOpen="True"
+                         Severity="Warning"
+                         IsClosable="False"
+                         Message="Toggle the theme using the theme switch button at the top right corner." />
+            </local:ControlExample.Options>
+        </local:ControlExample>
+        <InfoBar IsOpen="True"
+                 IsClosable="False"
+                 Title="Explore built-in theme resources"
+                 Message="WinUI 3 includes built-in theme resources for commonly used colors, such as `CardBackgroundFillColorDefaultBrush` and `TextFillColorPrimaryBrush`. For a deeper understanding, visit the `Design guidance → Color` page in this app."
```
Jay-o-Way, line 98:
```suggestion
                 Message="WinUI 3 includes built-in theme resources for commonly used colors, such as `CardBackgroundFillColorDefaultBrush` and `TextFillColorPrimaryBrush`. For a deeper understanding, visit the `Design guidance > Color` page in this app."
```
`>` is used for navigation

Jay-o-Way (changes requested):
My thought: printing **elements** and **properties** in bold (or _values_ in italic) is good practice in documentation, but might not be needed in this app. If this formatting would be removed, it would greatly simplify the XAML. Third opinion (from core developers) is welcome, though.

Jay-o-Way:
Did you know?

> [!NOTE]
> Starting in Windows 10, version 1607, the XAML framework provides a built in Boolean to Visibility converter. The converter maps **true** to the **Visible** enumeration value and **false** to **Collapsed** so you can bind a Visibility property to a Boolean without creating a converter. Note that this is not a feature of function binding, only property binding. To use the built in converter, your app's minimum target SDK version must be 14393 or later. You can't use it when your app targets earlier versions of Windows 10. For more info about target versions, see [Version adaptive code](https://learn.microsoft.com/windows/uwp/debug-test-perf/version-adaptive-code).

https://learn.microsoft.com/windows/uwp/xaml-platform/x-bind-markup-extension#remarks

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
```diff
@@ -0,0 +1,394 @@
...
+                            BorderBrush="{ThemeResource DividerStrokeColorDefaultBrush}"
+                            Background="{ThemeResource LayerFillColorAltBrush}">
+                        <TextBlock Text="What it is"
+                                   FontWeight="Bold"
+                                   TextAlignment="Center"
+                                   VerticalAlignment="Center"
+                                   Padding="10"
+                                   TextWrapping="Wrap" />
+                    </Border>
+                    <Border Grid.Row="1"
```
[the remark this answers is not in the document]
Zakariathr22: Lol, you're absolutely right—I totally missed that! Thanks for pointing it out; I'll make sure to apply that approach! :)

WinUIGallery/ControlPages/XamlFundamentals/ResourceDictionaryPage.xaml
Zakariathr22: My goal was to demonstrate how to create resources and resource dictionaries in different ways. Using built-in resources wouldn't make sense in this context, as the purpose is to showcase the process of defining new resources rather than reusing existing ones.
However, if you see a better approach, your input is always welcome! :)

Jay-o-Way: Of course, makes complete sense, just needed a second look at this!

WinUIGallery/ControlPages/XamlFundamentals/ResourcesPage.xaml
```diff
@@ -0,0 +1,162 @@
...
+                    </Paragraph>
+                    <Paragraph>
+                        <Run Text="2. Page-level: "
+                             FontWeight="Bold" />
+                        <Run Text="define resources that are specific to a particular page." />
+                    </Paragraph>
+                    <Paragraph>
+                        <Run Text="3. Control-level: "
+                             FontWeight="Bold" />
+                        <Run Text="define resources local to a specific control, such as a button or grid." />
```
Jay-o-Way, line 44: Semantics: Are these examples controls or elements?

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
```diff
@@ -0,0 +1,466 @@
...
+        </core:ControlExample>
+        
+        <core:ControlExample HeaderText="Using `Converter` in `x:Bind`"
+                             XamlSource="Binding\BindingSample3_xaml.txt"
+                             CSharpSource="Binding\BindingSample3_csharp.txt">
+            <core:ControlExample.Example>
+                <StackPanel Spacing="8">
+                    <DatePicker x:Name="DatePickerControl"
+                                Header="Select a date" />
+                    <TextBlock Text="{x:Bind FormatDate(DatePickerControl.SelectedDate), Mode=OneWay}" />
```
Jay-o-Way, line 347: Technically, this is a **function**, am I right? I don't see a `.. Converter="SomeConverter"`
https://learn.microsoft.com/windows/uwp/data-binding/function-bindings

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
```diff
@@ -0,0 +1,466 @@
...
+            </core:ControlExample.Example>
+            <core:ControlExample.Xaml>
+                <x:String xml:space="preserve">
+&lt;Slider x:Name="WidthSlider" Minimum="50" Maximum="300" Value="100" Width="300" /&gt;
+&lt;Rectangle Width="{x:Bind WidthSlider.Value, Mode=OneWay}" Height="50" Fill="#088" /&gt;
+        </x:String>
+            </core:ControlExample.Xaml>
+        </core:ControlExample>
+        
+        <core:ControlExample HeaderText="Using `Converter` in `x:Bind`"
```
Jay-o-Way, line 340:
```suggestion
        <core:ControlExample HeaderText="Using `Function` in `x:Bind`"
```

Jay-o-Way (changes requested):
Getting better :) just a couple more things...

- Let's remove those localizers like `en-us/' or `en-gb/` in URLs.

WinUIGallery/ControlPages/XamlFundamentals/ResourcesPage.xaml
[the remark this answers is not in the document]
Zakariathr22: Both belong to the `Microsoft.UI.Xaml.Controls` namespace, so I referred to them as controls.
References: [Button Class](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.button) , [Grid Class](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.grid)

WinUIGallery/ControlPages/XamlFundamentals/ResourceDictionaryPage.xaml
```diff
@@ -0,0 +1,163 @@
...
+                    <Paragraph>
+                        <Run Text="Where can you create a resource dictionary?"
+                             FontWeight="Bold" />
+                    </Paragraph>
+                    <Paragraph>
+                        <Run Text="1. "
+                             FontWeight="Bold" />
+                        <Run Text="Inline (app-level): "
+                             FontWeight="Bold" />
+                        <Run Text="define it in " />
```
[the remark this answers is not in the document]
Zakariathr22: Interesting, I'll fix this to avoid extra spaces.

Jay-o-Way (changes requested):
Few more things.
Still think both the _amount_ of bold text - and the _formatting_ - can be significantly reduced, but again, opinion from a developer is welcome here!
- Besides: should **bold** or **semibold** be used? (the latter is used on other pages, and also in the [Type Ramp](winui3gallery://item/Typography))
- Example: `<Paragraph FontWeight="SemiBold">Text</>` is also valid (omits the `Run`)
- Check consistency in the lists: sometimes the number is bold, sometimes it's not; sometimes there's a dash, sometimes a bullet...

Zakariathr22:
> Few more things. Still think both the _amount_ of bold text - and the _formatting_ - can be significantly reduced, but again, opinion from a developer is welcome here!
> 
> * Besides: **semibold** must be used. This is used on all other pages, and also in the Type Ramp. (Bold is a bit hard on the eyes)
> * Example: `<Paragraph FontWeight="SemiBold">Text</>` is also valid (omits the `Run`)
> * Check consistency in the lists: sometimes the number is bold, sometimes it's not; sometimes there's a dash, sometimes a bullet...

I used bold to highlight the important points, it seems a good idea if we replace it with semibold.
I agree with the other points you mentioned.

I'll work on it, thank you for your effort :)

WinUIGallery/ControlPages/XamlFundamentals/ResourcesPage.xaml
```diff
@@ -0,0 +1,162 @@
...
+
+    <StackPanel>
+        <Expander HorizontalAlignment="Stretch"
+                  HorizontalContentAlignment="Left"
+                  Margin="0 12 0 0">
+            <Expander.Header>
+                <TextBlock Text="Creating and using XAML resources in WinUI 3"
+                           Margin="0 24" />
+            </Expander.Header>
+            <StackPanel>
```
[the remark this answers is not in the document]
Zakariathr22: Actually, elements in other UserControls used in these pages like `ControlExample` uses individual Margins, so I followed the same approach and used margins for spaces.
**Screenshot from `ControlExample.xaml`**
image

Jay-o-Way:
In order to not spam the PR with suggestions, I thought I'd give you some of my ideas via these files, for example:
[ThemeResourcesPage.zip](https://github.com/user-attachments/files/17970605/ThemeResourcesPage.zip)

WinUIGallery/ControlPages/XamlFundamentals/ResourcesPage.xaml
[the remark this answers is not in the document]
Jay-o-Way: I see. Well spotted. I also know why, in this case, the Margin must be used: it is related to Visibility of a control. See, when a control is Collapsed, the surrounding Spacing(s) on either side would "stay" and still add up. However, Visibility is not an issue in the case of the text in this PR :)

Zakariathr22:
> In order to not spam the PR with suggestions, I thought I'd give you some of my ideas via these files, for example: [ThemeResourcesPage.zip](https://github.com/user-attachments/files/17970605/ThemeResourcesPage.zip)

I’ve granted you write access.
The remaining changes are minor, so feel free to apply them directly if you prefer.

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml.cs
Jay-o-Way, line 1: Good catch. I think a `switch` on the `Tag` would be even better :)

Jay-o-Way:
I think we should remove or change the "Binding to resources" example, because this is simply a Static/Theme resource, covered in the other page. And "Binding to a property in code-behind" is not using any C# at all, it is actually binding properties between two controls.

niels9001:
@zakariathr22 Thanks for the PR. Great to see the community putting in the effort to make onboarding to WinUI easier :-)!

Some high-level thoughts I'd love to get input on before I do a more detailed review:

1. Can we add the documentation inline vs. an expander? I think the Expander hides the information and you might miss it. Also the expanders kind of suggest that there's code (as they are prominently used in each code sample)
image


2. Would it make sense to combine these 3 pages? All of this content is very much related, and as a new developer I'd rather have a single page that introduces these concepts vs 3 separate pages? I might read the first one and then miss the others.

image

3. We need to make sure our samples meet a11y requirements.
image

Jay-o-Way:
Hey Niels, thanks for your message. My quick thoughts:
1. Sure, it's a possibility. We also used this on the Iconography page, and the fact that somebody chose this design too now, does mean something, I think.
2. Difficult. Theme resources and Styles are also related. Page(s) would become very long, especially considering point 1
3. The font color is always white on my end. Let's se if we should update Zakaria's repo... ⌛

Zakariathr22:
1. I used the expander to keep explanations brief and focus more on examples and code. It also worked well on the icons page.  

2. I might combine "Resources" and "ResourceDictionary" into one page, but I prefer keeping "Theme Resources" separate since it's an important topic.  

3. @Jay-o-Way Thanks for fixing that!

niels9001:
**Resources**
Do we need 3 samples here? It's pretty verbose for only small differences.


`xml
 
 <Application ...>
    <Application.Resources>
        <Color x:Key="AppLevelColor">#FF5733</Color>
    </Application.Resources>
</Application>


<Page ...>
    <Page.Resources> 
         <x:String x:Key="WelcomeString">Welcome to "WinUI 3 Gallery!</x:String>
    </Page.Resources>

    <Grid>
      <Grid.Resources> 
           <x:Double x:Key="ButtonFontSize">24</x:Double>
       </Grid.Resources>
           
        <Button FontSize="{StaticResource ButtonFontSize}" Content="{StaticResource WelcomeString}" Foreground="{StaticResource AppLevelColor}" />
        
        
    </Grid>
</Page>

niels9001:
> **Resources** Do we need 3 samples here? It's pretty verbose for only small differences.
> 
> Something like this would show the same thing, right?
> 
> ```
>  
> <Application ...>
>     <Application.Resources>
>         <Color x:Key="AppLevelColor">#FF5733</Color>
>     </Application.Resources>
> </Application>
> 
> 
> <Page ...>
>     <Page.Resources> 
>         <x:String x:Key="WelcomeString">Welcome to "WinUI 3 Gallery!</x:String>
>     </Page.Resources>
> 
>     <Grid>
>         <Grid.Resources> 
>             <x:Double x:Key="ButtonFontSize">24</x:Double>
>      </Grid.Resources>
>            
>         <Button FontSize="{StaticResource ButtonFontSize}" Content="{StaticResource WelcomeString}" Foreground="{StaticResource AppLevelColor}" />
>         
>     </Grid>
> </Page>
> ```

Actually.. would a single resources page like this work:

Structure:

**Resources**
Explanation on resources
Single resource sample

**ResourceDictionary**
Explanation on resources
Single ResourceDictionary sample

**ThemeResources**
Explanation on staticresource vs. themeresource
Single ThemeResource sample with colors, string and image?


That would keep everything on a single page.. and as a user you have one stop for all of your resourcing needs?

Jay-o-Way:
I agree with combining the Resources + Dictionary with that simpler example. About the Expander - both is fine for me.

Zakariathr22:
> Actually.. would a single resources page like this work:
> 
> Structure:
> 
> **Resources** Explanation on resources Single resource sample
> 
> **ResourceDictionary** Explanation on resources Single ResourceDictionary sample
> 
> **ThemeResources** Explanation on staticresource vs. themeresource Single ThemeResource sample with colors, string and image?
> 
> That would keep everything on a single page.. and as a user you have one stop for all of your resourcing needs?

How about organizing it like this:  

- **Expander**: Highlight the key points  
- Single **Resources** example  
- Single **ResourceDictionary** example  
- **StaticResource vs. ThemeResource** example  
- Additional **ThemeResource** example with colors, strings, and images

niels9001:
I'd maybe remove the Expanders on these pages. https://github.com/microsoft/WinUI-Gallery/pull/1679 will declutter the header a bit and give back some vertical space.

I think the info in those expanders are pretty important looking at the goal of the page? I wouldn't hide the information then. Also, it's not too much information.

To @jay-o-way 's point on using the Expander on the Icons page; it's not really relevant information, and this page more of a tool.. scrolling every time is annoying there as developers might go to this page a couple of times a day. I agree it's not a perfect solution, but it does serve a different purpose.

Zakariathr22:
> I'd maybe remove the Expanders on these pages.
> 
> I think the info in those expanders are pretty important looking at the goal of the page? I wouldn't hide the information then. Also, it's not too much information.

I will work on it

Zakariathr22:
@niels9001 any other suggestions?

@Jay-o-Way, your suggestions and adjustments are always appreciated!

Jay-o-Way:
Ah yes, I don't mind this change. All I think now is that the three "headers" are very redundant. "Creating and using (...) in WinUI 3" is more than a user needs.
Now, go have a nice weekend @Zakariathr22 it's late! 😉

WinUIGallery/Assets/ControlImages/CodeTagIcon.png
niels9001, line 1: We don't need an icon for this as this is not a control?

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
```diff
@@ -0,0 +1,324 @@
...
+                <Paragraph>1. Target: the property of a control that you want to bind to (e.g. Text, Background, Visibility).</Paragraph>
+                <Paragraph>2. Source: the data you are binding to, like a property in a class, another control, or a static resource.</Paragraph>
+                <Paragraph>3. Binding modes:</Paragraph>
+                <Paragraph xml:space="preserve">    - OneWay updates the target when the source changes.</Paragraph>
+                <Paragraph xml:space="preserve">    - TwoWay updates both the target and the source.</Paragraph>
+                <Paragraph xml:space="preserve">    - OneTime sets the target once and doesn't update afterward.</Paragraph>
+            </RichTextBlock>
+
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">{x:Bind} and {Binding} feature comparison</Paragraph>
```
niels9001, line 26: I'd say this table is maybe something we should keep in the docs only? We should focus on samples for WinUI Gallery, and the link you have to MS Learn might already be sufficient?

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
```diff
@@ -0,0 +1,171 @@
...
+        <ItemsPanelTemplate x:Key="StackPanelTemplate">
+            <StackPanel Orientation="Vertical" />
+        </ItemsPanelTemplate>
+    </Page.Resources>
+
+    <StackPanel>
+        <StackPanel Spacing="12"
+                    Margin="0,12,0,0">
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">Where can you create templates?</Paragraph>
```
niels9001, line 25: Maybe leave out these questions, it's not inline with how the docs are written.

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
```diff
@@ -0,0 +1,171 @@
...
+                <Paragraph>You can define templates at the app, page, or control level, just like styles and resources. The placement depends on where and how you want the template to be reused.</Paragraph>
```
niels9001, line 26:
```suggestion
                <Paragraph>Templates can be defined at the app, page, or control level, similar to styles and resources. The placement is determined by the intended scope and reuse of the template.</Paragraph>
```

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
```diff
@@ -0,0 +1,171 @@
...
+            </RichTextBlock>
+
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">Key points:</Paragraph>
+                <Paragraph>- Templates allow you to redefine how a control is built visually while keeping its behavior intact.</Paragraph>
+                <Paragraph>- Templates can be defined at the App, Page, or Control level, depending on the desired scope.</Paragraph>
```
niels9001, line 32: Maybe combine this with the text above so it's just a single paragraph of text.

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
```diff
@@ -0,0 +1,171 @@
...
+                <Paragraph>
+                    <Run Text="    -" />
+                    <Run FontWeight="SemiBold"
+                         Text="ControlTemplate" />
+                    <Run Text="customizes the structure of a control." />
+                </Paragraph>
+                <Paragraph>
+                    <Run Text="    -" />
+                    <Run FontWeight="SemiBold"
+                         Text="ItemsPanelTemplate" />
```
niels9001, line 43: I'm wondering if we should include `ItemsPanelTemplate`.. it is very specific. Thoughts?

WinUIGallery/ControlPages/XamlFundamentals/XamlResourcesPage.xaml
```diff
@@ -0,0 +1,247 @@
...
+    </Page.Resources>
+
+    <StackPanel>
+        <TextBlock Text="Creating and using XAML resources"
+                   Style="{ThemeResource SubtitleTextBlockStyle}"
+                   Margin="0 12" />
+        <StackPanel Spacing="12">
+            <RichTextBlock>
+                <Paragraph>
+                    <Run FontWeight="SemiBold" Text="Where can you create XAML resources?" />
```
niels9001, line 35: See earlier comments, let's avoid these questions

WinUIGallery/ControlPages/XamlFundamentals/XamlResourcesPage.xaml
```diff
@@ -0,0 +1,247 @@
...
+                </Paragraph>
+                <Paragraph>1. App-level: define resources globally, accessible throughout the application.</Paragraph>
+                <Paragraph>2. Page-level: define resources that are specific to a particular page.</Paragraph>
+                <Paragraph>3. Control-level: define resources local to a specific control, such as a button or grid.</Paragraph>
+            </RichTextBlock>
+
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">How to create and use XAML resources?</Paragraph>
+                <Paragraph>
+                    <Run Text="You define XAML Resources using the" />
```
niels9001, line 45: Try to avoid "You", instead e.g.. "Xaml Rresources xan be defined by using the..".

Copilot can help with rewriting :)!

WinUIGallery/ControlPages/XamlFundamentals/XamlStylesPage.xaml.cs
```diff
@@ -0,0 +1,37 @@
...
+        public XamlStylesPage()
+        {
+            this.InitializeComponent();
+
+            ApplyButtonStyle(MyButton);
+        }
+
+        private static void ApplyButtonStyle(Button myButton)
+        {
+            var buttonStyle = new Style(typeof(Button));
```
niels9001, line 29: I'd maybe focus here on getting a XAML defined style in code-behind as well? That feels like a use-case that is more often used?

WinUIGallery/DataModel/ControlInfoData.json
```diff
@@ -178,6 +178,128 @@
         }
       ]
     },
+    {
+      "UniqueId": "XamlFundamentalsItem",
```
niels9001, line 182: There's no need to add this to `ControlInfoData.json` right? The design guidance, a11y and XAML fundamentals sections are hardcoded on the navview page itself?

WinUIGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -272,6 +272,34 @@
...
 
+                <NavigationViewItem x:Name="XamlFundamentalsItem" AutomationProperties.AutomationId="XamlFundamentalsItem" Content="XAML fundamentals" Tag="XamlFundamentals">
+                    <NavigationViewItem.Icon>
+                        <FontIcon Glyph="&#xE943;" />
+                    </NavigationViewItem.Icon>
+                    <NavigationViewItem.MenuItems>
+                        <NavigationViewItem
+                                    x:Name="XamlResourcesPage"
+                                    AutomationProperties.AutomationId="XamlResources"
+                                    Content="XAML resources"
```
niels9001, line 283:
```suggestion
                                    Content="Resources"
```

Since this is part of 'XAML fundamentals", we could leave this out?

WinUIGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -272,6 +272,34 @@
...
+                                    Tag="XamlResources" />
+                        <NavigationViewItem
+                                    x:Name="XamlStylesPage"
+                                    AutomationProperties.AutomationId="XamlStyles"
+                                    Content="XAML styles"
```
niels9001, line 288:
```suggestion
                                    Content="Styles"
```

Since this is part of 'XAML fundamentals", we could leave this out?

niels9001 (changes requested):

WinUIGallery/Assets/ControlImages/CodeTagIcon.png
Zakariathr22: I believe this icon is necessary for two key reasons:  
1. It will be displayed when users search for topics related to XAML fundamentals.  
image
2. It helps avoid using unrelated control icons, as seen in the "Accessibility" section (where the AppBarSeparator icon is used), or reusing the same icon as the "Design guidance" sections and "Pivot" control, which could lead to confusion.  
image

To address this, I created a new icon specifically for this purpose.
```svg
<?xml version="1.0" ?>
<svg fill="none" height="72" viewBox="0 0 24 24" width="72"
    xmlns="http://www.w3.org/2000/svg">
    <defs>
        <linearGradient id="greenGradient" x1="0" y1="0" x2="1" y2="1">
            <stop offset="0%" stop-color="#00FF00" />
            <stop offset="100%" stop-color="#006400" />
        </linearGradient>
    </defs>
    <g fill="url(#greenGradient)" opacity="0.8">
        <path d="M8.08553 18.6112L14.0817 4.60666C14.2991 4.09895 14.8868 3.8636 15.3946 4.08097C15.866 4.28283 16.1026 4.80407 15.96 5.284L15.9202 5.39385L9.92409 19.3984C9.70671 19.9061 9.11892 20.1414 8.61121 19.9241C8.13977 19.7222 7.90316 19.201 8.04581 18.721L8.08553 18.6112L14.0817 4.60666L8.08553 18.6112ZM2.29289 11.2931L6.29289 7.29315C6.68342 6.90263 7.31658 6.90263 7.70711 7.29315C8.06759 7.65363 8.09532 8.22087 7.7903 8.61316L7.70711 8.70736L4.41421 12.0003L7.70711 15.2931C8.09763 15.6837 8.09763 16.3168 7.70711 16.7074C7.34662 17.0678 6.77939 17.0956 6.3871 16.7906L6.29289 16.7074L2.29289 12.7074C1.93241 12.3469 1.90468 11.7796 2.2097 11.3874L2.29289 11.2931L6.29289 7.29315L2.29289 11.2931ZM16.2921 7.29191C16.6526 6.93144 17.2198 6.90374 17.6121 7.20878L17.7063 7.29198L21.7071 11.2932C22.0678 11.6538 22.0953 12.2214 21.7899 12.6136L21.7066 12.7078L17.7058 16.7034C17.315 17.0936 16.6818 17.0932 16.2916 16.7024C15.9313 16.3417 15.904 15.7745 16.2093 15.3824L16.2925 15.2882L19.5854 11.9998L16.292 8.70613C15.9015 8.31558 15.9015 7.68242 16.2921 7.29191Z" />
    </g>
</svg>
```

WinUIGallery/DataModel/ControlInfoData.json
Zakariathr22: I simply followed the same approach used in the "Design Guidance" and "Accessibility" sections.

image
---
image

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
Zakariathr22: I noticed that `ItemsPanelTemplate` is used frequently in various control examples throughout the app, such as:  
- FlipView Page: 3rd example  
- GridView Page: 2nd example  
- ListView Page: 3rd and 4th examples  
- SplitButton Page: 1st example
- and it is utilized numerous times within the app itself

Given its repeated use in the app, I decided to include it.

WinUIGallery/DataModel/ControlInfoData.json
niels9001: Ah, my bad. Thought these were not part of the JSON. We probably want to add a description as well? We can do that in a separate PR if needed

Zakariathr22: I agree, I will open a PR after finishing this

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
Zakariathr22: I separated the text blocks to improve spacing.

WinUIGallery/Navigation/NavigationRootPage.xaml
```diff
@@ -272,6 +272,34 @@
                     </NavigationViewItem.MenuItems>
                 </NavigationViewItem>
 
+                <NavigationViewItem x:Name="XamlFundamentalsItem" AutomationProperties.AutomationId="XamlFundamentalsItem" Content="XAML fundamentals" Tag="XamlFundamentals">
+                    <NavigationViewItem.Icon>
+                        <FontIcon Glyph="&#xE943;" />
```
niels9001, line 277:
```suggestion
                        <PathIcon Data="M8.06562 18.9434L14.5656 4.44339C14.7351 4.06542 15.1788 3.89637 15.5568 4.0658C15.9033 4.22112 16.0742 4.60695 15.9698 4.96131L15.9344 5.05698L9.43438 19.557C9.26495 19.935 8.82118 20.104 8.44321 19.9346C8.09673 19.7793 7.92581 19.3934 8.03024 19.0391L8.06562 18.9434L14.5656 4.44339L8.06562 18.9434ZM2.21967 11.4699L6.46967 7.21986C6.76256 6.92696 7.23744 6.92696 7.53033 7.21986C7.7966 7.48612 7.8208 7.90279 7.60295 8.1964L7.53033 8.28052L3.81066 12.0002L7.53033 15.7199C7.82322 16.0127 7.82322 16.4876 7.53033 16.7805C7.26406 17.0468 6.8474 17.071 6.55379 16.8531L6.46967 16.7805L2.21967 12.5305C1.9534 12.2642 1.9292 11.8476 2.14705 11.554L2.21967 11.4699L6.46967 7.21986L2.21967 11.4699ZM16.4697 7.21986C16.7359 6.95359 17.1526 6.92938 17.4462 7.14724L17.5303 7.21986L21.7803 11.4699C22.0466 11.7361 22.0708 12.1528 21.8529 12.4464L21.7803 12.5305L17.5303 16.7805C17.2374 17.0734 16.7626 17.0734 16.4697 16.7805C16.2034 16.5143 16.1792 16.0976 16.3971 15.804L16.4697 15.7199L20.1893 12.0002L16.4697 8.28052C16.1768 7.98762 16.1768 7.51275 16.4697 7.21986Z" />
```

This is the </> icon that is more appropriate for XAML :). It's not part of the Windows Fluent Icons set, so we'd need to use a `PathIcon`.

image

niels9001 (changes requested):

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
Zakariathr22: @niels9001
Can this be considered resolved?

WinUIGallery/Navigation/NavigationRootPage.xaml
Zakariathr22: Good addition, thank you!

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
```diff
@@ -0,0 +1,199 @@
...
+                    <Run Text="Learn more about the differences between" />
+                    <Run FontFamily="Consolas"
+                         Text="x:Bind" />
+                    <Run Text="and" />
+                    <Run FontFamily="Consolas"
+                         Text="Binding" />
+                    <Run Text="by visiting" />
+                    <Hyperlink NavigateUri="https://learn.microsoft.com/windows/apps/develop/data-binding/data-binding-in-depth#xbind-and-binding-feature-comparison">
+                        Microsoft Learn
+                    </Hyperlink> .
```
niels9001, line 36:
```suggestion
                    </Hyperlink>.
```

WinUIGallery/ControlPages/XamlFundamentals/TemplatesPage.xaml
```diff
@@ -0,0 +1,171 @@
...
+                    <TextBox
+                        Padding="8"
+                        BorderBrush="{ThemeResource AccentFillColorDefaultBrush}"
+                        Header="Enter text here"
+                        Template="{StaticResource CustomTextBoxTemplate}" />
+                </Grid>
+            </local:ControlExample.Example>
+        </local:ControlExample>
+
+        <local:ControlExample
```
niels9001, line 92: I'd change the order: ControlTemplate => ItemTemplate => LayoutTemplate

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
```diff
@@ -0,0 +1,199 @@
...
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:WinUIGallery.ControlPages"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <StackPanel>
+        <StackPanel Spacing="12"
+                    Margin="0,12,0,0">
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">Key Concepts:</Paragraph>
```
niels9001, line 15:
```suggestion
                <Paragraph FontWeight="SemiBold">Key concepts:</Paragraph>
```

WinUIGallery/ControlPages/XamlFundamentals/XamlStylesPage.xaml
```diff
@@ -0,0 +1,118 @@
...
+    <StackPanel>
+        <StackPanel Spacing="12"
+                    Margin="0,12,0,0">
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">Placement of XAML Styles:</Paragraph>
+                <Paragraph>The placement of styles is the same as other resources: app-level, page-level, control-level.</Paragraph>
+            </RichTextBlock>
+
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">Key Points:</Paragraph>
```
niels9001, line 20:
```suggestion

```

Do we need the header here? The above text is only a single line and this would fit with that?

niels9001 (changes requested):

WinUIGallery/ControlPages/XamlFundamentals/XamlStylesPage.xaml
Zakariathr22: You are absolutely right

WinUIGallery/Assets/ControlImages/CodeTagIcon.png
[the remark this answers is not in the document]
Zakariathr22: Wow, this is amazing! Thank you for this 🚀

niels9001 (approved):
LGTM! Amazing work @Zakariathr22!

Zakariathr22:
@niels9001 
Thank you for your help and reviews; they made a significant difference.

niels9001:
@Zakariathr22 We've reviewed the PR and saw some (minor) improvements we could make.

Would you mind getting the latest changes from `main` and can I PR against your branch? It's probably faster that way :)!

Zakariathr22:
> Would you mind getting the latest changes from `main` and can I PR against your branch? It's probably faster that way :)!

@niels9001 Sure, feel free to go ahead, thanks for helping out!

Zakariathr22:
@niels9001
I can also grant you a write access if you like :)

marcelwgn:
Below my thoughts on the current content:

Theme resources page:
* Also, do we need a "visual" example for resource dictionaries if it looks almost the same as the previous example?
* I think having the following XAML as example of how resource scoping works might be a bit more helpful and in the source code, we reference the same brush every time so it is clear what is happening.
Image of three nested boxes


Styling page:
* The blue styled button flashes white upon hovering (at least in dark theme)
* Do we want to show non key based styling? It feels a bit of an anti pattern to teach it since I think you would ignore the base style through that pattern
* Not sure what the last example is trying to tell since the first example already uses BasedOn. I think it would be better to explain it there and skip that example

Binding page:
* For the one way and two way binding, I would probably use two columns to be able to compare it side by side instead of that toggle.
* Maybe we should use a more interesting converter than the nullability/visibility one given that already is handled somewhat natively already?
* Should we call out some of the key differences between x:Bind and binding instead of linking to the docs outright?

AndrewKeepCoding:
IMHO, the API details button (the `i` button) is confusing in this case because **Resources**, **Styles**, **Binding** and **Templates** represent features rather than classes.

image

Jay-o-Way:
> * The blue styled button flashes white upon hovering (at least in dark theme)

Yep, the hover state is not affected...

> * Maybe we should use a more interesting converter than the nullability/visibility one given that already is handled somewhat natively already?

Good one! 

> * Should we call out some of the key differences between x:Bind and binding instead of linking to the docs outright?

We actually had that, but Niels wanted it changed https://github.com/microsoft/WinUI-Gallery/pull/1661#discussion_r1875046280

Zakariathr22:
Hi @AndrewKeepCoding, thanks for sharing your opinion,

> IMHO, the API details button (the `i` button) is confusing in this case because **Resources**, **Styles**, **Binding** and **Templates** represent features rather than classes.

- `Style` is a class in the `Microsoft.UI.Xaml` namespace. [Check it out here](https://learn.microsoft.com/en-gb/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.style?view=windows-app-sdk-1.6).
- The templates, including `ControlTemplate`, `DataTemplate`, and `ItemsPanelTemplate`, are classes in the `Microsoft.UI.Xaml.Controls` namespace. You can explore them here: [ControlTemplate](https://learn.microsoft.com/en-gb/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.controltemplate?view=windows-app-sdk-1.6), [DataTemplate](https://learn.microsoft.com/en-gb/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.datatemplate?view=windows-app-sdk-1.6), and [ItemsPanelTemplate](https://learn.microsoft.com/en-gb/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.itemspaneltemplate?view=windows-app-sdk-1.6).
- As for the others, it’s not entirely clear. We have [ResourceDictionary](https://learn.microsoft.com/en-gb/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.resourcedictionary?view=windows-app-sdk-1.6) as a class, but `Resources` and `Theme Resources` aren't. 
- The same applies to [Binding](https://learn.microsoft.com/en-gb/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.data.binding?view=windows-app-sdk-1.6), but `x:Bind` is not a class.

Do you have any suggestions? 

@Jay-o-Way, I'd like also to know your thoughts on this.

Zakariathr22:
@marcelwgn https://github.com/microsoft/WinUI-Gallery/pull/1661#issuecomment-2577405947

WinUIGallery/ControlPages/XamlFundamentals/XamlResourcesPage.xaml
```diff
@@ -0,0 +1,233 @@
...
+                <Paragraph>• Inline (page-level): Defined in a specific page to limit the scope to that page.</Paragraph>
+                <Paragraph>
+                    <Run Text="• Standalone file: Defined in a separate" />
+                    <Run FontFamily="Consolas" Text=".xaml" />
+                    <Run Text="file to organize resources and reuse them in different parts of the app." />
+                </Paragraph>
+            </RichTextBlock>
+
+            <RichTextBlock>
+                <Paragraph FontWeight="SemiBold">Reasons for using resource dictionaries:</Paragraph>
```
marcelwgn, line 114: I think most of these points are more accurate for resources. ResourceDictionaries on their own are generally only reusable, and I have not seen that very often (i.e. a ResourceDictionary used multiple times throughout one project). In my experience, most of the time RDs are created because XAML needs you to so you can define resources. I would argue we could remove the RD section given that the code sample is also very similar. @niels9001 what do you think?

marcelwgn (commented):
Thank you for making those adjustments @Zakariathr22 ! Below my thoughts:

> The blue styled button flashes white upon hovering (at least in dark theme)

I think this is addressed now?

> I think by presenting non-key-based styles, the user understands how styles are automatically applied to specific control types. This helps the user grasp the concept of TargetType and see the connection between styles and controls more clearly.

I see your point, it is good to showcase how TargetType works. I guess we can leave it in.

> Actually, the BasedOn property wasn’t implemented in the first example initially. We only added it to meet a11y requirements, as Niels requested https://github.com/microsoft/WinUI-Gallery/pull/1661#issuecomment-2523200753
> For that reason, I can modify the first example as I suggested earlier without needing to use BasedOn. Do you think that would be a good idea?

I think the A11Y issues could have been fixed without BasedOn but I think its still good practice to use based on. Can we merge the first and basedon sample to avoid this duplication maybe?

WinUIGallery/ControlPages/XamlFundamentals/XamlResourcesPage.xaml
Zakariathr22: @marcelwgn @niels9001 
I included the `ResourceDictionary` part because it's very important for defining new theme resources, as it's explained in the next part, and it helps organize resources within a theme. However, if you think explaining it is not as necessary, I can remove it and focus on the core concept of using theme resources directly.

Zakariathr22: @niels9001 Should we keep the `ResourceDictionary` part for its role in defining and organizing theme resources, or remove it?

WinUIGallery/ControlPages/XamlFundamentals/XamlResourcesPage.xaml
```diff
@@ -0,0 +1,173 @@
...
+
+                        <!-- StackPanel with control-level resources defined within its own Resources -->
+                        <StackPanel Padding="8" Margin="8">
+                            <StackPanel.Resources>
+                                <!-- Define control-level resources for font size and text -->
+                                <Color x:Key="BackgroundColor">#E2241A</Color>
+                                <x:String x:Key="Description">Using control-level resources</x:String>
+                            </StackPanel.Resources>
+                            <Grid Background="{StaticResource BackgroundColor}" Padding="8">
+                                <TextBlock Text="{StaticResource Description}" Foreground="White"/>
```
marcelwgn, line 74: Why are we using a different brush for each level instead of overwriting the same one to highlight this hierarchy?

marcelwgn (commented):
Thank you for making the changes, @Zakariathr22! Ive left a few comments. Also given that @AndrewKeepCoding PR for the file scoped namespaces was merged, can you please adjust the CS files to match this change in our project?

WinUIGallery/ControlPages/XamlFundamentals/BindingPage.xaml
Zakariathr22: > Do we need this heading?

no, we don't really need that

WinUIGallery/ControlPages/XamlFundamentals/XamlResourcesPage.xaml
Zakariathr22: We're using a different brush for each level because all the stack panels are contained within the page. Overwriting the same brush would make it difficult to distinguish between App resources and Page resources.
Instead, we could consider only illustrating the hierarchy between Page resources and Control resources for better clarity?

marcelwgn: You are right, I think we can leave this as is and figure out a better way for this later. Thank you for your help!

marcelwgn (approved):

PR #1664, opened 2024-11-23
Adding accessible context menus to TabView pages

llongley:
In order for TabView pages to be sufficiently accessible, there needs to be a way for users to rearrange tabs using a single pointer input without the need for precise manipulation.  To achieve this, I've added context menus for each TabView that allows users to move tabs left and right, as well as moving tabs to other windows in the case of the tab tear-out example.

In working on this change, I found that there's a bug in WinUI 3 where moving a TabViewItem from one TabView in one window to another in another window retains the previous XamlRoot value, which throws an exception when we try to open a context menu on that TabViewItem, since it doesn't know what its correct XamlRoot is.  I've filed a bug on that issue, but for now, to work around that, I've switched from using explicitly defined TabViewItems to using a data item source, which causes each TabView to generate its own TabViewItems.

Finally, I also found a bug in Win32WindowHelper - newWndProc and oldWndProc should not be static, because the class itself is instantiated on a per-window basis.  Having these be static can result in an EngineExecutionException as a result of these delegates being garbage collected while there's still a native reference to them.  Changing them to be instance fields fixes this issue.


niels9001:
@llongley FYI, looks like a conflict popped up

niels9001:
@llongley A few conflicts popped up because of the indention changed caused by #1700. Once those are resolved we can merge this in :)

niels9001:
@llongley FYI I fixed the conflicts and enabled Mica on the tabs demo

WinUIGallery/TabViewPages/TabViewWindowingSamplePage.xaml.cs
```diff
@@ -99,64 +129,175 @@ private void Tabs_ExternalTornOutTabsDropping(TabView sender, TabViewExternalTor
...
+    private static void MoveDataItems(ObservableCollection<TabItemData> source, ObservableCollection<TabItemData> destination, object[] dataItems, int index)
+    {
+        foreach (object tabItemData in dataItems)
         {
-            if (current is TabView tabView)
-            {
-                return tabView;
-            }
+            source.Remove((TabItemData)tabItemData);
+            destination.Insert(index, (TabItemData)tabItemData);
```
marcelwgn, line 154: This can crash the app when you drag one tab out of the initial sample and then drag it out of its window.

marcelwgn (dismissed):

marcelwgn:
Closing this PR for now since there is quite a lot of work needed to get this mergeable but this PR can of course be reopened if you wish to continue working on this.

sabareesanms:
Hi @marcelwgn . Currently we are working on this PR and fixing the issues addressed.

niels9001:
> Hi @marcelwgn . Currently we are working on this PR and fixing the issues addressed.

@sabareesanms FYI - since the opening of this PR there were many changes in terms of project/solution structure and where files now live. That's why there are so many conflicts.

We are in a much better state now and we are not anticipating any other big changes like that at this point!

WinUIGallery/TabViewPages/TabViewWindowingSamplePage.xaml.cs
sabareesanms: @marcelwgn @llongley  Addressed this issue by adding the `tabItemDataList.Count > 1` in the `Tabs_TabTearOutRequested` event.

marcelwgn:
It seems that you can crash the app when dragging the last tab from a window 🤔

marcelwgn (changes requested):
It seems that you can crash the app when dragging the last tab from a window 🤔

sabareesanms:
@marcelwgn we have already added checks for not allowing drag when there is only one tab exists. Can you give more details if you still see the issue?

akanpatel2206:
Can we have screenshots/videos of before and after fix for visualization?

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        var item = (TabViewItem)contextMenu.Target;
+        ListView tabViewListView = null;
+        TabView tabView = null;
+
+        DependencyObject current = item;
+
+        while (current != null)
+        {
+            DependencyObject parent = VisualTreeHelper.GetParent(current);
```
Hemantxk, line 22: What is the immediate parent of TabViewItem? Is it ListView or TabView?

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+        int index = tabViewListView.IndexFromContainer(item);
+
+        if (index > 0)
+        {
+            MenuFlyoutItem moveLeftItem = new() { Text = "Move tab left" };
+            moveLeftItem.Click += (s, args) =>
+            {
+                if (tabView.TabItemsSource is IList itemsSourceList)
+                {
+                    var item = itemsSourceList[index];
```
Hemantxk, line 61: Will `index` get copied over to lambda as it will be cleaned up from this method before the Click event actually happen?

WinUIGallery/Helpers/UIHelper.cs
```diff
@@ -63,4 +63,21 @@ static public void AnnounceActionForAccessibility(UIElement ue, string annouceme
...
                                     AutomationNotificationProcessing.ImportantMostRecent, annoucement, activityID);
     }
+
+    public static T GetParent<T>(DependencyObject child) where T : DependencyObject
+    {
+        DependencyObject current = child;
+
+        while (current != null)
+        {
+            if (current is T parent)
```
Hemantxk, line 73: What's behavior of this function that **we want**  (not current behavior) when current DO is of type `T` and it has a parent of type `T`?

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+                    tabView.TabItems.Insert(index + 1, item);
+                }
+            };
+            contextMenu.Items.Add(moveRightItem);
+        }
+
+        // If the context menu ended up with no items at all, then we'll prevent it from being shown.
+        if (contextMenu.Items.Count == 0)
+        {
+            contextMenu.Hide();
```
Hemantxk, line 99: Shouldn't this be the job of caller?

WinUIGallery/Samples/SamplePages/TabViewWindowingSamplePage.xaml.cs
```diff
@@ -1,16 +1,55 @@
...
+
+    public bool IsInProgress
+    {
+        get { return (bool)GetValue(IsInProgressProperty); }
+        set { SetValue(IsInProgressProperty, value); }
+    }
+
+    public static readonly DependencyProperty HeaderProperty = DependencyProperty.Register("Header", typeof(string), typeof(TabItemData), new PropertyMetadata(""));
+    public static readonly DependencyProperty ContentProperty = DependencyProperty.Register("Content", typeof(string), typeof(TabItemData), new PropertyMetadata(""));
+    public static readonly DependencyProperty IsInProgressProperty = DependencyProperty.Register("Content", typeof(bool), typeof(TabItemData), new PropertyMetadata(false));
```
godlytalias, line 40: Shouldn't this be `IsInProgress`?

WinUIGallery/Samples/SamplePages/TabContentSampleControl.xaml
```diff
@@ -13,7 +13,7 @@
         <TextBlock Text="{Binding}" Style="{ThemeResource TitleTextBlockStyle}" />
         <TextBlock Text="Drag the Tab outside of the window to spawn a new window." Style="{ThemeResource SubtitleTextBlockStyle}" />
         <TextBlock Text="Notice that the state of the Tab is maintained in the new window. For example, if you toggle the ToggleSwitch ON, it will remain ON in the new window." Style="{ThemeResource BodyTextBlockStyle}" />
-        <ToggleSwitch x:Name="ControlToggle" Header="Turn on ProgressRing" Margin="0,8" />
-        <ProgressRing IsActive="{x:Bind ControlToggle.IsOn, Mode=OneWay}" HorizontalAlignment="Left" />
+        <ToggleSwitch x:Name="ControlToggle" Header="Turn on ProgressRing" Margin="0,8" IsOn="{x:Bind IsInProgress, Mode=TwoWay}" />
```
Hemantxk, line 16: I am unclear on why this change is being made & net impact on UX. Can you explain or add screenshot in PR Description.

WinUIGallery/Samples/SamplePages/TabViewWindowingSamplePage.xaml.cs
```diff
@@ -31,34 +70,51 @@ private void TabViewWindowingSamplePage_Loaded(object sender, RoutedEventArgs e)
...
+            };
+
+            inputNonClientPointerSource.ExitedMoveSize += (s, args) =>
+            {
+                tabTearOutWindow = null;
+            };
+
+            currentWindow.Closed += (s, args) =>
+            {
+                windowList.Remove(currentWindow);
```
godlytalias, line 99: Don't we have to unregister the `Closed` event as well from `currentWindow`?

WinUIGallery/Helpers/TabViewHelper.cs
marcelwgn: Its a TabViewListView which is a derivative of ListView

WinUIGallery/Helpers/UIHelper.cs
marcelwgn: I think it is reasonable to return the first result, otherwise you would need to scan up to the root to find the latest T

WinUIGallery/Samples/SamplePages/TabViewWindowingSamplePage.xaml.cs
```diff
@@ -85,64 +137,174 @@ private void Tabs_ExternalTornOutTabsDropping(TabView sender, TabViewExternalTor
...
+    private static void MoveDataItems(ObservableCollection<TabItemData> source, ObservableCollection<TabItemData> destination, object[] dataItems, int index)
+    {
+        foreach (object tabItemData in dataItems)
         {
-            if (current is TabView tabView)
-            {
-                return tabView;
-            }
+            source.Remove((TabItemData)tabItemData);
+            destination.Insert(index, (TabItemData)tabItemData);
```
marcelwgn, line 160: @sabareesanms This is where the app is crashing for me, trying to insert at index 1 while destination is empty. The repro is to move the last tab of a window out and you should be able to this the crash.

WinUIGallery/Samples/SamplePages/TabViewWindowingSamplePage.xaml
```diff
@@ -9,22 +9,36 @@
...
-            VerticalAlignment="Stretch"
-            AddTabButtonClick="Tabs_AddTabButtonClick"
-            CanTearOutTabs="True"
-            ExternalTornOutTabsDropped="Tabs_ExternalTornOutTabsDropped"
-            ExternalTornOutTabsDropping="Tabs_ExternalTornOutTabsDropping"
-            TabCloseRequested="Tabs_TabCloseRequested"
-            TabTearOutRequested="Tabs_TabTearOutRequested"
-            TabTearOutWindowRequested="Tabs_TabTearOutWindowRequested">
+        <TabView 
+		 x:Name="Tabs"
```
Hemantxk, line 13: Can you fix indentation and why order of event subscription being changed for all the items (only diff seems addition of `TabItemsSource`)?

WinUIGallery/Helpers/TabViewHelper.cs
sabareesanms: No. It displays empty contextMenu. So, we are explicitly adding it here to avoid UI fluctuations.

Hemantxk: 1> I see the caller function - TabViewContextMenu_Opening, already calls Hide on contextMenu if items are zero. Why is this needed here?
2> Are there other callers than the one I mentioned in this comment?
3> Did you check if you see UI fluctuations if it's not here?

WinUIGallery/Samples/SamplePages/TabViewWindowingSamplePage.xaml.cs
```diff
@@ -85,64 +137,174 @@ private void Tabs_ExternalTornOutTabsDropping(TabView sender, TabViewExternalTor
...
+        if (TabItemDataList.Count == 0)
         {
             WindowHelper.GetWindowForElement(this).Close();
         }
     }
+
+    private void TabViewContextMenu_Opening(object sender, object e)
+    {
+        // The contents of the context menu depends on the state of the application, so we'll build it dynamically.
+        MenuFlyout contextMenu = (MenuFlyout)sender;
```
Saharsh979, line 196: Is it safe to assume the sender is always a MenuFlyout?

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace WinUIGallery.Helpers;
+
+public static class TabViewHelper
+{
+    public static void PopulateTabViewContextMenu(MenuFlyout contextMenu)
+    {
+        contextMenu.Items.Clear();
+
+        var item = (TabViewItem)contextMenu.Target;
```
hiteshkrmsft, line 14: Should we check type first before typecasting?

WinUIGallery/Samples/SamplePages/TabViewWindowingSamplePage.xaml.cs
marcelwgn: Even if it is unlikely not to be, I would rather use `as` and return if it is in fact not a MenuFlyout instead of having the potential to crash here.

WinUIGallery/WinUIGallery.csproj
```diff
@@ -625,4 +625,8 @@
             </_NoneWithTargetPath>
         </ItemGroup>
     </Target>
+    <!-- Needed in order for running unpackaged to work properly. -->
```
marcelwgn, line 628: Why do we need this change now and not beforehand?

Zakariathr22:
I was thinking if it is possible to place the TabView items inside the left header of a TitleBar control?

niels9001:
Closing this PR as it seems to become stale. Feel free to reopen when work continues.

PR #1616, opened 2024-08-23
fix: Update Scroll View Animation Example and Sample Code Presenter

Zakariathr22:
## Description
Updated ScrollView Animation sample to display C# code, include default, accordion, and teleportation animations, and fixed code visibility handling to prevent overlap. With dynamically updated animation duration in the code displayed, main changes are:

- **Added Sample Code Files**:
  - Included three new sample code files for different scroll animations: - `ScrollViewSample3_DefaultAnimation_cs.txt` - `ScrollViewSample3_AccordionAnimation_cs.txt` - `ScrollViewSample3_TeleportationAnimation_cs.txt`

- **Updated Sample Code for Scroll Animations**:
  - Provided sample code for three types of animations (Default, Accordion, Teleportation) to be used in the `ScrollView_ScrollAnimationStarting` method.

- **Updated Sample Code for Animation Duration Handling**:
  - Implemented dynamic updating to ensure that the animation duration is reflected in the displayed sample code when the value changes.

- **Updated `ReevaluateVisibility` Method**:
  - Modified the `ReevaluateVisibility` method in `SampleCodePresenter` to only collapse the visibility when the content is empty. Leave the updated method visibility management to `HandlePresenterVisibility()` only in `ControlExample` to prevent overlap of XAML and C# code.

## Motivation and Context
By adding and updating sample code for different scroll animations (default, accordion, and teleportation), users can better understand and implement various animation styles in their ScrollView components.

It is also fixing this issue: https://github.com/microsoft/WinUI-Gallery/issues/1603

## How Has This Been Tested?
The changes were **manually tested** by verifying that the correct sample code is displayed for different animations, ensuring dynamic updates to the animation duration in the displayed example code, and confirming that the code presenter visibility is correctly managed.

## Screenshots:
image
image
image

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #1666, opened 2024-11-25
Refactor `SampleCodePresenter` to adjust code presenter visibility handling

Zakariathr22:
## Description
Refactored `SampleCodePresenter` to improve code presenter visibility handling and align its behavior with broader application usage patterns, the only change is:
- Updated the `SampleCodePresenter_Loaded` method in `SampleCodePresenter.xaml.cs` to enhance visibility handling for `CodePresenter` by removing logic to collapse code presenter visibility by default.

## Motivation and Context
- The previous implementation collapsed the visibility of `CodePresenter` in scenarios where it was not explicitly needed. This conflicted with scenarios like the `Iconography Page`, where an empty `CodePresenter` is defined and modified dynamically later.
- The responsibility for managing visibility when switching between code presenters (e.g., XAML and C#) already exists in the `ControlExample.HandlePresenterVisibility` method, making redundant visibility management unnecessary within `SampleCodePresenter` in most cases.
- To prevent issues in cases requiring default-loaded `CodePresenter`, its visibility is now set to Visible by default, and changes are applied only when truly needed.

## How Has This Been Tested?
**Manual Testing:**
- Verified CodePresenter visibility remains consistent across pages requiring dynamic updates.
- Tested transitions between XAML and C# in scenarios where `HandlePresenterVisibility()` manages visibility logic.

## Screenshots:
**- Before:** Code presenters collapsed unnecessarily where they defined empty to be modified dynamically later.
image

**- After:** Proper handling of empty code presenters and visibility.
image

## Types of changes

### DIFF
--- a/WinUIGallery/Controls/SampleCodePresenter.xaml.cs
+++ b/WinUIGallery/Controls/SampleCodePresenter.xaml.cs
@@ -91,7 +91,6 @@ namespace WinUIGallery.Controls
 
         private void SampleCodePresenter_Loaded(object sender, RoutedEventArgs e)
         {
-            ReevaluateVisibility();
             VisualStateManager.GoToState(this, GetSampleLanguageVisualState(), false);
             if (Substitutions != null)
             {


marcelwgn:
Thank you for your contribution. Unfortunately, this change is only concealing the root cause of the issue you are trying to address. The bug on the icons page was caused by the change introduced to SampleCodePresenter in #1616. If you like, we can repurpose this PR to fix that issue by reverting the change introduced in #1616 to the SampleCodePresenter class.

marcelwgn (changes requested):
Thank you for your contribution. Unfortunately, this change is only concealing the root cause of the issue you are trying to address. The bug on the icons page was caused by the change introduced to SampleCodePresenter in #1616. If you like, we can repurpose this PR to fix that issue by reverting the change introduced in #1616 to the SampleCodePresenter class.

Zakariathr22:
@marcelwgn
Thank you for the feedback and for pointing out the root cause of the issue. I understand now that it’s related to the changes in #1616. I’m happy to update this PR to revert those changes and fix the issue properly, I’ll work on it immediately. Please let me know if there’s anything else I should keep in mind while making the update.

marcelwgn:
Thank you for all your effort @zakariathr22 ! I think the best approach would be to leave the code as is for now until we have a proper fix instead of reverting all of 1616 since that would also mean losing the valuable changes you made in there.

If you like, I can provide some pointers how I would fix the issue you are describing while reverting the Visibility feature of the SampleCodePresenter :)

PR #1660, opened 2024-11-19
Update installation instructions

niels9001:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #1667, opened 2024-11-25
[UX] New icon

niels9001:
## Description
This PR adds the new WinUI Gallery icon, that is part of the new WASDK/WinUI icon family.

- Updated the tile/icon assets with the new logo.
- Updated the readme with new images, and added a screenshot to the readme, so new developers understand what the app would look like
- Updated the "Getting started" icon with the new WinUI logo: it is about Getting started with WinUI after all, so having the WinUI logo there makes much more sense!

image
image

image


Jay-o-Way:
Maybe this icon should be used for this repo too

Jay-o-Way:
Here you add the "3" to the app name in ReadMe, and in #1660 you specifically remove it. What's the full name?

Jay-o-Way:
Nit - I just saw this coloring.

README.md
```diff
@@ -1,15 +1,27 @@
-# WinUI 3 Gallery 
+![WinUI Gallery hero image](README_Images/WinUIGalleryLanding.PNG)
 
+<h1 align="center">
+    WinUI 3 Gallery
+</h1>
+
+<p align="center">
 [![Build Status](https://dev.azure.com/ms/microsoft-ui-xaml/_apis/build/status%2Fmicrosoft.WinUI-Gallery?branchName=main)](https://dev.azure.com/ms/microsoft-ui-xaml/_build/latest?definitionId=612&branchName=main) [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square)](http://makeapullrequest.com) [![GitHub license](https://img.shields.io/badge/license-MIT-blue.svg?style=flat-square)](https://github.com/your/your-project/blob/master/LICENSE)
```
marcelwgn, line 8: I think adding this p-tag broke the links in our readme.

marcelwgn (commented):
Thank you for updating our icon! It seems that our readme links are now broken so I'm holding off on approving this until we find the issue.

P.S: Thank you for removing the old images

README.md
Jay-o-Way: Most likely. Mixing HTML and MD can do that.

Jay-o-Way: Opening that URL shows a (error) message:

> {"count":1,"value":{"Message":"Build pipeline microsoft.WinUI-Gallery was not found."}}

karkarl:
Wondering why we need 5 million versions of the icon in the repo and .csproj? :)

niels9001:
> Wondering why we need 5 million versions of the icon in the repo and .csproj? :)

This is the output from the design file. It ensures that the icon is crisp on every DPI/resolution combo.

When using the tool in the Package manifest, it generates different files from a single image, and as a result, the icon isn't as crisp.

If only we could just set an .ico or .svg 🫠

marcelwgn (approved):

Jay-o-Way:
@niels9001 I think there's an issue with the icon in the top-left of the window. The file _BadgeLogo_ (at 100%) is 24×24 pixels. In the XAML, the icon is scaled to 18px. This makes it unsharp.

PR #1668, opened 2024-11-27
Resolve `Iconography Page` issue, improve `Control Example` behavior and `ScrollViewPage` Example

Zakariathr22:
## Description  
This PR reverts changes introduced in #1616 to the `ReevaluateVisibility` method in the `SampleCodePresenter` class. The modifications in #1616 caused unintended side effects, including the issue observed on the icons page. Reverting these changes resolves the problem and restores expected behavior.  

Additionally, this PR includes:  
- **Fix for overlapping issue with `SampleCodePresenter`:** Updates `ControlExample` to use `ContentPresenter` for visibility management, ensuring no overlapping when C# code changed.  
- **Ensure correct C# code sample updates when animation duration changes.**

## Motivation and Context  
Fixing the issue on the icons page caused by unintended side effects of changes in #1616, ensuring proper visibility management for code presenters without affecting other functionality. Additional fixes address overlapping issues and dynamic sample updates to enhance usability.  

## How Has This Been Tested?  
Manually tested:  
- Verified that reverting #1616 changes to the `ReevaluateVisibility` method.  
- Confirmed that overlapping of code presenters is resolved.  
- Ensured the animation duration is correctly updated in the displayed C# code in the ScrollView example.

- **Before reverting:**  
image  
- **After reverting:**  
image  

- **Before resolving overlapping issue:**  
image
- **After resolving overlapping issue:**  
image

## Types of changes  
  
### DIFF
--- a/WinUIGallery/ControlPages/ScrollViewPage.xaml
+++ b/WinUIGallery/ControlPages/ScrollViewPage.xaml
@@ -209,6 +209,7 @@
                     <NumberBox x:Name="nbAnimationDuration"
                         Grid.Row="1" Grid.Column="1"
                         Minimum="1000"
+                        Value="1500"
                         Maximum="5000"
                         SpinButtonPlacementMode="Inline"
                         SmallChange="500"
--- a/WinUIGallery/ControlPages/ScrollViewPage.xaml
+++ b/WinUIGallery/ControlPages/ScrollViewPage.xaml
@@ -209,7 +209,6 @@
                     <NumberBox x:Name="nbAnimationDuration"
                         Grid.Row="1" Grid.Column="1"
                         Minimum="1000"
-                        Value="1500"
                         Maximum="5000"
                         SpinButtonPlacementMode="Inline"
                         SmallChange="500"
--- a/WinUIGallery/ControlPages/ScrollViewPage.xaml.cs
+++ b/WinUIGallery/ControlPages/ScrollViewPage.xaml.cs
@@ -20,7 +20,7 @@ namespace WinUIGallery.ControlPages
 {
     public sealed partial class ScrollViewPage : Page
     {
-        private double prvAnimationDuration;
+        private double prvAnimationDuration = 1500;
 
         public ScrollViewPage()
         {
@@ -48,6 +48,7 @@ namespace WinUIGallery.ControlPages
             };
             nbZoomFactor.NumberFormatter = formatter;
 
+            nbAnimationDuration.Value = prvAnimationDuration;
             Example3.CSharp = Example3.CSharp.Replace("nbAnimationDuration.Value", nbAnimationDuration.Value.ToString());
             prvAnimationDuration = nbAnimationDuration.Value;
 
--- a/WinUIGallery/ControlPages/ScrollViewPage.xaml.cs
+++ b/WinUIGallery/ControlPages/ScrollViewPage.xaml.cs
@@ -11,6 +11,7 @@ using Microsoft.UI.Composition;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Numerics;
 using Windows.Globalization.NumberFormatting;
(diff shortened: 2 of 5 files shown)


Zakariathr22:
@marcelwgn
Thank you for the suggestion! I’ve implemented your recommended approach, wrapping the two `SampleCodePresenter` controls in separate `ContentPresenter` elements and modifying the visibility of the `ContentPresenter` instead. It works well, and everything is functioning as expected.  Could you please review the changes when you have a chance?

WinUIGallery/ControlPages/ScrollViewPage.xaml
```diff
@@ -209,7 +209,6 @@
                     <NumberBox x:Name="nbAnimationDuration"
                         Grid.Row="1" Grid.Column="1"
                         Minimum="1000"
-                        Value="1500"
```
marcelwgn, line 212: Related?

WinUIGallery/ControlPages/ScrollViewPage.xaml.cs
```diff
@@ -20,7 +20,7 @@ namespace WinUIGallery.ControlPages
 {
     public sealed partial class ScrollViewPage : Page
     {
-        private double prvAnimationDuration;
```
marcelwgn, line 23: Same here, is this related to the fix in question?

marcelwgn (approved):
Thank you for making these changes, looks good to me.

WinUIGallery/ControlPages/ScrollViewPage.xaml.cs
Zakariathr22: These changes are related to **the third example in the ScrollView page** and ensure that the displayed C# code correctly updates when the animation duration changes. 
It's a **minor update** to ensure that the C# code dynamically reflects the changes to the animation duration in the proper way (it is not related to the `SampleCodePresenter` or `ControlExample`.)

marcelwgn: I am not sure your changes fix anything. After reverting your changes to ScrollViewPage locally, I don't see any faulty behavior.

Zakariathr22: If you define the value of the `NumberBox` in XAML, its initial value will briefly be set to `0` before changing to the specified XAML value.

In the code, which updates a "previous duration" value whenever the `NumberBox` value changes, this behavior causes an issue: setting the value to `1500` in XAML first changes it to `0`, then to `1500`. This unintended sequence results in every occurrence of `0` in the C# sample code being replaced with `1500`.  

To resolve this, I initialized both the "previous duration" value and the `NumberBox` value to `1500` directly in the C# code, ensuring it starts with the correct value from the beginning and ensure correct C# code sample updates when animation duration changes

**Before Changes**
img-1

**After Changes**
img-2

This is just my own approach, and if you have any other suggestions, I’d be happy to implement them!

marcelwgn: Ah I see, good catch! I think the more elegant approach would be to have `nbAnimationDuration_ValueChanged` not do that old/new value replacement but rather have an a copy of the original text and replace on top of that. That would also allow us to get rid of that `prvAnimationDuration` field which might be a bit confusing.

Zakariathr22: I had considered this approach before but initially avoided it due to performance concerns since it involves repeatedly reading from disk using File I/O. However, you're absolutely right—relying on the previous value for replacement isn't very elegant and can be confusing.

To address both code clarity and performance, I refactored the code to use a caching mechanism. The sample code files are now loaded once and stored in memory, ensuring we avoid repeated disk reads while maintaining a clear and more maintainable structure. For example, the `GetExample3CodeContent` method ensures that file content is read only on the first request, with subsequent updates working on cached data.

I think this implementation is both efficient and easier to understand. Let me know if you have further suggestions!

marcelwgn: Good idea with cache, like that approach here!

marcelwgn (approved):
Looks good to me, thank you for making these changes!

Zakariathr22:
> Looks good to me, thank you for making these changes!

Thanks, this was thanks to your suggestions, exited to keep contributing. 🚀

Jay-o-Way:
@karkarl This one really should have been in the 2.5.0 release

PR #1685, opened 2024-12-10
Remove Reveal Focus sample

niels9001:
Reveal Focus is not applicable to WinUI 3.


marcelwgn (approved):
Good catch, the page is really unspectacular without reveal...

PR #1711, opened 2024-12-24
Add titlebar height setting sample

InfyniteHeap:
## Description

Provide a sample to demonstrate how to set the height of titlebar after extending contents into titlebar.
## Motivation and Context

WinUI supports to set titlebar height, but the `TitleBar` page doesn't contain a sample to that.
## How Has This Been Tested?

Just build and run this app, and trun to `TitleBar` page, and then test whether the combobox works as expected.

屏幕截图 2024-12-24 222750
屏幕截图 2024-12-24 222754
屏幕截图 2024-12-24 222757

## Types of changes

### DIFF
--- a/WinUIGallery/ControlPages/TitleBarPage.xaml.cs
+++ b/WinUIGallery/ControlPages/TitleBarPage.xaml.cs
@@ -211,18 +211,17 @@ namespace WinUIGallery.ControlPages
 
         private void TitleBarHeightComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e)
         {
-            var selectedHeight = ((ComboBoxItem)titlebarHeight.SelectedItem)?.Tag?.ToString();
+            var selectedHeight = titlebarHeight.SelectedItem.ToString();
             var window = WindowHelper.GetWindowForElement(this);
 
-            if (selectedHeight != null && window.ExtendsContentIntoTitleBar)
+            if (selectedHeight != null && window != null && window.ExtendsContentIntoTitleBar)
             {
                 window.AppWindow.TitleBar.PreferredHeightOption = selectedHeight switch
                 {
-                    "Collapsed" => TitleBarHeightOption.Collapsed,
-                    "Standard" => TitleBarHeightOption.Standard,
                     "Tall" => TitleBarHeightOption.Tall,
-                    // This case will never be reached
-                    _ => throw new InvalidOperationException("Unreachable code reached in string pattern matching.")
+                    "Standard" => TitleBarHeightOption.Standard,
+                    "Collapsed" => TitleBarHeightOption.Collapsed,
+                    _ => TitleBarHeightOption.Tall
                 };
             }
         }
--- a/WinUIGallery/ControlPages/TitleBarPage.xaml.cs
+++ b/WinUIGallery/ControlPages/TitleBarPage.xaml.cs
@@ -209,6 +209,24 @@ namespace WinUIGallery.ControlPages
             SetClickThruRegions(rectArr);
         }
 
+        private void TitleBarHeightComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            var selectedHeight = ((ComboBoxItem)titlebarHeight.SelectedItem)?.Tag?.ToString();
+            var window = WindowHelper.GetWindowForElement(this);
+
+            if (selectedHeight != null && window.ExtendsContentIntoTitleBar)
+            {
+                window.AppWindow.TitleBar.PreferredHeightOption = selectedHeight switch
+                {
+                    "Collapsed" => TitleBarHeightOption.Collapsed,
+                    "Standard" => TitleBarHeightOption.Standard,
+                    "Tall" => TitleBarHeightOption.Tall,
+                    // This case will never be reached
+                    _ => throw new InvalidOperationException("Unreachable code reached in string pattern matching.")
+                };
+            }
+        }
+
         private void AddInteractiveElements_Click(object sender, RoutedEventArgs e)
         {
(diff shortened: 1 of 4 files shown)


Jay-o-Way:
Is this based on on the new experimental titlebar control (coming hopefully in WASDK 1.7) maybe? I think there's an issue with the phrasing like "... now allows ..." It is unclear what "now" means: is it a SDK version, a specific date or something else? And if it relies on a SDK version, this must be clearly mentioned.
Also, after a while - "now" will become outdated at some point in time.


Besides that: try and keep redundancy to a minimum and use normal sentence casing (see "Titlebar customizations")

WinUIGallery/ControlPages/TitleBarPage.xaml.cs
```diff
@@ -209,6 +209,24 @@ private void setTxtBoxAsPasthrough(FrameworkElement txtBoxNonClientArea)
...
+
+            if (selectedHeight != null && window.ExtendsContentIntoTitleBar)
+            {
+                window.AppWindow.TitleBar.PreferredHeightOption = selectedHeight switch
+                {
+                    "Collapsed" => TitleBarHeightOption.Collapsed,
+                    "Standard" => TitleBarHeightOption.Standard,
+                    "Tall" => TitleBarHeightOption.Tall,
+                    // This case will never be reached
+                    _ => throw new InvalidOperationException("Unreachable code reached in string pattern matching.")
```
Zakariathr22, line 225:
```suggestion
                    _ => TitleBarHeightOption.Standard
```
This is better than throwing an exception (even if the case is unreachable) because it provides a safer fallback, ensuring stability without unnecessary error handling.

Zakariathr22 (changes requested):
Good observation! I think demonstrating how to customize the TitleBar height is an important addition.

I have several suggestions to improve your contribution:
- Modify the new `ControlExample` to look like this:
```xaml
        <local:ControlExample HeaderText="Titlebar Customizations : Titlebar height">
                <StackPanel Orientation="Vertical" Spacing="10">
                    <TextBlock TextWrapping="Wrap">
                        WinUI custom titlebar supports setting a custom height by configuring the PreferredHeightOption property of the titlebar. This feature is useful when you need to adjust the titlebar height to match your app's design or accommodate additional content.
                    </TextBlock>
                    <InfoBar IsOpen="true" IsClosable="false" Severity="Warning">
                        <TextBlock>
                                This option will be invalid when the style of titlebar is reset to system titlebar!
                        </TextBlock>
                    </InfoBar>
                    <StackPanel>
                        <ComboBox x:Name="titlebarHeight"
                                  Header="Set titlebar height"
                                  SelectionChanged="TitleBarHeightComboBox_SelectionChanged"
                                  SelectedIndex="0">
                            <x:String>Tall</x:String>
                            <x:String>Standard</x:String>
                            <x:String>Collapsed</x:String>
                        </ComboBox>
                    </StackPanel>
                </StackPanel>
            </local:ControlExample.Example>
            <local:ControlExample.CSharp>
                <x:String xml:space="preserve">
// Retrieve the main window and extend its content into the title bar for custom styling.
Window window = App.MainWindow;
window.ExtendsContentIntoTitleBar = true;

// Set the preferred height option for the title bar
window.AppWindow.TitleBar.PreferredHeightOption = TitleBarHeightOption.$(TitleBarHeightOption);
                </x:String>
            </local:ControlExample.CSharp>
            <local:ControlExample.Substitutions>
                <local:ControlExampleSubstitution Key="TitleBarHeightOption"
                                                  Value="{x:Bind titlebarHeight.SelectedItem, Mode=OneWay}" />
        </local:ControlExample>
```
> Note that the C# element preserves whitespace formatting inside the string using `xml:space="preserve"`.

- Then, update the `TitleBarHeightComboBox_SelectionChanged` method as follows:
```csharp
private void TitleBarHeightComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e)
{
  var selectedHeight = titlebarHeight.SelectedItem.ToString();
  var window = WindowHelper.GetWindowForElement(this);

  if (selectedHeight != null && window != null && window.ExtendsContentIntoTitleBar)
  {
      window.AppWindow.TitleBar.PreferredHeightOption = selectedHeight switch
      {
          "Tall" => TitleBarHeightOption.Tall,
          "Standard" => TitleBarHeightOption.Standard,
          "Collapsed" => TitleBarHeightOption.Collapsed,
          _ => TitleBarHeightOption.Tall
      };
  }
}
```
- Lastly, you can delete the unused `TitleBarSample4.txt` file after making these changes.

These changes will ensure the following:
1. Improved phrasing, as mentioned by @Jay-o-Way .
2. Simpler C# code with dynamic updates when the selected TitleBar height changes.

WinUIGallery/ControlPages/TitleBarPage.xaml.cs
```diff
@@ -209,6 +209,23 @@ private void setTxtBoxAsPasthrough(FrameworkElement txtBoxNonClientArea)
...
+
+            if (selectedHeight != null && window != null && window.ExtendsContentIntoTitleBar)
+            {
+                window.AppWindow.TitleBar.PreferredHeightOption = selectedHeight switch
+                {
+                    "Tall" => TitleBarHeightOption.Tall,
+                    "Standard" => TitleBarHeightOption.Standard,
+                    "Collapsed" => TitleBarHeightOption.Collapsed,
+                    _ => TitleBarHeightOption.Tall
+                };
```
MGGSK, line 225: I think it would be better to use the App.GetEnum function instead.

```suggestion
                window.AppWindow.TitleBar.PreferredHeightOption = App.GetEnum<TitleBarHeightOption>(selectedHeight);
```

MGGSK (changes requested):

karkarl (approved):

PR #1703, opened 2024-12-14
Fixed 24H2 build number

HotCakeX:
Fixed 24H2 build number in the issue template. The correct build is `26100` not `26052`.

PR #1715, opened 2025-01-05
Revert "Fixed 24H2 build number"

marcelwgn:
Reverts microsoft/WinUI-Gallery#1703

While technically, 26052 was the first build of 24H2, 26100 are the GA releases and I think it makes more sense to stick with that version.


niels9001 (approved):

Jay-o-Way:
@marcelwgn the change contradicts your message. You reverted back to **26052**.

marcelwgn:
Oh you are completely right, I must have somehow misread the diff. Terribly sorry about that.

Issue #1712, opened 2024-12-30
WinUI Themes Resources (GitHub) link in the WinUI 3 Gallery is broken


downie2k:
![Image](https://github.com/user-attachments/assets/4f407fce-626a-4d0b-9026-4e45229a6976)

![Image](https://github.com/user-attachments/assets/4f6ebf05-faa8-434c-9a1f-2d6a1c6cee71)

Pictures should be sufficient, if you have any question feel free.

Dominic

PR #1716, opened 2025-01-06
Fix the theme resource link

MGGSK:
Fixed the "WinUI Theme Resources (Github)" being broken.

## Description
Changed the link in "ControlInfoData.json" from "https://github.com/microsoft/microsoft-ui-xaml/blob/winui3/release/X.Y-stable/controls/dev/CommonStyles/Common_themeresources_any.xaml" to "https://github.com/microsoft/microsoft-ui-xaml/blob/winui3/release/X.Y-stable/src/controls/dev/CommonStyles/Common_themeresources_any.xaml"

## Motivation and Context
Fixes https://github.com/microsoft/WinUI-Gallery/issues/1712

## How Has This Been Tested?
Manual

## Screenshots:
image
image

## Types of changes

### DIFF
--- a/WinUIGallery/DataModel/ControlInfoData.json
+++ b/WinUIGallery/DataModel/ControlInfoData.json
@@ -24,7 +24,7 @@
             },
             {
               "Title": "WinUI Theme Resources (GitHub)",
-              "Uri": "https://github.com/microsoft/microsoft-ui-xaml/blob/winui3/release/X.Y-stable/controls/dev/CommonStyles/Common_themeresources_any.xaml"
+              "Uri": "https://github.com/microsoft/microsoft-ui-xaml/blob/winui3/release/X.Y-stable/src/controls/dev/CommonStyles/Common_themeresources_any.xaml"
             }
           ]
         },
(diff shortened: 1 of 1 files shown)


marcelwgn (changes requested):
Thank you for fixing this link, however I agree with @jay-o-way. If we fix the link, we should fix every occurrence of the faulty version.

Regarding if we should point to main, I think pointing to the latest stable release is the most helpful for most people since a majority will be on stable releases. And I agree we should strip the en/us from the documentation links.

marcelwgn (approved):
Thank you for making these changes @MGGSK! LGTM!

Jay-o-Way:
> Regarding if we should point to main, I think pointing to the latest stable release is the most helpful for most people since a majority will be on stable releases.

But @marcelwgn , `main` is the latest stable release now. Check out pr and the connected issue.

marcelwgn:
Oh I guess we can move to the main branch then for those two links too. Good callout @Jay-o-way!

niels9001:
@jay-o-way is right, we no longer need to deal with versions in URLs since the WinUI repo moved all the latest code into `main` a couple of months back!

Issue #1614, opened 2024-08-22
Update WinUIGallery directory to be aligned with WinUI Templates


karkarl:

Jay-o-Way:
Does this touch #1246 and/or #1542 ?

lostluna1:
I guess what the publisher of this theme is referring to is to make WinUIGallery consistent with the design patterns created by the WinUI extension

Jay-o-Way:
@karkarl what did you mean?

karkarl:
> [@karkarl](https://github.com/karkarl) Please describe the idea for this issue

Updated the description.

Zakariathr22:
Well done! Does this PR Close #1614?

Jay-o-Way:
@niels9001 this isn't fixed yet

niels9001:
> @niels9001 this isn't fixed yet

I think we have done a major restructuring of the project that is along these lines.

If there are other specific things we'd want to change, we can create specific issues for it.

Jay-o-Way:
> I think we have done a major restructuring of the project that is along these lines.
> 
> If there are other specific things we'd want to change, we can create specific issues for it.

True, but I still have #1744 in mind...

niels9001:
> > I think we have done a major restructuring of the project that is along these lines.
> > If there are other specific things we'd want to change, we can create specific issues for it.
> 
> True. Many thanks to you! But I still have [#1744](https://github.com/microsoft/WinUI-Gallery/issues/1744) in mind...

Sure, but that's another issue so we can track any follow-up there!

niels9001:
Closing this issue as this is fixed in 2.6.0.0

Issue #1673, opened 2024-12-02
Code quality: consolidate xmlns


Jay-o-Way:
Just looking at the namespaces for `using:WinUIGallery.ControlPages` for example. Well, I think this screenshot says it all...
![Image](https://github.com/user-attachments/assets/7e3a6a86-721d-4f58-bd13-98c6b3962053)

Jay-o-Way:
Let's decide on the proper namespaces here:

| xmlns: | using: |
| --- | --- |
| local | WinUIGallery |
|  | WinUIGallery.Common |
|  | WinUIGallery.Data |
|  | WinUIGallery.DesktopWap.Controls |
|  | WinUIGallery.ControlPages |
|  | CommunityToolkit.WinUI |
|  | CommunityToolkit.WinUI.Controls |
|  | Microsoft.UI.Xaml.Media |
|  | Microsoft.UI.Xaml.Controls.AnimatedVisuals |
|  | Microsoft.UI.Xaml.Controls?IsApiContractPresent |

P.S. For what is `IsApiContractPresent` needed?

marcelwgn:
Yeah I think consolidating the xmlns namespaces sounds good though I wouldnt say "local" is a fixed name. From my experience, local usually often just refers to the namespace of the control.

Re API Contract: I think this was just missed when moving to WinUI 3 and cleaning up the UWP remnants.

Jay-o-Way:
Thanks. The combination `local` + `WinUIGallery` is actually very common in the repo, that's why I took the liberty to fill it in, as the first example. But if you want it to be something else, let me know. You can edit the above comment if you want.

I think it's a matter of preference to use abbreviations or words for the namespaces. Short words might be easier to decipher.

So, that means the API contract thing can be removed, am I right?

Jay-o-Way:
Idea: we could rename `WinUIGallery.Controls` to (xmlns:internal) `InternalControls`? Given the many other things that use the word "controls" - it's what this app is all about 😆

Jay-o-Way:
Five cases of `xmlns:converters` are left: `WinUIGallery.Converters` versus `CommunityToolkit.WinUI.Converters`

marcelwgn:
Thanks for the ping @Jay-o-Way ! I think we should do the following:
- If the page is in the same namespace, use local for that namespace e.g. a Control in the folder WinUIGallery.Controls uses `local` for that namespace
- Use toolkit for CommunityToolkit
[...]

Issue #1542, opened 2024-05-20
UI Navigation: Controls, Design, Accessibility, Fundamentals


Jay-o-Way:
### Issue type

enhancement

### Description

This app used to be named "Control Gallery", but has evolved to encompass many more things. The controls are very prominent in the current navigation structure. But the other categories should "live" on the same level. Note that in the UI: Design Guidance > Accessibility. 

Idea: one **Navigation** split in two areas: a top part with these categories, and the second part is dynamic, filled with sub-items for the selected category.

``` txt
╔════════════════════════════════════╗
║ WinUI Gallery                   –□×║
║                 ┌──────────────────║
║ Fundamentals    │                  ║
║ Design          │                  ║
║ Controls        │                  ║
║ Accessibility   │                  ║
║─────────────────┤                  ║
║ (dynamic menu)  │                  ║
║                 │                  ║
║─────────────────┤                  ║
║ Settings        │                  ║
╚════════════════════════════════════╝
```

Jay-o-Way:
Also note that in the UI: **Design Guidance** > **Accessibility**.

Jay-o-Way:
ping

niels9001:
I'd separate the way the .csproj is structured vs. the UI discsion to keep things simpler, as those are 2 separate things.

Regarding the UI, here's a simple fix proposal. Implementing dynamic menus would require additional work and isn’t commonly seen in NavigationView patterns across Windows apps:

![Image](https://github.com/user-attachments/assets/b6d0b95c-4873-4b1a-b2a7-9882307fb5ac)

- The NavigationView button and search moves to the titlebar once the new `TitleBar` control is available in WASDK 1.7 (we need to do this regardless to comply with the latest design guidance), and it saves some valuable vertical space in the menu.
- Re-order the Fundamentals / Design / A11y items
- Make a clear(er) separation between those pages and the actual control samples by adding a "Controls" header, rename "All samples" to "All" as it's already available under "Controls".

Q: Do we need the "All samples" page at all?


@Jay-o-Way @marcelwgn @zinniaa thoughts?

marcelwgn:
Thank you for raising this issue @Jay-o-Way. Given we have the fundamentals section now, I think it would be a good idea to move the a11y section to the root.

Jay-o-Way:
> Implementing dynamic menus would require additional work and isn’t commonly seen in NavigationView patterns across Windows apps

I hear you.

[...]

PR #1732, opened 2025-01-28
Cleanup & folder structure improvements

niels9001:
This PR removes unused files, and restructures the entire project:

- Control pages and sample-related code now lives in the `Samples` folder.
- Gallery related helpers, pages and other things live in their respective folders.
- A lot of XAML styler formatting (we need to create a script for this).
- Picking up the changes from: https://github.com/microsoft/WinUI-Gallery/pull/1733
- Removing a bunch of dead-code / unused namings

image


Jay-o-Way:
Whoah, this is huge! Interesting! I see you're touching a few issues, like
- #1542 (larger idea for nav in issue)
- #1673

WinUIGallery/Samples/ControlPages/Accessibility/AccessibilityColorContrastPage.xaml
```diff
@@ -37,21 +37,16 @@
...
                 visual acuity, but this will also ensure visibility and legibility under a wide range of lighting
-                conditions, screens, and device settings.
-                <LineBreak />
+                conditions, screens, and device settings.<LineBreak />
                 <LineBreak />
-                Check out the
-                <Hyperlink NavigateUri="https://accessibilityinsights.io/">Accessibility Insights</Hyperlink>
-                app to help you find and fix accessibility issues in your Windows apps.
-            </Paragraph>
+                Check out the<Hyperlink NavigateUri="https://accessibilityinsights.io/">Accessibility Insights</Hyperlink>
```
Jay-o-Way, line 48: FYI: Xaml Styler always messes up the spaces before/after elements like these :(

niels9001: Fixed!

karkarl:
Looks great! 

I wonder if "Controls" directory containing WinUI-Gallery defined userControls should be renamed to "UserControls" or something else instead? It always confuses me since this is a controls app gallery. This can be a separate PR.

Zakariathr22:
Well done! Does this PR Close #1614?

Jay-o-Way:
> Looks great! 
> 
> I wonder if "Controls" directory containing WinUI-Gallery defined userControls should be renamed to "UserControls" or something else instead? It always confuses me since this is a controls app gallery. This can be a separate PR.

Totally agree! I've been thinking about that for a while too. "XamlControls" for the other folder/namespace makes sense too...

WinUIGallery/Assets/Data/ControlInfoDataSchema.json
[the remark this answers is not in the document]
niels9001: Moved it to the samples folder, so now everything sample related is in a single place.

niels9001:
> > Looks great!
> > I wonder if "Controls" directory containing WinUI-Gallery defined userControls should be renamed to "UserControls" or something else instead? It always confuses me since this is a controls app gallery. This can be a separate PR.
> 
> Totally agree! I've been thinking about that for a while too. "XamlControls" for the other folder/namespace makes sense too... Let's discuss this in #1673

We've moved most sample-related content in the Samples folder, so if you don't need to touch the Gallery-shell that'd be the place to be. Everything else is related to the app itself. I want to avoid having differences in folder names and namespaces, and updating those now is a pain due to the many conflicts and changes required - if we want to do this is in the future, better to do it in a separate PR.

PR #1738, opened 2025-01-29
Adding XAML styler config

niels9001:
Using the same XAML styler config that we use in the Windows Community Toolkit: https://github.com/CommunityToolkit/Windows/blob/9830e43451225fb63b1d5785408368fbf69c6c7f/settings.xamlstyler


settings.xamlstyler
```diff
@@ -0,0 +1,42 @@
...
+        "mc:Ignorable, d:IsDataSource, d:LayoutOverrides, d:IsStaticText",
+        "Storyboard.*, From, To, Duration"
+    ],
+    "FirstLineAttributes": "",
+    "OrderAttributesByName": true,
+    "PutEndingBracketOnNewLine": false,
+    "RemoveEndingTagOfEmptyElement": true,
+    "SpaceBeforeClosingSlash": true,
+    "RootElementLineBreakRule": 0,
+    "ReorderVSM": 1,
```
marcelwgn, line 31: Shouldnt this keep VSMs at the top?

niels9001: I think setting this to `2` would put them `Last`, `1` to `First`

marcelwgn: When I opened your PR and saved the file, the VSM were at the bottom though. Is there something else I need to do?

michael-hawker:
Should a check be added to the CI? I imagine a pass needs to be done on everything already here though too?

settings.xamlstyler
AndrewKeepCoding: It's not working because the config file is required to be in/under solution folders. Moving it to WinUI-Gallery/WinUIGallery should work.

marcelwgn: Thanks for pointing this out, lets wait for https://github.com/microsoft/WinUI-Gallery/pull/1748 before merging this then :)

AndrewKeepCoding:
The *xamlstyler.settings* file itself LGTM. But we need to decide how to workaround/fix #1747 first.

marcelwgn:
I agree with @AndrewKeepCoding, having XAML Styler cause crashes because the properties are resorted is not good

niels9001:
> I agree with @AndrewKeepCoding, having XAML Styler cause crashes because the properties are resorted is not good

Agreed. Let's close it for now.

Issue #1742, opened 2025-02-04
NavigationView highlights "Accessibility" when navigating to "All samples"


MGGSK:
### Which version of the app?

WinUI 3 Gallery

### Description

Steps to reproduce the bug:
1. Open the app
2. Open the "All samples" page

### Screenshots

![Image](https://github.com/user-attachments/assets/259f7dbf-c9eb-4cf8-938d-b93d677535fa)

### Windows version

Windows 11 24H2 (26100)

### Additional context

This is the code in `AllControlsPage.xaml.cs`.`OnNavigatedTo` that causes the issue:
```
NavigationRootPageArgs args = (NavigationRootPageArgs)e.Parameter;

var menuItem = (Microsoft.UI.Xaml.Controls.NavigationViewItem)args.NavigationRootPage.NavigationView.MenuItems.ElementAt(2);
menuItem.IsSelected = true;
```

PR #1746, opened 2025-02-05
Fix All Samples selection

niels9001:
To test:

- Click on All Samples
- All Samples item should now be selected, vs "Accessibility"

Closes: https://github.com/microsoft/WinUI-Gallery/issues/1742


marcelwgn (approved):
Great fix, surprised we didnt fix this earlier

zinniaa (approved):

PR #1749, opened 2025-02-09
Fixed a small XAML error

HotCakeX:
Fixed a small XAML error. x2 double quotes.


marcelwgn (approved):
Good catch, thanks for the fix!

PR #1778, opened 2025-02-16
Minor menu reordering

niels9001:
- A few items in the NavView were either not alphabetically sorted, or just had a weird ordering in general.
- Moved ItemsRepeater under Collections. This control is more used in context of binding a collection, like ItemsView/ListView etc. vs. in context of a layout panel like a Grid or StackPanel

### DIFF
--- a/WinUIGallery/Samples/Data/ControlInfoData.json
+++ b/WinUIGallery/Samples/Data/ControlInfoData.json
@@ -782,6 +782,46 @@
           ]
         },
         {
+          "UniqueId": "ItemsRepeater",
+          "Title": "ItemsRepeater",
+          "BaseClasses": [
+            "Object",
+            "DependencyObject",
+            "UIElement",
+            "FrameworkElement"
+          ],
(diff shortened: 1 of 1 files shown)


WinUIGallery/Samples/Data/ControlInfoData.json
```diff
@@ -701,6 +644,64 @@
             "GridView",
             "ListView"
           ]
+        },
+        {
+          "UniqueId": "StandardUICommand",
```
marcelwgn, line 649: Should this be before Swipecard or is the grouping of the UI commands intentional here?

niels9001: Figured `StandUICommand` and `XAMLUICommand` are related, so putting them closer together?

niels9001: Looking at these.. should those 2 pages be combined and called `UICommand`?

marcelwgn (approved):

PR #1788, opened 2025-02-26
Fix the window cannot be dragged after resetting to system titlebar then reset it back

### LINKED ISSUE #1783: The window cannot be dragged after resetting to system titlebar then reset it back
### Which version of the app?

WinUI 3 Gallery

### Description

Image

Actually why using it when it's still experimental? Is WinUI Gallery supposed to demo experimental apis?

### Screenshots

_No response_

### Windows version

Windows 11 24H2 (26100)

### Additional context

_No response_

HO-COOH:
Closes #1783 

## Description
1. You never need to set `window.SetTitlebar(null)` when you have `ExtendsContentIntoTitleBar == false`
2. When you click the "Reset to default custom titlebar" button to reset to the default `AppWindowTitleBar` that have `ExtendsContentIntoTitleBar == true`, you should not pass it to `null`.

## Motivation and Context
Because it's a bug

## How Has This Been Tested?
Manual.

## Types of changes


marcelwgn (approved):

WinUIGallery/Samples/ControlPages/TitleBarPage.xaml.cs
```diff
@@ -51,14 +51,13 @@ private void SetTitleBar(UIElement titlebar, bool forceCustomTitlebar = false)
         {
             titleBarElement.Visibility = Visibility.Visible;
             window.ExtendsContentIntoTitleBar = true;
-            window.SetTitleBar(titlebar);
+            window.SetTitleBar(titlebar == null? titleBarElement : titlebar);
```
ayushjai30, line 54: Is it necessary to pass the titlebar as a parameter in the SetTitleBar method since we are always retrieving titleBarElement within the method itself? Can we not use titleBarElement directly here?

HO-COOH: Yes, you are correct. It is always retriving the same titlebarElement if needed in all three calls. They can all be removed.

PR #1794, opened 2025-03-06
Enabling Arm64EC Platform for WinUIGallery

kythant:
## Description
WinUI is now building arm64ec config which requires us to setup arm64ec for WinUI Gallery

## Motivation and Context
Arm64ec support will improve performances when using x64 apps run on arm64 devices with WinUI. 

## How Has This Been Tested?
Ran a build in WinUI against this branch

## Types of changes


marcelwgn (approved):
LGTM! 

We probably should also have an ARM pipeline for this repo and verify ARM doesnt break (which isnt very likely anyway)

PR #1789, opened 2025-03-02
Update the spacing in the sample code display of the `ControlExample` control

Zakariathr22:
## Description
This PR eliminates the unnecessary top and bottom space occupied by the sample code in the source code expander. The only change is updating the `Padding` of the `Expander` element to improve layout and content spacing.

## Motivation and Context
For a better and cleaner UX.

## How Has This Been Tested?
Manually tested

## Screenshots (if appropriate):
**Before**
image

**After**
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

Issue #1533, opened 2024-05-05
Request: move CpoyButton lower


Jay-o-Way:
### Issue type

enhancement

### Which version of the app?

WinUI 3 Gallery

### Description

Currently, the CopyButton is usually displayed on one line, next to the related value. However, for the colors, it's displayed in the top-right corner, while the value is on the bottom. It would be nice if the CopyButton sticks immediately to the right of the value everywhere.

### Screenshots

![image](https://github.com/microsoft/WinUI-Gallery/assets/65828559/1fbffb46-a228-4728-a53e-6b9406601901)

### Windows version

Windows 11 23H2 (22631)

### Additional context

Check validity / status for this message.
![image](https://github.com/microsoft/WinUI-Gallery/assets/65828559/104e94ac-cd07-4892-8871-1692a3ef3081)

Jay-o-Way:
I can create a PR for this 💪

niels9001:
Understand the reasoning! But we moved it to the top because the brush names tend to be much longer than the titles. As a result, brush names are either clipped or wrapped, both not great.

karkarl:
Is there a reason why the copy buttons don't only show up on hover? That might solve the issue with longer brush names. I remember reading a Microsoft design guideline that discourages it.

niels9001:
> Is there a reason why the copy buttons don't only show up on hover? That might solve the issue with longer brush names. I remember reading a Microsoft design guideline that discourages it.

That would require adding keyboarding support as well :(

Jay-o-Way:
> we moved it to the top because the brush names tend to be much longer than the titles. As a result, brush names are either clipped or wrapped, both not great.

That argument is not valid in every case:
![image](https://github.com/microsoft/WinUI-Gallery/assets/65828559/7f2e5c46-2c6d-489a-b9e9-139addb7db4f)

Jay-o-Way:
@karkarl can you confirm if the message is outdated and can be removed?

>This brush might not (yet) be available in WinUI

niels9001:
I don't think we cam up with a better solution, so closing this issue for now

PR #1799, opened 2025-03-08
Improve `SampleCodePresenter` layout and `CopyCodeButton` behavior

Zakariathr22:
## Description
This update improves the `SampleCodePresenter` by changing the position of the `CopyCodeButton`. Now, it is in the same column as the `CodePresenter`,making it appear as a floating element instead of occupying unnecessary space. The padding and margins on the left and right sides have also been adjusted. The button’s visibility now changes when the user selects text or moves the horizontal scrollbar.

## Motivation and Context
Before this change, the `CopyCodeButton` was in a different column, which wasted space. Moving it into the same column as the `CodePresenter` makes the layout better and more useful. The button now appears only when needed, making the experience smoother. This design was inspired by GitHub’s code presenter.

- Related to https://github.com/microsoft/WinUI-Gallery/pull/1789#issuecomment-2708231442

## How Has This Been Tested?
- Manually tested different code samples.

#### Before
image
#### After
image
image
#### GitHub behaviour
GH1 GH2
## Types of changes


Jay-o-Way:
@Zakariathr22 I was thinking of the exact same thing! 😁😁 Already created a local branch to try things out! What I learned so far, is that we have to account for the situation when the button is above code, _especially_ in a one-line scenario. Look at the code example in **ColorPage.xaml**

Jay-o-Way:
I was also thinking of simplifying/minimizing the animations (opacity/scale/etc) but that might be a better question for @niels9001 ...

WinUIGallery/Controls/SampleCodePresenter.xaml
```diff
@@ -10,31 +10,27 @@
...
+          CornerRadius="{x:Bind CornerRadius, Mode=OneWay}">
         <ScrollViewer
             x:Name="CodeScrollViewer"
             VerticalAlignment="{x:Bind VerticalContentAlignment, Mode=OneWay}"
             HorizontalScrollBarVisibility="Auto"
             HorizontalScrollMode="Auto"
             VerticalScrollBarVisibility="Auto"
-            VerticalScrollMode="Auto">
+            VerticalScrollMode="Auto"
+            Loaded="CodeScrollViewer_Loaded">
```
Jay-o-Way, line 24: What is this for, exactly?

WinUIGallery/Controls/ControlExample.xaml
```diff
@@ -169,7 +169,7 @@
                     </Grid.RowDefinitions>
                     <SelectorBar
                         x:Name="SelectorBarControl"
-                        Margin="-12,0,0,0"
+                        Margin="4,0,0,0"
```
Jay-o-Way, line 172: - If it were me, I would delete it all together.
- If we change this SelectorBar, should we change similar selector bars? Like on **ColorPage**

WinUIGallery/Controls/SampleCodePresenter.xaml
```diff
@@ -10,31 +10,27 @@
...
             <ContentPresenter
                 x:Name="CodePresenter"
                 Margin="0,0,0,16"
+                Padding="16,0"
```
Jay-o-Way, line 28: How about merging these two lines?

WinUIGallery/Controls/SampleCodePresenter.xaml
```diff
@@ -10,31 +10,27 @@
...
                 Loaded="CodePresenter_Loaded" />
         </ScrollViewer>
         <local:CopyButton
             x:Name="CopyCodeButton"
-            Grid.Column="1"
-            Margin="5,0,0,0"
+            Margin="0,0,8,0"
```
Jay-o-Way, line 33: This doesn't seem like a safe thing. A code example can be one line. We will have to place it at the very top.

Jay-o-Way (commented):
few things

niels9001:
> I was also thinking of simplifying/minimizing the animations (opacity/scale/etc) but that might be a better question for @niels9001 ... Also related to background: would `ControlOnImageFillColorDefaultBrush` make sense now?

I think we need to change the backdrop of the button here, because this will not pass accesibility reviews:

image

Maybe wrap the button in a `Grid` and use the `ControlOnImageFillColorDefaultBrush`?

WinUIGallery/Controls/ControlExample.xaml
```diff
@@ -157,7 +157,7 @@
                 HorizontalContentAlignment="Stretch"
                 Background="{ThemeResource CardBackgroundFillColorSecondaryBrush}"
                 CornerRadius="0,0,8,8"
-                Padding="16,0">
+                Padding="0">
```
niels9001, line 160: Do we need this if the padding is 0? Or is there a default value we are overriding?

niels9001: Nevermind, we are overriding :)

WinUIGallery/Controls/ControlExample.xaml
niels9001: I think we need this to ensure the selectorbar item text and XAML is properly aligned?

image

WinUIGallery/Controls/SampleCodePresenter.xaml.cs
```diff
@@ -233,10 +235,26 @@ private void FormatAndRenderSampleFromString(string sampleString, ContentPresent
         }
         else
         {
-            var sampleCodeRTB = new RichTextBlock { FontFamily = new FontFamily("Consolas") };
+            sampleCodeRTB = new RichTextBlock { FontFamily = new FontFamily("Consolas") };
```
niels9001, line 238: Should we start using `Cascadia Code` with `Consolas` as fallback?

niels9001 (changes requested):
Just a bunch of nits, otherwise this is looking great :)!

WinUIGallery/Controls/ControlExample.xaml
Zakariathr22: > I think we need this to ensure the selectorbar item text and XAML is properly aligned?

Exactly!

Jay-o-Way:
> Maybe wrap the button in a Grid and 

Why add another grid?

WinUIGallery/Controls/SampleCodePresenter.xaml
Zakariathr22: We need to make sure that `CodeScrollViewer` is fully loaded so that its [HorizontalScrollBar](https://github.com/microsoft/microsoft-ui-xaml/blob/589ced40542b9b4de2fddf0e58a1e3eecd164e2f/src/controls/dev/CommonStyles/ScrollViewer_themeresources.xaml#L214C36-L214C55) is part of the visual tree. This ensures that `FindHorizontalScrollBar` can locate it.
The purpose of this is to add the `HorizontalScrollBar_Scroll` event and handle scrollbar dragging to control the visibility of the `CopyCodeButton`.

niels9001:
> > Maybe wrap the button in a Grid and 
> 
> Why add another grid?

Because setting the background would still make it semi-transparent on hover or pressed. Restyling the entire template might be tedious, so simply wrapping the button in a grid (or border) and setting the brush on that grid/border is a faster fix :)!

WinUIGallery/Controls/SampleCodePresenter.xaml.cs
Zakariathr22: I don't think that's necessary because Consolas comes preinstalled with Windows (see the font lists for [Windows 10](https://learn.microsoft.com/en-us/typography/fonts/windows_10_font_list) and [Windows 11](https://learn.microsoft.com/en-us/typography/fonts/windows_11_font_list)). Users will have it on their machines unless they manually delete it.

niels9001: Right. That's why it could be a fallback to Cascadia Code, which is the font that e.g. Terminal uses.

It's a nit though, perfectly fine as is.

WinUIGallery/Controls/SampleCodePresenter.xaml
```diff
@@ -10,36 +10,34 @@
...
             HorizontalScrollBarVisibility="Auto"
             HorizontalScrollMode="Auto"
             VerticalScrollBarVisibility="Auto"
-            VerticalScrollMode="Auto">
+            VerticalScrollMode="Auto"
+            Loaded="CodeScrollViewer_Loaded">
             <ContentPresenter
                 x:Name="CodePresenter"
-                Margin="0,0,0,16"
+                Margin="0,0,0,16" Padding="16,0"
```
Jay-o-Way, line 27: 😆 What I meant was, using one property?

niels9001: Margin="16,0,16,16" or Padding

Zakariathr22: Ah, I get what you mean 😆

niels9001 (approved):

Jay-o-Way:
Did we check keyboard accessibility? I remember this  https://github.com/microsoft/WinUI-Gallery/issues/1533#issuecomment-2115927683...

Zakariathr22:
> Did we check keyboard accessibility?

I think we don’t have a problem with this.

PR #1791, opened 2025-03-06
Limit SettingsPage width for better usability

Zakariathr22:
## Description  
Previously, the settings cards stretched across the entire screen, making them impractical to use on larger displays. Now, the layout is more consistent and user-friendly. Main changes are:

- Replaced `ScrollViewer` with `ScrollView` for improved scrolling behavior  
- Added `MaxWidth` to `PageHeader` and `StackPanel` for layout consistency  
- Removed unnecessary `VisualStateManager` as adaptive padding is handled differently  
- Improved spacing and alignment for better UI consistency  

## Motivation and Context  
The previous design caused settings cards to take up the full screen width, which was not optimal for usability. By setting a max width, the layout aligns with the design approach used in the **Windows Settings app** and **AI Dev Gallery**, ensuring a more structured and familiar experience.  

## How Has This Been Tested?  
- **Manually tested** on different screen sizes to ensure the layout remains consistent.  

## Screenshots (if appropriate):  
**Bofore**
image

**After**
image

## Types of changes  
- [x] Bug fix (non-breaking change which fixes an issue)  
- [ ] New feature (non-breaking change which adds functionality)  
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #1801, opened 2025-03-10
Fix slight blurriness in Settings StackPanel

Zakariathr22:
## Description  
Wrapped the `StackPanel` inside a `Border` in the **Settings** page to fix a slight blurriness issue when the main window is maximized.  

## Motivation and Context  
- This change improves the visual clarity of items inside the `StackPanel`.
- This was a subtle and hard-to-notice side effect caused by #1791.
- It seems that modifying the visual properties of a `StackPanel` inside a `ScrollView` or `ScrollViewer` can cause rendering issues when the window is maximized.
- Wrapping the StackPanel inside a Border or Grid can help stabilize the rendering.

## How Has This Been Tested?  
- Manually tested by maximizing and resizing the window to check for blurriness.  

The difference is subtle and not easily captured in screenshots. The blurriness requires deep observation while maximizing the window to notice it.

## Types of changes  

### DIFF
--- a/WinUIGallery/Pages/SettingsPage.xaml
+++ b/WinUIGallery/Pages/SettingsPage.xaml
@@ -36,112 +36,114 @@
             Padding="36,0,36,0"
             VerticalScrollBarVisibility="Auto"
             VerticalScrollMode="Auto">
-            <StackPanel MaxWidth="1064" Spacing="{StaticResource SettingsCardSpacing}">
-                <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="Appearance &amp; behavior" />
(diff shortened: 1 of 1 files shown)


marcelwgn:
Thank you for this fix. I am inclined to say this is a WinUI issue and not a gallery issue. If you like, could you open an issue on the WinUI repository (and if you happen to find a minimal repro that would be even greater but I think this PR is fine enough for that).

marcelwgn (approved):

Zakariathr22:
> could you open an issue on the WinUI repository (and if you happen to find a minimal repro that would be even greater but I think this PR is fine enough for that).

Done 👍

PR #1811, opened 2025-03-19
Add new windowing APIs

niels9001:
This PR adds the new Min/Max Width/Height APIs for OverlappedPresenter, introduced in WASDK 1.7

image


WinUIGallery/Samples/SamplePages/SampleWindow3.xaml.cs
```diff
@@ -11,17 +11,21 @@ public sealed partial class SampleWindow3 : Window
...
     private OverlappedPresenter presenter;
 
-    public SampleWindow3(bool IsAlwaysOnTop, bool IsMaximizable, bool IsMinimizable, bool IsResizable, bool HasBorder, bool HasTitleBar)
+    public SampleWindow3(bool IsAlwaysOnTop, bool IsMaximizable, bool IsMinimizable, bool IsResizable, bool HasBorder, bool HasTitleBar, int MinWidth, int MinHeight, int MaxWidth, int MaxHeight)
     {
         this.InitializeComponent();
 
         appWindow = GetAppWindowForCurrentWindow();
-
+        appWindow.Resize(new Windows.Graphics.SizeInt32(800, 500));
```
marcelwgn, line 19: Why do we need this resize?

WinUIGallery/Samples/SamplePages/SampleWindow3.xaml.cs
```diff
@@ -11,17 +11,21 @@ public sealed partial class SampleWindow3 : Window
...
         presenter = OverlappedPresenter.Create();
         presenter.IsAlwaysOnTop = IsAlwaysOnTop;
         presenter.IsMaximizable = IsMaximizable;
         presenter.IsMinimizable = IsMinimizable;
         presenter.IsResizable = IsResizable;
+        presenter.PreferredMinimumWidth = MinWidth;
+        presenter.PreferredMinimumHeight = MinHeight;
+        presenter.PreferredMaximumWidth = MaxWidth;
+        presenter.PreferredMaximumHeight = MaxHeight;
```
Zakariathr22, line 28: I think we should implement this in a new sample window and set its `IsMaximizable` property to `false`. Additionally, we should add a note explaining that this is the best practice when setting a maximum width or height.  

If we set a maximum value for the width or height of the `AppWindowPresenter` and try to maximize the window, it will result in **unwanted behavior in the layout**.
image

Zakariathr22 (changes requested):

WinUIGallery/Samples/SamplePages/SampleWindow3.xaml.cs
niels9001: Greeeeeat catch. Yeah, that makes sense! Should we keep the MinWidth/Height in this sample? it feels to me something that most developers would want vs the MaxWidth/Height?

Zakariathr22: Wouldn't it make more sense to keep both `MinWidth/Height` and `MaxWidth/Height` together?

niels9001: I've created a new sample per your suggestion.. renamed a bunch of files because it felt better to have this sample with the other overlapped presenters as I think this is a common use case. Please do another pass on the PR whenever you have time 😀

WinUIGallery/Samples/ControlPages/AppWindowPage.xaml
```diff
@@ -328,72 +328,160 @@ public sealed partial class SampleWindow3 : Window
...
+                <controls:ControlExampleSubstitution Key="IsMaximizable" Value="{x:Bind BoolToLowerString(IsMaximizable.IsOn), Mode=OneWay}" />
+                <controls:ControlExampleSubstitution Key="IsMinimizable" Value="{x:Bind BoolToLowerString(IsMinimizable.IsOn), Mode=OneWay}" />
+                <controls:ControlExampleSubstitution Key="IsResizable" Value="{x:Bind BoolToLowerString(IsResizable.IsOn), Mode=OneWay}" />
+                <controls:ControlExampleSubstitution Key="HasBorder" Value="{x:Bind BoolToLowerString(HasBorder.IsOn), Mode=OneWay}" />
+                <controls:ControlExampleSubstitution Key="HasTitleBar" Value="{x:Bind BoolToLowerString(HasTitleBar.IsOn), Mode=OneWay}" />
+            </controls:ControlExample.Substitutions>
+        </controls:ControlExample>
+
+
+        <controls:ControlExample HeaderText="Setting the minimum and maximum width / height on an AppWindow" XamlSource="AppWindow\AppWindowSample3_xaml.txt">
```
Zakariathr22, line 342:
```suggestion
        <controls:ControlExample HeaderText="Setting the minimum and maximum width / height on an AppWindow using OverlappedPresenter" XamlSource="AppWindow\AppWindowSample3_xaml.txt">
```
Minimum and maximum width/height are properties of the `OverlappedPresenter` class, so it would be better to point to it in the example header.

Zakariathr22 (approved):
Just a small suggestion. Otherwise, great work!

WinUIGallery/Samples/ControlPages/AppWindowPage.xaml
```diff
@@ -151,119 +153,117 @@ public sealed partial class SampleWindow1 : Window
...
+                                with <Run FontWeight="SemiBold">OverlappedPresenter</Run>
+                                , if the title bar
                                 (<Run FontWeight="SemiBold">HasTitleBar = true</Run>) is enabled, the window must have a border
-                                (<Run FontWeight="SemiBold">HasBorder = true</Run>). Setting
-                                <Run FontWeight="SemiBold">HasBorder</Run> to
-                                <Run FontWeight="SemiBold">false</Run> while
-                                <Run FontWeight="SemiBold">HasTitleBar</Run> is
-                                <Run FontWeight="SemiBold">true</Run> will result in a fatal error.
-                            </Paragraph>
+                                (<Run FontWeight="SemiBold">HasBorder = true</Run>). Setting<Run FontWeight="SemiBold">HasBorder</Run>
```
marcelwgn, line 195:
```suggestion
                                (<Run FontWeight="SemiBold">HasBorder = true</Run>). Setting <Run FontWeight="SemiBold">HasBorder</Run>
```

marcelwgn (approved):

PR #1812, opened 2025-03-20
Fix issue with Copy Button text scaling

michael-hawker:
## Description

Related to https://github.com/microsoft/microsoft-ui-xaml/pull/9644

Current with text scaling:
image

With change, it stays the same (like the other buttons do).

## Types of changes


niels9001 (approved):
Great catch!

PR #1813, opened 2025-03-22
Add RichEditBox Math Mode Examples

Zakariathr22:
## **Description**  
This PR adds new examples demonstrating the **Math Mode** features in **RichEditBox**.  

## **Motivation and Context**  
Helping developers understand how to use math mode. 

## **How Has This Been Tested?**  
Manually tested.

## **Screenshots (if appropriate):**  
image
image

## **Types of Changes**  


Zakariathr22:
@niels9001 @marcelwgn @Hemantxk

Before going further with this PR, could you review it? The remaining work is mostly repetitive, and the main ideas are already in place. Do I have the green light to complete it?

niels9001:
> @niels9001 @marcelwgn @Hemantxk
> 
> Before going further with this PR, could you review it? The remaining work is mostly repetitive, and the main ideas are already in place. Do I have the green light to complete it?

I think it looks great :)! Please go ahead!

WinUIGallery/Samples/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -29,10 +29,17 @@ public sealed partial class RichEditBoxPage : Page
...
+    ObservableCollection<MathSymbol> SymbolsCollection = MathModeHelper.GetSymbolsCollection();
+    ObservableCollection<MathStucture> StructuresCollection = MathModeHelper.GetStructuresCollection();
+    ObservableCollection<MathStucture> IntegralsCollection = MathModeHelper.GetIntegralsCollection();
+    ObservableCollection<MathStucture> LargeOperators = MathModeHelper.GetLargeOperatorsCollection();
     public RichEditBoxPage()
     {
         this.InitializeComponent();
+
+        MathEditor.TextDocument.SetMathMode(RichEditMathMode.MathOnly);
+        MathModeDescription.Text = "Math mode enables users to have input automatically recognized and converted to MathML while being received.\r\nFor example, \"4^2\" is converted to \"4\u00b2\", and \"\\pi\" is converted to \"\u03c0\".\r\nMath mode might change formatting (fonts), context menus, and other aspects of the input.";
```
Hemantxk, line 41: Suggestion:
Math mode might change formatting (fonts), context menus, and other aspects of the input.
Modify this to:
Enabling math mode in a RichEditBox automatically switches the input font to Cambria Math. Additionally, toggling math mode clears any existing content and undo stack.

WinUIGallery/Samples/ControlPages/RichEditBoxPage.xaml
```diff
@@ -206,5 +208,265 @@
             </RelativePanel>
         </controls:ControlExample>
 
+        <controls:ControlExample x:Name="MathModeExample" HeaderText="Rich edit text document in math mode">
```
Hemantxk, line 211: Rich edit box* in math mode

WinUIGallery/Samples/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -230,4 +237,159 @@ private void Editor_TextChanged(object sender, RoutedEventArgs e)
...
+                MathModeHelper.TypeCommand(insertionBtn.Tag.ToString());
+            }
+        }
+    }
+
+    private void SelectorBar_SelectionChanged(SelectorBar sender, SelectorBarSelectionChangedEventArgs args)
+    {
+        if(sender is SelectorBar selectorBar)
+        {
+            if (selectorBar.SelectedItem.Tag.ToString() == "Symbols")
```
Saharsh979, line 278: You're calling .ToString() on selectorBar.SelectedItem.Tag without checking if SelectedItem or Tag is null.

Hemantxk:
Thank you! Really appreciate the efforts you put into checking in these examples. Can you also attach few more screenshots of the new sections from different pages? Structures/Integrals/Accent etc.

WinUIGallery/Helpers/MathModeHelper.cs
```diff
@@ -0,0 +1,261 @@
...
+
+    // Import the necessary functions from user32.dll to manipulate keyboard input.
+    [DllImport("user32.dll")]
+    private static extern IntPtr LoadKeyboardLayout(string pwszKLID, uint Flags);
+
+    [DllImport("user32.dll")]
+    private static extern short VkKeyScanA(char ch); // Converts a character to its virtual-key code for the current keyboard layout.
+
+    [DllImport("user32.dll", SetLastError = true)]
+    private static extern void keybd_event(byte bVk, byte bScan, uint dwFlags, int dwExtraInfo);
```
Hemantxk, line 175: At some point we should switch to SendInput (Not necessarily now)

Hemantxk (approved):

marcelwgn (changes requested):
When pasting a character in using the button below the sample showcases, the keyboard layout of my OS changes, this should definitely not happen.

Also not sure if we need to document MathML in this app but Ill let others chime in.

Zakariathr22:
> Can you also attach few more screenshots of the new sections from different pages? Structures/Integrals/Accent etc.

Done!

WinUIGallery/Helpers/MathModeHelper.cs
Zakariathr22: Yes, I know. `SendInput` is the recommended way over `keybd_event`, but `keybd_event` is not deprecated and still works well. So, I think we can switch to `SendInput` when it becomes necessary in the future.

Zakariathr22:
> When pasting a character in using the button below the sample showcases, the keyboard layout of my OS changes,

The `TypeCommand` function simulates typing a command character by character using virtual keyboard events. This is necessary because math formulas are rendered dynamically as they are typed (setting text in a `RichEditBox` would not trigger formula rendering).  

For this reason, changing the keyboard layout is necessary. If the user is using a different keyboard layout (e.g., AZERTY or Arabic ...etc), the input would be affected by the current layout.

> this should definitely not happen.

Actually, using `LoadKeyboardLayout` is fine and poses no risks. Additionally, for languages that do not use Latin letters, users would likely switch their keyboard layout manually to type mathematical formulas anyway.

> Also not sure if we need to document MathML in this app but Ill let others chime in.

Yes, you're right! we don't need to document Math Markup Language since [the official documentation](https://www.w3.org/TR/MathML2/) is sufficient.  

My plan is to showcase two examples:  
1. The first example focuses on typing in math mode using [Unicode Nearly Plain-Text Encoding of Mathematics](https://unicode.org/notes/tn28/UTN28-PlainTextMath-v2.pdf) (the current under development example). This approach is widely used by Microsoft, including in Office.  
2. The second example will demonstrate the `SetMathML` method.  
**Example:**  
Passing the following string to `SetMathML` will render:  
*x³ + y > 5*  
```xml
<mml:math xmlns:mml="http://www.w3.org/1998/Math/MathML" display="block">
    <mml:msup>
        <mml:mrow>
            <mml:mi mathcolor="#000000">x</mml:mi>
        </mml:mrow>
        <mml:mrow>
            <mml:mn mathcolor="#000000">3</mml:mn>
        </mml:mrow>
    </mml:msup>
    <mml:mo mathcolor="#000000">+</mml:mo>
    <mml:mi mathcolor="#000000">y</mml:mi>
    <mml:mo mathcolor="#000000">&#x3E;</mml:mo>
    <mml:mn mathcolor="#000000">5</mml:mn>
</mml:math>
```
This will be a simple example, and we can direct users to the official MathML documentation for more details.

---
These are my thoughts, and any suggestions for changes would be very welcome! 😊

marcelwgn:
> > this should definitely not happen.
> 
> Actually, using `LoadKeyboardLayout` is fine and poses no risks. Additionally, for languages that do not use Latin letters, users would likely switch their keyboard layout manually to type mathematical formulas anyway.

I disagree. Changing the users keyboard layout without them knowing is not good. And well I use a German keyboard layout, if I click on the button, it switches to English. I do not need to switch to English to type the equation. But suddenly realizing that some keys now type a different character is a bad experience.

-----

Regarding the samples, I missunderstood what MathML is. Your approach to showcasing that sample is good. My concern was about showcasing all of the `Unicode Nearly Plain-Text Encoding of Mathematics` macros. I would prefer the gallery not duplicating existing documentation, especially if its largely tables anyone hardly completely needs. @niels9001 @Hemantxk @Saharsh979 what are your thoughts?

Zakariathr22:
> its largely tables anyone hardly completely needs.

After thinking...
Yeah, this makes sense. I hadn’t thought about it that way. How often do we actually see applications using math mode? Maybe we should keep both examples simple and just direct the user to documentation for more details.

Zakariathr22:
> I disagree. Changing the users keyboard layout without them knowing is not good. And well I use a German keyboard layout, if I click on the button, it switches to English. I do not need to switch to English to type the equation. But suddenly realizing that some keys now type a different character is a bad experience.

@marcelwgn 
By this, this is fixed now! The keyboard layout no longer changes, and input is simulated correctly using `SendInput`. You can check out [Raymond Chen's blog](https://devblogs.microsoft.com/oldnewthing/20250319-00/?p=110979) about simulating keyboard input for more insights on handling it properly.  

Thanks for your review! Honestly, after this fix, I look at the previous method as a disaster 😆. If you have a moment, could you review the new approach and share your thoughts? 😊

> I would prefer the gallery not duplicating existing documentation, especially if its largely tables anyone hardly completely needs.

And about this, I am so confused what to do, I'll move on to the second example of `SetMathML` and `GetMathML` in `RichEditBox` until others provide their feedback on it.

WinUIGallery/Helpers/MathModeHelper.cs
Zakariathr22: Done! Actually, this fixed a complex issue where switching the keyboard layout caused a bad user experience.

Hemantxk:
@Zakariathr22 it's good that we are showcasing the next level functionality here, but I would like to keep the performance in-check too when loading so many images in memory. I hope we are not seeing too much perf regression with having these many examples. Like to hear your opinion too @niels9001 @marcelwgn whether we should go ahead with current version or trim down the number of samples.

Zakariathr22:
> @Zakariathr22 it's good that we are showcasing the next level functionality here, but I would like to keep the performance in-check too when loading so many images in memory. I hope we are not seeing too much perf regression with having these many examples. Like to hear your opinion too @niels9001 @marcelwgn whether we should go ahead with current version or trim down the number of samples.

You're right, maybe it's a bit too many assets. However, in terms of performance, the total size of all images is currently 280KB (less than 300KB).
image

I chose this approach because I was between two options:
- either using `RichEditBox` in math mode with `GetMathML` with each math structure, which is a very heavy control that would impact performance when loading many instances at once (also requiring extra work for async loading),
- **or** using lightweight images since this is static data, I went with it in the first place for optimization.

I believe the user experience remains smooth, especially with `ItemsRepeater`, which optimizes performance by only creating UI elements for visible items and reusing them when scrolling.  

That said, if you think trimming down the samples would still be necessary, I’m open to discussing potential adjustments!

Zakariathr22:
> [!NOTE]
> 
> This PR is ready for review.  
>
> The first example highlights the most commonly used cases in mathematical equations. While the Unicode NPT documentation covers more advanced concepts, this example is not a duplication but rather a simplified demonstration of the fundamentals. Additionally, I have ensured that the example runs as smoothly as possible for optimal performance.

WinUIGallery/Helpers/MathModeHelper.cs
```diff
@@ -0,0 +1,469 @@
...
+    /// Sends a sequence of Unicode characters individually.
+    /// </summary>
+    /// <param name="unicodeText">The Unicode string to be sent character by character.</param>
+    /// <remarks>
+    /// This method sends each character in the provided Unicode string using <c>SendUnicodeCharacter()</c>.  
+    /// 
+    /// - Intended for feeding Unicode input to text-processing components.  
+    /// - Can be used for entering structured expressions like math formulas.  
+    /// </remarks>
+    public static void SendUnicodeText(string unicodeText)
```
Saharsh979, line 462: SendUnicodeCharacter throws exceptions on failure, but SendUnicodeText doesn't handle these exceptions - consider adding try/catch blocks

Saharsh979 (approved):

WinUIGallery/Samples/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -230,4 +240,221 @@ private void Editor_TextChanged(object sender, RoutedEventArgs e)
...
+        if (sender is Button insertionBtn)
+        {
+            if (MathEditor == null) return;
+
+            MathEditor.Focus(FocusState.Programmatic);
+
+            var selection = MathEditor.Document.Selection;
+            if (selection != null)
+            {
+                InputSender.SendUnicodeText(insertionBtn.Tag.ToString());
```
Hemantxk, line 272: From here how does the keyboard inputs lands in the desired RichEditBox as there might be more TextBoxes in the window?

Zakariathr22: In this case, we assume that the math input is intended for a single `RichEditBox`. Before sending the input, we set focus to it using `MathEditor.Focus(FocusState.Programmatic);`.

However, if there are multiple `RichEditBox` controls in the window, the user is expected to manually set focus to the desired one. The code then uses the `GotFocus` and `LostFocus` events to check whether the currently focused `RichEditBox` is in **math mode**, in order to adjust the visibility or enabled state of the symbols and structures tables accordingly.

Maybe there are better ways to handle this, but this was the first idea that came to mind.

Zakariathr22: > are you suggesting that this could be done as above if needed?

Yes, this is exactly what I meant.

Zakariathr22:
> Thank you @Zakariathr22 for your continuously keeping this PR up to date! As of right now, I think the best direction right now would be to not include the table of symbols due to its complexity. Otherwise the PR looks good to me 😊

@marcelwgn Just to clarify, did you mean removing just this table, or the entire section (symbols and structures)?
image

marcelwgn:
I think we should keep the samples but get rid of the tables altogether. I know you worked hard on this and you did an excellent job so I am sorry we are taking this turn but I think right now this is the best direction for these samples.

Zakariathr22:
> I think we should keep the samples but get rid of the tables altogether. I know you worked hard on this and you did an excellent job so I am sorry we are taking this turn but I think right now this is the best direction for these samples.

I think GitHub should really add a 💔 reaction for moments like this 😄
Just wish the decision came **a bit earlier** before I fixed this 😂

> When pasting a character in using the button below the sample showcases, the keyboard layout of my OS changes, this should definitely not happen.

But to be fair, your suggestion makes sense and I’ll go with it 👍

WinUIGallery/Samples/ControlPages/RichEditBoxPage.xaml.cs
```diff
@@ -230,4 +233,60 @@ private void Editor_TextChanged(object sender, RoutedEventArgs e)
...
+            "      <mml:mi mathcolor=\"#000000\">A</mml:mi>\r\n" +
+            "    </mml:mrow>\r\n" +
+            "  </mml:mfenced>\r\n" +
+            "  <mml:mo mathcolor=\"#000000\">\u2194</mml:mo>\r\n" +
+            "  <mml:mi mathcolor=\"#000000\">x</mml:mi>\r\n" +
+            "  <mml:mo mathcolor=\"#000000\">\u2286</mml:mo>\r\n" +
+            "  <mml:mi mathcolor=\"#000000\">A</mml:mi>\r\n" +
+            "</mml:math>";
+
+        if(mathEditor2.ActualTheme == ElementTheme.Dark)
```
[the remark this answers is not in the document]
Zakariathr22: > I wish this wasnt necessary but I guess this is the way it is. I hope it wasnt too difficult to figure this out ^^

Yeah, this is the only way I could figure out to do it.

I initially thought it was an issue in the `SetMathML` method and that it should automatically update the color from white to black, until I realized we can't know whether the color was set by the user using the `\color` Unicode NPT or not. 

So I guess this is the simplest way to handle it in our use case.

PR #1821, opened 2025-03-27
Added the context menu for the tab view for left and right actions.

sabareesanms:
**Bug** [52757718](https://dev.azure.com/microsoft/OS/_workitems/edit/52757718) Lack of Alternative Single Pointer Method for Tab Movement in TabView Screen.

**Fix** - Adding context menu for the TabView example in the WinUI Gallery which allows the tabs to be moved left and right with single click in addition to the dragging method.

**Testing** - Tested in the WinUI Gallery app. 

**Screenshots** 

Screenshot 2025-03-27 123727

**Changes** 

1. **WinUIGallery/Helpers/TabViewHelper.cs** - Added a function PopulateTabViewContextMenu(MenuFlyout contextMenu) 
- Gets the parent tab view and tab view item of the context menu.
- Add left move menu if index > 0 with its action (Adds except 1st item)
- Add right move menu if index < count -1   (Adds except last item)
- Closes the menu if the context menu is empty (To avoid empty context menu showing). 

2. **WinUIGallery/Samples/ControlPages/TabViewPage.xaml**
 - Defines the context menu **TabViewContextMenu** in page resource for reusability
 - Adds the context menu to all tabviewitem in the example page. 

3. **WinUIGallery/Samples/ControlPages/TabViewPage.xaml.cs**
 - Added the event handler for the context menu opening event to populate the context menu with the tab view helper function

### DIFF
--- a/WinUIGallery/Helpers/TabViewHelper.cs
+++ b/WinUIGallery/Helpers/TabViewHelper.cs
@@ -7,11 +7,15 @@ namespace WinUIGallery.Helpers;
 
 public static class TabViewHelper
 {
-    public static void PopulateTabViewContextMenu(MenuFlyout contextMenu)
+    public static void PopulateTabViewContextMenu(MenuFlyout menuFlyout)
     {
-        contextMenu.Items.Clear();
+        menuFlyout.Items.Clear();
 
-        var item = (TabViewItem)contextMenu.Target;
+        var item = (TabViewItem)menuFlyout.Target;
+        if(item == null)
+        {
+            return;
+        }
         ListView tabViewListView = null;
         TabView tabView = null;
 
@@ -69,7 +73,7 @@ public static class TabViewHelper
                     tabView.TabItems.Insert(index - 1, item);
                 }
             };
-            contextMenu.Items.Add(moveLeftItem);
+            menuFlyout.Items.Add(moveLeftItem);
         }
 
         if (index < tabViewListView.Items.Count - 1)
@@ -90,13 +94,7 @@ public static class TabViewHelper
                     tabView.TabItems.Insert(index + 1, item);
                 }
             };
-            contextMenu.Items.Add(moveRightItem);
-        }
-
-        // If the context menu ended up with no items at all, then we'll prevent it from being shown.
-        if (contextMenu.Items.Count == 0)
-        {
-            contextMenu.Hide();
+            menuFlyout.Items.Add(moveRightItem);
         }
     }
 }
--- a/WinUIGallery/Helpers/TabViewHelper.cs
+++ b/WinUIGallery/Helpers/TabViewHelper.cs
@@ -1,0 +1,102 @@
+using System.Collections;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+
+namespace WinUIGallery.Helpers;
+
+public static class TabViewHelper
+{
+    public static void PopulateTabViewContextMenu(MenuFlyout contextMenu)
+    {
+        contextMenu.Items.Clear();
+
+        var item = (TabViewItem)contextMenu.Target;
+        ListView tabViewListView = null;
+        TabView tabView = null;
+
+        DependencyObject current = item;
+
+        while (current != null)
+        {
+            DependencyObject parent = VisualTreeHelper.GetParent(current);
+
+            if (parent is ListView parentTabViewListView)
+            {
+                tabViewListView = parentTabViewListView;
+            }
+            else if (parent is TabView parentTabView)
+            {
+                tabView = parentTabView;
+            }
+
+            if (tabViewListView != null && tabView != null)
+            {
+                break;
+            }
+
+            current = parent;
+        }
+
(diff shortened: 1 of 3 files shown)


WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    var item = tabView.TabItems[index];
+                    tabView.TabItems.RemoveAt(index);
+                    tabView.TabItems.Insert(index + 1, item);
+                }
+            };
+            contextMenu.Items.Add(moveRightItem);
+        }
+
+        // If the context menu ended up with no items at all, then we'll prevent it from being shown.
```
godlytalias, line 96: This logic would be better to be at the caller side only right?

godlytalias (approved):

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+        int index = tabViewListView.IndexFromContainer(item);
+
+        if (index > 0)
+        {
+            MenuFlyoutItem moveLeftItem = new() { Text = "Move tab left" };
+            moveLeftItem.Click += (s, args) =>
+            {
+                if (tabView.TabItemsSource is IList itemsSourceList)
+                {
+                    var item = itemsSourceList[index];
```
Hemantxk, line 61: Does index value gets copied over in this lambda for later use?

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+                    tabView.TabItems.RemoveAt(index);
+                    tabView.TabItems.Insert(index - 1, item);
+                }
+            };
+            contextMenu.Items.Add(moveLeftItem);
+        }
+
+        if (index < tabViewListView.Items.Count - 1)
+        {
+            MenuFlyoutItem moveRightItem = new() { Text = "Move tab right" };
```
Hemantxk, line 77: How about adding items like "Move the tab to the end" or "Move the tab to beginning"?

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+                    tabView.TabItems.Insert(index + 1, item);
+                }
+            };
+            contextMenu.Items.Add(moveRightItem);
+        }
+
+        // If the context menu ended up with no items at all, then we'll prevent it from being shown.
+        if (contextMenu.Items.Count == 0)
+        {
+            contextMenu.Hide();
```
Hemantxk, line 99: As my comment on previous PR: this should be the job of caller.

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace WinUIGallery.Helpers;
+
+public static class TabViewHelper
+{
+    public static void PopulateTabViewContextMenu(MenuFlyout contextMenu)
+    {
+        contextMenu.Items.Clear();
+
+        var item = (TabViewItem)contextMenu.Target;
```
Hemantxk, line 14: We can have a null check after this point, if `item` is null return.

WinUIGallery/Helpers/TabViewHelper.cs
```diff
@@ -0,0 +1,102 @@
+using System.Collections;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+
+namespace WinUIGallery.Helpers;
+
+public static class TabViewHelper
+{
+    public static void PopulateTabViewContextMenu(MenuFlyout contextMenu)
```
Hemantxk, line 10: should we rename this param -> menuFlyout?

WinUIGallery/Helpers/TabViewHelper.cs
sabareesanms: yes. it will get copied over.

WinUIGallery/Helpers/TabViewHelper.cs
sabareesanms: We added these similar to edge and PowerShell windows. It is good use case. We can add this based on requirements. @Hemantxk

WinUIGallery/Helpers/TabViewHelper.cs
sabareesanms: Updated

WinUIGallery/Helpers/TabViewHelper.cs
sabareesanms: Added a null check

WinUIGallery/Helpers/TabViewHelper.cs
sabareesanms: Renamed the param

Hemantxk (approved):

Saharsh979:
I see that the pipeline is failing because of missing dependencies, turns out we were looking for a newer version of some package which was not there on the feed. Have updated the feed now!

PR #1837, opened 2025-04-17
add warning to AnimatedIcon page

manabu-nakamura:
## Description

Add the warning to the AnimatedIcon page until https://github.com/microsoft/microsoft-ui-xaml/issues/10243 is resolved.

## Motivation and Context

AnimatedChevronRightDownSmallVisualSource and AnimatedChevronUpDownSmallVisualSource don't work properly (https://github.com/microsoft/microsoft-ui-xaml/issues/10243). Add the warning to the AnimatedIcon page until https://github.com/microsoft/microsoft-ui-xaml/issues/10243 is resolved.

## How Has This Been Tested?

The WinUI-Gallery was built and run.
s

s

## Types of changes

- [x] Other

### DIFF
--- a/WinUIGallery/Samples/ControlPages/AnimatedIconPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/AnimatedIconPage.xaml
@@ -22,7 +22,10 @@
                             Lottie-Windows
                         </Hyperlink>
                         .
-                        For guidance on how to properly structure your animation file see the AnimatedIcon Guidance page.<LineBreak />
+                        For guidance on how to properly structure your animation file see the AnimatedIcon Guidance page.
+                        Warning: <Hyperlink NavigateUri="https://github.com/microsoft/microsoft-ui-xaml/issues/10243">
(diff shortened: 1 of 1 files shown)


Zakariathr22:
@manabu-nakamura 
Good idea, and I see what you're going for. That said, I think we might not need to warn users about an issue they’re already seeing, it doesn’t really offer a solution, just points to the problem with more detail. Maybe we could focus on improving the experience instead?

manabu-nakamura:
> @manabu-nakamura Good idea, and I see what you're going for. That said, I think we might not need to warn users about an issue they’re already seeing, it doesn’t really offer a solution, just points to the problem with more detail. Maybe we could focus on improving the experience instead?

I wasted a lot of time until I came across https://github.com/microsoft/microsoft-ui-xaml/issues/10243. The best solution is to fix https://github.com/microsoft/microsoft-ui-xaml/issues/10243.

manabu-nakamura:
What an unkind demonstration! What is the purpose of this demonstration? Who is this demonstration for?

PR #1838, opened 2025-04-17
Update `VersionHelper` for improved SDK version handling

Zakariathr22:
## Description
This PR simplifies SDK versioning by using `ReleaseInfo` and `RuntimeInfo` (new APIs introduced in WinAppSDK 1.7), main changes are:
- Updated `WinAppSdkDetails` to use `ReleaseInfo.AsString`.
- Simplified `WinAppSdkRuntimeDetails` to directly use `RuntimeInfo.AsString`, removing the previous dynamic retrieval logic.

## Motivation and Context
Providing both cleaner code and more complete version details.

## How Has This Been Tested?
Manually tested

## Types of changes

### DIFF
--- a/WinUIGallery/Helpers/VersionHelper.cs
+++ b/WinUIGallery/Helpers/VersionHelper.cs
@@ -3,9 +3,8 @@ using Microsoft.Windows.ApplicationModel.WindowsAppRuntime;
 namespace WinUIGallery.Helpers;
 internal static class VersionHelper
 {
-    public static string WinAppSdkDetails =>
-        $"Windows App SDK {ReleaseInfo.AsString}" +
-        (string.IsNullOrEmpty(ReleaseInfo.VersionTag) ? "" : $" ({ReleaseInfo.VersionTag})");
+    public static string WinAppSdkDetails => 
+        $"Windows App SDK {ReleaseInfo.AsString}";
 
     public static string WinAppSdkRuntimeDetails => 
         WinAppSdkDetails + $", Windows App Runtime {RuntimeInfo.AsString}";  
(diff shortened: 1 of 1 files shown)


WinUIGallery/Helpers/VersionHelper.cs
```diff
@@ -1,35 +1,11 @@
...
 namespace WinUIGallery.Helpers;
 internal static class VersionHelper
 {
-    public static string WinAppSdkDetails
-    {
-        // TODO: restore patch number and version tag when WinAppSDK supports them both
-        get => string.Format("Windows App SDK {0}.{1}",
-            WASDK.Release.Major, WASDK.Release.Minor);
-    }
+    public static string WinAppSdkDetails =>
```
marcelwgn, line 6: I think we should stick with the template string here.

marcelwgn (commented):
Excellent idea, thank you for cleaning this up!

WinUIGallery/Helpers/VersionHelper.cs
Zakariathr22: > I think we should stick with the template string here. I dont think it is helpful to know it is WASDK 1.7.0, 1.7 is enough information.

@marcelwgn I included it because it was part of the TODO comment.
```cs
// TODO: restore patch number and version tag when WinAppSDK supports them both
get => string.Format("Windows App SDK {0}.{1}",
    WASDK.Release.Major, WASDK.Release.Minor);
```
The `ReleaseInfo.AsString` method returns the full release info in the format `major.minor.patch[-tag]`, for example `"1.7.0-experimental1"`. If there's no tag, it just returns `"1.7.0"`.

Do you want me to keep the tag, or drop it also and just show `"1.7"`?

niels9001 (approved):
Amazing! Tested, and works!

PR #1856, opened 2025-04-27
add border to icon

### LINKED ISSUE #1854: Idea: use a border or background on icon
Many icons are used as an overlay, where the content is in one of the corners. To show where inside the area the content is placed, we could choose to use a border and/or a background for the icon in the SidePanel. It would help users see and find the right icon for the right situation.

Example of a border, a background and both:

Image Image Image

Further, it would show that a few icons seem to have a variable aspect ratio - they're not square. I saw a couple related to Punctuation.

Image Image

Jay-o-Way:
## Description

Adds a border and background to the larger icon in the SidePanel. This shows how any "overlay icons" are placed inside the available area.

Closes #1854

## Motivation and Context

Clarity. May help finding the right icon.

image image

## Types of changes

### DIFF
--- a/WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
@@ -109,7 +109,8 @@
                             How to get the font
                         </Paragraph>
                         <Paragraph>
-                            Segoe Fluent Icons comes preinstalled on Windows 11. On Windows 10, it requires a manual installation. You can download it <Hyperlink NavigateUri="https://learn.microsoft.com/windows/apps/design/downloads/#fonts">here</Hyperlink>.</Paragraph>
+                            Segoe Fluent Icons comes preinstalled on Windows 11. On Windows 10, it requires a manual installation. You can download it <Hyperlink NavigateUri="https://learn.microsoft.com/windows/apps/design/downloads/#fonts">here</Hyperlink>.
+                        </Paragraph>
                     </RichTextBlock>
                 </toolkit:SettingsCard>
                 <toolkit:SettingsCard
@@ -186,55 +187,58 @@
                     Grid.Column="1"
                     Padding="16,16,8,16"
                     Background="{ThemeResource CardBackgroundFillColorDefaultBrush}"
-                    BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
+                    BorderBrush="{ThemeResource DividerStrokeColorDefaultBrush}"
                     BorderThickness="1,0,0,0"
                     DataContext="{x:Bind}"
                     Visibility="Collapsed">
                     <StackPanel Spacing="2">
-                        <FontIcon
+                        <Border
                             Margin="0,0,0,24"
                             HorizontalAlignment="Left"
-                            VerticalAlignment="Center"
-                            FontFamily="{StaticResource SymbolThemeFontFamily}"
-                            FontSize="48"
-                            Glyph="{Binding SelectedItem.Character, Mode=OneWay}" />
+                            BorderBrush="{ThemeResource ControlStrokeColorDefaultBrush}"
+                            BorderThickness="1">
+                            <FontIcon
+                                FontFamily="{StaticResource SymbolThemeFontFamily}"
+                                FontSize="48"
+                                Glyph="{Binding SelectedItem.Character, Mode=OneWay}" />
+                        </Border>
                         <TextBlock
-                            FontSize="12"
+                            Style="{StaticResource CaptionTextBlockStyle}"
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="Icon name" />
                         <controls:SampleCodePresenter Code="{Binding SelectedItem.Name, Mode=OneWay}" Style="{StaticResource CodeValuePresenterStyle}" />
                         <TextBlock
-                            FontSize="12"
+                            Style="{StaticResource CaptionTextBlockStyle}"
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="Text glyph" />
                         <controls:SampleCodePresenter Code="{Binding SelectedItem.TextGlyph, Mode=OneWay}" Style="{StaticResource CodeValuePresenterStyle}" />
                         <TextBlock
-                            FontSize="12"
+                            Style="{StaticResource CaptionTextBlockStyle}"
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="Code glyph" />
                         <controls:SampleCodePresenter Code="{Binding SelectedItem.CodeGlyph, Mode=OneWay}" Style="{StaticResource CodeValuePresenterStyle}" />
                         <TextBlock
-                            FontSize="12"
+                            Style="{StaticResource CaptionTextBlockStyle}"
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="XAML" />
                         <controls:SampleCodePresenter x:Name="XAMLCodePresenter" Style="{StaticResource CodeValuePresenterStyle}" />
                         <TextBlock
-                            FontSize="12"
+                            Style="{StaticResource CaptionTextBlockStyle}"
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="C#" />
                         <controls:SampleCodePresenter x:Name="CSharpCodePresenter" Style="{StaticResource CodeValuePresenterStyle}" />
                         <TextBlock
                             Margin="0,4,0,0"
-                            FontSize="12"
+                            Style="{StaticResource CaptionTextBlockStyle}"
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="Tags" />
                         <ItemsView
                             x:Name="TagsItemsView"
                             Margin="0,8,0,4"
-                            x:Load="{x:Bind SelectedItem.Tags, Mode=OneWay, Converter={StaticResource emptyCollectionToFalseConverter}}"
+                            x:Load="{x:Bind SelectedItem.Tags, Converter={StaticResource emptyCollectionToFalseConverter}}"
                             IsItemInvokedEnabled="True"
                             ItemInvoked="TagsItemsView_ItemInvoked"
-                            ItemsSource="{x:Bind SelectedItem.Tags, Mode=OneWay}"
+                            ItemsSource="{x:Bind SelectedItem.Tags}"
                             SelectionMode="None">
                             <ItemsView.Layout>
                                 <toolkit:WrapLayout HorizontalSpacing="4" VerticalSpacing="4" />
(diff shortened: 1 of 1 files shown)


WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -186,55 +187,59 @@
...
+                        <Border
                             Margin="0,0,0,24"
                             HorizontalAlignment="Left"
-                            VerticalAlignment="Center"
-                            FontFamily="{StaticResource SymbolThemeFontFamily}"
-                            FontSize="48"
-                            Glyph="{Binding SelectedItem.Character, Mode=OneWay}" />
+                            Background="{ThemeResource ControlFillColorDefaultBrush}"
+                            BorderBrush="{ThemeResource ControlStrokeColorDefaultBrush}"
+                            BorderThickness="1">
```
niels9001, line 200: We should add a 4px CornerRadius

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -246,15 +251,15 @@
                                         AutomationProperties.Name="{x:Bind}"
                                         CornerRadius="12">
                                         <Grid
-                                            Padding="8,3,8,4"
+                                            Padding="8,0"
+                                            Height="24"
```
niels9001, line 255: We shouldn't hardcode the height here, because I'm not sure what's going to happen when text scaling is > 100%. Please revert

niels9001 (changes requested):
The tags are no longer showing up for me? Please revert those styling changes as they are not in scope for this issue and might introduce regressions with text scaling enabled.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -186,55 +187,59 @@
...
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="Tags" />
                         <ItemsView
                             x:Name="TagsItemsView"
                             Margin="0,8,0,4"
-                            x:Load="{x:Bind SelectedItem.Tags, Mode=OneWay, Converter={StaticResource emptyCollectionToFalseConverter}}"
+                            x:Load="{x:Bind SelectedItem.Tags, Converter={StaticResource emptyCollectionToFalseConverter}}"
                             IsItemInvokedEnabled="True"
                             ItemInvoked="TagsItemsView_ItemInvoked"
-                            ItemsSource="{x:Bind SelectedItem.Tags, Mode=OneWay}"
```
niels9001, line 237: Please revert, I think this is breaking the tags. These get updated everytime a user select a different icon, so we need `OneWay` for sure.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -186,55 +187,59 @@
...
                         <TextBlock
                             Margin="0,4,0,0"
-                            FontSize="12"
+                            Style="{StaticResource CaptionTextBlockStyle}"
                             Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                             Text="Tags" />
                         <ItemsView
                             x:Name="TagsItemsView"
                             Margin="0,8,0,4"
-                            x:Load="{x:Bind SelectedItem.Tags, Mode=OneWay, Converter={StaticResource emptyCollectionToFalseConverter}}"
```
niels9001, line 234: Please revert, I think this is breaking the tags. These get updated everytime a user select a different icon, so we need `OneWay` for sure.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
Jay-o-Way: I don't think so. It'll cut off the corner of the icon.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
Jay-o-Way: > We shouldn't hardcode the height here

I tweaked things because the corner radius is hardcoded at 12. So I added height=24 to make sure it always matches the corner. The padding at the top/bottom can only cut into the text.

niels9001: I don't think WinUI have a hardcoded height for the reason that things need to scale (e.g. see button padding). So let's stick to that.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
niels9001: Do we have an example of that?

image

99% of Segoe Fluent Icons are designed with rounded corners, so we should be good?

Jay-o-Way: Yes, check the **GripperResize**, it's in the main comment ⬆️

niels9001: image

Even here it renders fine? And it's the only one I could find 🤔

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
Jay-o-Way: If both height _and_ vertical padding are removed, the corners get messed up. What I think is that the **padding** was used to fix this. I think the best compromise is no vertical padding and `MinHeight="24"`.

Jay-o-Way: Off-topic: try setting text size to something extremely large, like 175% or 200%. Note that the content of the SidePanel can get clipped at the bottom. This is because the **SampleThemeListener** has `MaxHeight="600"`.

niels9001: Alright, let's stick to with the MinHeight 👍.

```<ScrollViewer Padding="16,16,8,16">```

We should make this scrollable - we can wrap the with a `ScrollViewer` and that should fix it :)

File: WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml (hunk 1 above)
Comment (Jay-o-Way) [reply WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml:200]: Looks like the cornerradius doesn't actually cut into the icon, but goes behind it. Lucky. Maybe image might be a bit strange. But I agree it's not a huge deal - it's only a few pixels.

File: WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml (hunk 2 above)
Comment (Jay-o-Way) [reply WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml]: > We should make this scrollable

Let's create a new issue or PR for this, because we already have a scroller on the page _and_ a scroller in the **IconsItemsView** and this is already a bit "finniky".

File: WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml [hunk 5]
```diff
@@ -246,15 +251,15 @@
                                         AutomationProperties.Name="{x:Bind}"
                                         CornerRadius="12">
                                         <Grid
-                                            Padding="8,3,8,4"
+                                            Padding="8,2"
```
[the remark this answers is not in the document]
Jay-o-Way: Yeah I believe that's just how Text(block) work...

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
Zakariathr22: What about adding 2px padding to the border?

Jay-o-Way: > What about adding 2px padding to the border?

Understandable idea, but then it looks like the padding is actually part of the icon and that's exactly what I don't want :)

Zakariathr22: I had an idea for the design (just a suggestion)
imageimage
imageimage

It follows Fluent Design principles by **applying radius**, **adds spacing** to prevent the corners from being cut off, and ensures that **the spacing doesn't appear to be part of the icon**.

Jay-o-Way: @Zakariathr22 That looks extremely complicated and crowded. The most simple thing is a single border without any radius. It's not a control or an overlay. I think we can get away with that. 🤷‍♂️

Zakariathr22:
**Out of topic:**
It's funny how simply wrapping a `FontIcon` inside a `Border` led us to all of this 😂

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
niels9001: Let's stick to border only.

@jay-o-way Brought this up during our design sync, the team suggested to add rounded corners as this now stands out. They referred to our internal icon tool, and mentioned that an additional padding would still have the same visual effect:

image

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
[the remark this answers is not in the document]
niels9001: I want to get this PR in, we can fix this later.

niels9001 (changes requested):
@jay-o-way Please add the rounded corners (and optional padding to overcome to corner issue) as proposed by the design team, then we can get this in.

Thanks!

Jay-o-Way:
@niels9001 Why don't you just finish it for me. Since we can't seem to agree on anything. Just do it your way.

niels9001 (approved):

Jay-o-Way:
In general; setting a **Padding** on a **Scrollviewer** may produce "ugly" results at the borders. Setting the margin on the first content will solve that.
image

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -184,92 +184,99 @@
...
-                                    <ItemContainer
-                                        HorizontalAlignment="Left"
-                                        AutomationProperties.Name="{x:Bind}"
-                                        CornerRadius="12">
-                                        <Grid
-                                            Padding="8,3,8,4"
-                                            Background="{ThemeResource CardBackgroundFillColorDefaultBrush}"
-                                            BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
-                                            BorderThickness="1"
+                    <ScrollViewer Padding="16,16,8,16">
```
Jay-o-Way, line 192: In general; setting a **Padding** on a **Scrollviewer** may produce "ugly" results at the borders. Setting the margin on the first content will solve that.
image

Jay-o-Way (commented):
nit

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
niels9001: Great catch! Resolved!

PR #1870, opened 2025-05-06
Remove unused constructor from `ControlInfoDocLink` class

Zakariathr22:
## Motivation and Context
- Clean code.
- Deprecate the old versioning approach.

## How Has This Been Tested?
Manually tested

## Types of changes

### DIFF
--- a/WinUIGallery/Models/ControlInfoData.cs
+++ b/WinUIGallery/Models/ControlInfoData.cs
@@ -3,7 +3,7 @@
 
 using System.Collections.ObjectModel;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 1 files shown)


Zakariathr22:
> Thank you for making this change, however I dont think we still need this code at all. Searching for X.Y did not yield any results beside that replacement 😄

Good point, should we remove that logic?

marcelwgn:
Yes I think we can remove it now given we should point to winui main anyway.

Zakariathr22:
> Yes I think we can remove it now given we should point to winui main anyway.

Done!

marcelwgn (approved):

PR #1875, opened 2025-05-09
Add personalization features: Favorites and Recently Viewed Samples

### LINKED ISSUE #1862: Proposal: Adding Favorites and Recently Viewed Samples
### Issue type

New feature

### Which version of the app?

WinUI 3 Gallery

### Description

This focuses on implementing personalization features like `favorites` and `recently viewed samples`.
These features aim to improve user navigation and engagement by providing quick access to preferred and recently interacted content.

Zakariathr22:
## Description
This PR introduces **Favorites** and **Recently Viewed Samples** to enhance personalization. These features aim to improve navigation efficiency and user engagement.
**Main changes are:**
- Added StringListSettingsHelper class to manage storing/retrieving recently viewed and favorite samples.
- Integrated favorite button in page headers with full persistence logic.
- Introduced HorizontalScrollContainer control (adapted from AI Dev Gallery's TileGallery).
- Refactored HomePage layout to include recent/favorite tabs using the new components.
- Created and applied new styles: SingleRowGridViewItemStyle, TokenViewSelectorBarStyle, and TokenViewSelectorBarItemStyle. (SelectorBar styles was adapted from AI Dev Gallery)
- Moved sample click logic into the ItemTemplate for better reuse and separation of concerns.
- Removed obsolete click-handling logic from all pages now using the updated template.
- Redesigned ItemTemplate layout to align visually with the HorizontalScrollContainer.
- Added new settings card to SettingsPage with options to clear recently visited samples and unfavorite all items.
- Design tweaks.

## Motivation and Context
- This update brings user-centric personalization, allowing quick access to preferred and recent samples.
- Closes #1862

## How Has This Been Tested?
Manually tested.

## Screenshots (if appropriate)
image
image
image
image
image

## Types of changes

## References
- [ApplicationData - Windows App SDK](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.windows.storage.applicationdata?view=windows-app-sdk-1.7)
- https://datatracker.ietf.org/doc/html/rfc8259#section-7
- [Unicode U+001F](https://www.compart.com/en/unicode/U+001F)


Zakariathr22:
@niels9001 @marcelwgn
This PR introduces significant logic and design updates and will require some time to complete.
Before proceeding, could you please confirm whether the proposed plan is approved or not?
I look forward to your decision.

Zakariathr22:
> @niels9001 @marcelwgn This PR introduces significant logic and design updates and will require some time to complete. Before proceeding, could you please confirm whether the proposed plan is approved or not? I look forward to your decision.

Couldn't stop overthinking, so I pushed a PR with breaking changes just to shut my brain up.
Now I’ve replaced overthinking with regret 😂, I wish it will be merged.
gif

niels9001:
Hey @Zakariathr22, thanks for opening this PR! 

We discussed this PR internally and we believe the recent/favorite tabs brings additional value (can't wait to use it myself😁!). We just need to ensure that we are not overcomplicating things, therefore we are suggesting a few changes:

### **Tabs**
Would it make sense to just have 2 tabs? 

**Tab 1: Recent**
- horizontal scroller with the last (5? 6? 7?) recent samples you've visited. (See AI Dev Gallery for an example)
- gridview with 'recently and new samples'

**Tab 2: favorites**
- gridview of favorite items

This keeps things simple, without providing too many options.

### **Tab controller**
Creating custom controls can always introduce potential a11y issues - so I was wondering if we could use a simple `SelectorBar`, with a custom style? We've used this custom style in AI Dev Gallery that we could consider to bring over? https://github.com/microsoft/ai-dev-gallery/blob/main/AIDevGallery/Styles/SelectorBar.xaml

### **Other changes**
- We were wondering if there's a lot of value in having the "send feedback" button in the PageHeader. Would it make sense to replace it with the favorite button? Thoughts?
- These changes on the homepage might trigger additional styling tweaks to the cards carousel and the items itself. I'll sync with the design team to figure out if any tweaks are needed, and we can help with those :)!

All and all, super appreciative of the work! It's awesome 🚀

Zakariathr22:
> Would it make sense to just have 2 tabs?

I think the data about the recently added and recently updated samples is also important.
Maybe we can change the order to: **recently visited, favorite samples, recently added, recently updated.**
Also, if we remove the recently added and recently updated samples, what should we do with their data in the JSON file?

---
> * last (5? 6? 7?) recent samples

It already retrieves the last 15 visited samples. We can simply edit it in `PageHeader.xaml.cs`.
```csharp
private const int _maxRecentlyVisitedPages = 15;
```

---
> Creating custom controls can always introduce potential a11y issues

But it does solve the overflow issue, and we’ll need it if we keep the list as I initially suggested.
Can we discuss these a11y issues first? Maybe I can fix it.

---
> We've used this custom style in AI Dev Gallery that we could consider to bring over?

Yes, it makes sense only if we're going to remove the recently added and recently updated data.
If we agree on that, I’ll be happy to handle it.

---
> * We were wondering if there's a lot of value in having the "send feedback" button in the PageHeader.

Personally, it was the first thing that guided me to this repo 😂.
Maybe we’re not using it as developers, but from a user’s perspective, I think it’s important.

---
> * I'll sync with the design team to figure out if any tweaks are needed, and we can help with those :)!

Could you also take a look at the `CriticalButtonStyle`? I used `SystemFillColorCriticalBrush` as the background and created the other theme resources, but I’m not sure if it follows the best design practices.

---
Generally, the implemented design was inspired by the Windows File Explorer, which adds familiarity and aligns with the overall Windows design language.
I was also thinking, can we move all the samples to the home page (as they’re already shown in this PR) and remove that extra page, so all navigation stays on the home page?

These my thoughts, so if what you're proposing is just a suggestion, I might disagree on a few things 😅.
But if it's a final decision, I’ll be happy to implement it.

niels9001:
The proposal was to have 2 tabs. The default tab (Recent - or if we have a better name for it) would combine these 3 into a single view:
- Recently visited (new)
- Recently or new samples

The reason for this is having 5 tabs with kind of similiar things make the UX more complex. We want the landingspage to be as calm as possible and reduce the number of ways to navigate to similar content.

Zakariathr22:
> The proposal was to have 2 tabs. The default tab (Recent - or if we have a better name for it) would combine these 3 into a single view:
> 
> * Recently visited (new)
> * Recently or new samples
> 
> The reason for this is having 5 tabs with kind of similiar things make the UX more complex. We want the landingspage to be as calm as possible and reduce the number of ways to navigate to similar content.

@niels9001, do you mean something like this?
Recent

---
Fav

niels9001:
Exactly!

Zakariathr22:
> Exactly!

🤝 Got it
It might just take a bit of time to get done.

niels9001 (approved):
Pushed a few design tweaks - amazing work @Zakariathr22!!

WinUIGallery/Styles/ItemTemplates.xaml
```diff
@@ -5,78 +5,76 @@
...
-                    RelativePanel.Below="titleText"
-                    Style="{StaticResource CaptionTextBlockStyle}"
-                    Text="{x:Bind Subtitle}"
-                    TextTrimming="CharacterEllipsis" />
-                <InfoBadge
-                    Grid.Column="2"
-                    VerticalAlignment="Top"
-                    Style="{StaticResource AttentionIconInfoBadgeStyle}"
-                    Visibility="{x:Bind BadgeString, Converter={StaticResource nullToVisibilityConverter}}" />
+        <Button
```
niels9001, line 8: Q: Why are we using a Button here instead of using the GridView click event?

Zakariathr22: Before, we used a `GridView`, but now we've switched to an `ItemsRepeater`. This change was necessary because we needed a `SelectorBar`, which required restructuring the page and wrapping everything in a `ScrollViewer`. If we kept using `GridView`, virtualization would be lost.

Additionally, in the previous `GridView` setup, there was a visual issue with sample items. The corner radius of the item container didn’t match the hover overlay border, creating visible gaps that affected the rounded design (see issue #1876).

That’s why switching to `ItemsRepeater` and moving the clicking logic directly to the Item was the best approach.

niels9001 (changes requested):
See comments

WinUIGallery/Styles/ItemTemplates.xaml
niels9001: Yeah, I like that! I think we should limit the amount of recent items anyways, as it's not really useful to have too many items there!

niels9001:
FYI, refactored the HomePage controls a bit so things are a bit easier to maintain :)

Zakariathr22:
> FYI, refactored the HomePage controls a bit so things are a bit easier to maintain :)

thanks for this! btw, I really liked the idea of using VisualStates.

WinUIGallery/App.xaml.cs
```diff
@@ -65,7 +65,7 @@ protected override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventArgs ar
         startupWindow.ExtendsContentIntoTitleBar = true;
 
         win32WindowHelper = new Win32WindowHelper(startupWindow);
-        win32WindowHelper.SetWindowMinMaxSize(new Win32WindowHelper.POINT() { x = 500, y = 500 });
+        win32WindowHelper.SetWindowMinMaxSize(new Win32WindowHelper.POINT() { x = 532, y = 500 });
```
marcelwgn, line 68: Where does that number come from?

WinUIGallery/Controls/PageHeader.xaml.cs
```diff
@@ -98,5 +101,50 @@ private void UserControl_Loaded(object sender, RoutedEventArgs e)
...
+        return (bool)isFavorite ? "Remove from favorites" : "Add to favorites";
+    }
+
+    private async void FavoriteButton_Click(object sender, RoutedEventArgs e)
+    {
+        if (sender is ToggleButton toggleButton && Item != null)
+        {
+            if (toggleButton.IsChecked == true)
+            {
+                if (!StringListSettingsHelper.TryAddItem(SettingsKeys.Favorites, Item.UniqueId, InsertPosition.Last, maxSize:12, trimEnabled:false))
```
marcelwgn, line 127: Why only 12? I do not think that we should limit this. On the topic of limiting, even if you add every page as favorite, the resulting string is less than 3KB so we do not need this check if we do not want to limit favorites.

WinUIGallery/Helpers/StringListSettingsHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+    /// Insert at the end of the list.
+    /// </summary>
+    Last
+}
+
+/// <summary>
+/// Helper for storing and managing string-based lists (e.g., favorites, history)
+/// in LocalSettings of ApplicationData using a compact, delimiter-based format.
+/// </summary>
+public static class StringListSettingsHelper
```
marcelwgn, line 29: I think this should be a settings helper instead of being specific to persisting StringList.

WinUIGallery/Helpers/StringListSettingsHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// <summary>
+    /// Checks whether the specified item exists in the list stored under the key.
+    /// </summary>
+    /// <param name="key">Settings key</param>
+    /// <param name="item">Item to check for</param>
+    /// <returns>True if item exists, false otherwise</returns>
+    public static bool Contains(string key, string item)
+    {
+        return GetList(key).Contains(item);
```
marcelwgn, line 118: Should use HashSet instead of a list for this.

WinUIGallery/Pages/NavigationRootPage.xaml
```diff
@@ -270,7 +270,7 @@
                             Content="Spacing"
                             Tag="Spacing">
                             <NavigationViewItem.Icon>
-                                <FontIcon Glyph="&#xE8A9;" />
+                                <FontIcon Glyph="&#xE8B3;" />
```
marcelwgn, line 273: Related?

WinUIGallery/Pages/SettingsPage.xaml.cs
```diff
@@ -39,6 +39,9 @@ public SettingsPage()
     {
         this.InitializeComponent();
         Loaded += OnSettingsPageLoaded;
+        ClearVisitedSamplesCard.IsEnabled = StringListSettingsHelper.GetList(SettingsKeys.RecentlyVisited).Count > 0;
+        UnfavoriteSamplesCard.IsEnabled = StringListSettingsHelper.GetList(SettingsKeys.Favorites).Count > 0;
+        SamplesSettingsExpander.IsExpanded = StringListSettingsHelper.GetList(SettingsKeys.RecentlyVisited).Count > 0 || StringListSettingsHelper.GetList(SettingsKeys.Favorites).Count > 0;
```
marcelwgn, line 44: Avoid loading data again, use ClearVisitedSamples.IsEnabled and UnfavoriteSamplesCard.IsEnabled instead. Ideally make it a computed property and reuse in the two click handlers below.

WinUIGallery/Styles/CriticalButtonStyle.xaml
```diff
@@ -0,0 +1,191 @@
...
+                <SolidColorBrush x:Key="CriticalButtonForegroundPointerOver"
+                                 Color="#101010" />
+                <SolidColorBrush x:Key="CriticalButtonForegroundPressed"
+                                 Color="#202020" />
+                <SolidColorBrush x:Key="CriticalButtonForegroundDisabled"
+                                 Color="#8A8A8A" />
+            </ResourceDictionary>
+
+            <!-- High Contrast -->
+            <ResourceDictionary x:Key="HighContrast">
```
marcelwgn, line 63: Do not hardcode colors like this, please use the existing high contrast as outlined here: https://learn.microsoft.com/en-us/windows/apps/design/accessibility/high-contrast-themes
Reason is that you have no way of knowing if they align with the users high contrast theme or if the chosen colors actually are not accessible for them.

WinUIGallery/Styles/CriticalButtonStyle.xaml
```diff
@@ -0,0 +1,191 @@
...
+            </ResourceDictionary>
+
+        </ResourceDictionary.ThemeDictionaries>
+    
+    <Style x:Key="CriticalButtonStyle"
+           TargetType="Button">
+        <Setter Property="Foreground"
+                Value="{ThemeResource CriticalButtonForeground}" />
+        <Setter Property="Background"
+                Value="{ThemeResource SystemFillColorCriticalBrush}" />
```
marcelwgn, line 97: I think we should define a CriticalButtonBackground brush and have it point it to this for completeness sake.

WinUIGallery/Styles/GridViewItem.xaml
```diff
@@ -7,6 +7,13 @@
         <Setter Property="Margin" Value="0,0,12,12" />
     </Style>
 
+    <Style
```
marcelwgn, line 10: Why do we need this?

WinUIGallery/Pages/HomePage.xaml.cs
```diff
@@ -7,85 +7,70 @@
...
-                case "New":
-                    item.Title = "Recently added samples";
-                    break;
-                case "Updated":
-                    item.Title = "Recently updated samples";
-                    break;
-                case "Preview":
-                    item.Title = "Preview samples";
-                    break;
+               StringListSettingsHelper.TryRemoveItem(settingsKey, id);
```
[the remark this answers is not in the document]
Zakariathr22: Is this really necessary?

WinUIGallery/Styles/SelectorBar.xaml
```diff
@@ -0,0 +1,339 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+    <ResourceDictionary.MergedDictionaries>
+        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
```
marcelwgn, line 5: Dont think thats needed.

WinUIGallery/Styles/SelectorBar.xaml
```diff
@@ -0,0 +1,339 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
marcelwgn, line 1: Please remove commented out XAML code from this file.

marcelwgn (changes requested):

WinUIGallery/Pages/HomePage.xaml.cs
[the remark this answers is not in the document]
marcelwgn: I dont think it is, if we want to do this we should while loading the initial data source.

WinUIGallery/Controls/OpacityMaskView.xaml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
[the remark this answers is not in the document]
niels9001: That's on me. I've made UX changes to accommodate the new features of this PR.

WinUIGallery/Helpers/VersionHelper.cs
```diff
@@ -4,7 +4,7 @@ namespace WinUIGallery.Helpers;
 internal static class VersionHelper
 {
     public static string WinAppSdkDetails => 
-        $"Windows App SDK {ReleaseInfo.Major}.{ReleaseInfo.Minor}";
+        $"Windows App SDK {ReleaseInfo.Major}.{ReleaseInfo.Minor}.{ReleaseInfo.Patch}";
```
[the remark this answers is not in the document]
niels9001: Again, this is on me. We might be upgrading more frequently going forward and so we can ensure who is using what version.

WinUIGallery/App.xaml.cs
Zakariathr22: To avoid overlapping buttons in the header
image

WinUIGallery/Pages/NavigationRootPage.xaml
Zakariathr22: I did this when I was going to add the `all samples` list to the home page, to avoid confusion between the icons.

WinUIGallery/Styles/SelectorBar.xaml
Zakariathr22: I think this is the best design because it creates familiarity with what is used in File Explorer and AI Dev Gallery. Additionally, the current SelectorBar has an issue that I couldn't overlook 😅

https://github.com/user-attachments/assets/1c9448f0-ef60-4935-80b4-0a7b06a5d8a9

Zakariathr22: > I think to Marcels point this is about adding that MergedDictionaries reference.
> 
> In the past this could cause crashes when not added, but I believe it was foxed in one of the recent WASDK updates. Not sure if we are already on that version though.

Yes, It was fixed in WASDK 1.7.0

WinUIGallery/Helpers/VersionHelper.cs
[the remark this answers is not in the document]
marcelwgn: But is the right way to check that to have people show their home page instead of going into settings and telling us the app version?

WinUIGallery/Styles/SelectorBar.xaml
Zakariathr22: > @Zakariathr22 Can you point me to the source for that? I would be very surprised that would crash given how WinUI resource lookup has worked (assuming you adhere to ordering conventions)

https://www.youtube.com/watch?v=1ROCCmAHdNs
minute 20:26

WinUIGallery/Styles/CriticalButtonStyle.xaml
Zakariathr22: should we simply remove it?

niels9001: Yeah.. maybe easier. I do think there's a use case, and maybe we could start thinking how a default CriticicalButtonStyle in WinUI could look like in the WinUI repo

WinUIGallery/Controls/PageHeader.xaml.cs
```diff
@@ -98,5 +99,50 @@ private void UserControl_Loaded(object sender, RoutedEventArgs e)
...
+        if (sender is ToggleButton toggleButton && Item != null)
+        {
+            if (toggleButton.IsChecked == true)
+            {
+                if (!SettingsHelper.TryAddItem(SettingsKeys.Favorites, Item.UniqueId, InsertPosition.Last))
+                {
+                    // Revert toggle state since add failed
+                    toggleButton.IsChecked = false;
+
+                    var contentDialog = new ContentDialog
```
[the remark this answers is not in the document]
Zakariathr22: Since IDs are stored as strings in JSON, I had to ensure the delimiter couldn’t possibly appear within any valid ID. According to the JSON spec (RFC 8259, section 7), all Unicode characters are allowed inside a string except for three that must be escaped: the quotation mark (`"`), the backslash (`\`), and control characters in the range `U+0000` to `U+001F`.

Given that, using a control character like the Unit Separator (`U+001F`) as a delimiter is safer than something common like a comma, it can't legally appear in a valid JSON string without being escaped, and no sane producer of JSON will include control characters like `U+001F` in a normal string unless explicitly intended.

Reference: [RFC 8259, Section 7](https://datatracker.ietf.org/doc/html/rfc8259#section-7)

WinUIGallery/Helpers/SettingsHelper.cs
```diff
@@ -0,0 +1,183 @@
...
+public static class SettingsHelper
+{
+    // The maximum number of items to keep in the Favorites list.
+    public const int MaxRecentlyVisitedSamples = 7;
+
+    // An instance of ApplicationData for the current user.
+    private static ApplicationData appData = ApplicationData.GetDefault();
+
+    // Unit Separator: invisible, JSON-safe delimiter for internal string serialization.
+    private const char Delimiter = '\u001f';
```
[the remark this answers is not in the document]
Zakariathr22: Since IDs are stored as strings in JSON, I had to ensure the delimiter couldn’t possibly appear within any valid ID. According to the JSON spec (RFC 8259, section 7), all Unicode characters are allowed inside a string except for three that must be escaped: the quotation mark (`"`), the backslash (`\`), and control characters in the range `U+0000` to `U+001F`.

Given that, using a control character like the Unit Separator (`U+001F`) as a delimiter is safer than something common like a comma, it can't legally appear in a valid JSON string without being escaped, and no sane producer of JSON will include control characters like `U+001F` in a normal string unless explicitly intended.

Reference: [RFC 8259, Section 7](https://datatracker.ietf.org/doc/html/rfc8259#section-7)

WinUIGallery/Helpers/SettingsHelper.cs
```diff
@@ -0,0 +1,183 @@
...
+            Debug.WriteLine($"Storage limit exceeded for key: {key}.");
+            return false;
+        }
+
+        appData.LocalSettings.Values[key] = joined;
+        return true;
+    }
+}
+
+public static class SettingsKeys
```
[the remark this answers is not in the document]
Zakariathr22: maybe we can also include other keys in a separate PR

WinUIGallery/Controls/PageHeader.xaml.cs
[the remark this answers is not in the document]
Zakariathr22: Since we have no control over the maximum number of favorites or the length of future string IDs, we can’t reliably predict how large the data might get. In most cases, the warning message won’t be needed, but for **robust, future-proof code**, it absolutely should be there.

WinUIGallery/Controls/PageHeader.xaml.cs
```diff
@@ -98,5 +99,50 @@ private void UserControl_Loaded(object sender, RoutedEventArgs e)
         {
             APIDetailsBtn.Visibility = Visibility.Collapsed;
         }
+        if (Item != null)
+        {
+            FavoriteButton.IsChecked = SettingsHelper.Contains(SettingsKeys.Favorites, Item.UniqueId);
+            SettingsHelper.TryAddItem(SettingsKeys.RecentlyVisited, Item.UniqueId, InsertPosition.First, SettingsHelper.MaxRecentlyVisitedSamples);
```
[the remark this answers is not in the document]
Zakariathr22: I'm trying to keep the helper class general-purpose. What do you think about adding a `MarkPageVisited` method to the `PageHeader` control instead? Or at the very least, adding a comment to explain what this line does.

WinUIGallery/Helpers/SettingsHelper.cs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /// <summary>
+    /// Checks whether a serialized list exists under the specified key in LocalSettings.
+    /// </summary>
+    /// <param name="key">The settings key to check.</param>
+    /// <returns>
+    /// True if the key exists and contains a non-empty serialized string; otherwise, false.
+    /// </returns>
+    public static bool HasList(string key)
```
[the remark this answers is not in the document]
Zakariathr22: > Technically both of those dont know about lists, but rather only delete or check existence for that key. I think its better to name them what they are.

good point

WinUIGallery/Controls/PageHeader.xaml.cs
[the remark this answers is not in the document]
marcelwgn: I dont think a PersonalisationHelper would actually do much more than calling SettingsHelper with specific parameters but we can do that in a different PR. What I think we should rather address is that it is a bit strange that PageHeader somehow is tracking which page is being navigated to instead of where the navigation is actually being done, i.e. Frame.Navigate is being called.

WinUIGallery/Controls/PageHeader.xaml.cs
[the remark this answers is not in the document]
marcelwgn: We sort of can predict that, we do have control over the IDs and adding recovery code that is literally untestable with the current setup is not ideal. FWIW, I think it might be better to have a unit test that ensures that we could favorite every element and it still working fine (might as well add some tests for settings helper in general)

WinUIGallery/Controls/PageHeader.xaml.cs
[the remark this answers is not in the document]
Zakariathr22: > I dont think a PersonalisationHelper would actually do much more than calling SettingsHelper with specific parameters but we can do that in a different PR. What I think we should rather address is that it is a bit strange that PageHeader somehow is tracking which page is being navigated to instead of where the navigation is actually being done, i.e. Frame.Navigate is being called.

Actually, that's a very good point, I hadn’t thought of it that way.
I've moved the logic for marking recently visited samples from PageHeaderControl to NavigationRootPage. Please review :-)

WinUIGallery/Pages/NavigationRootPage.xaml.cs
```diff
@@ -166,6 +166,14 @@ public void Navigate(Type pageType, object targetPageArguments = null, Navigatio
             Parameter = targetPageArguments
         };
         rootFrame.Navigate(pageType, args, navigationTransitionInfo);
+
+        // Ensure the NavigationView selection is set to the correct item to mark the sample's page as visited
+        if (!pageType.Equals(typeof(ItemPage)) || targetPageArguments == null)
```
marcelwgn: nit: I think the inverted case makes more sense instead of an early return right before the last statement of the file

marcelwgn (approved):

niels9001 (approved):
@zakariathr22 LGTM! Again.. AMAZING work!!!!

PR #1877, opened 2025-05-15
Remove ToolTip from each icon on the Iconography page

AndrewKeepCoding:
## Description

This PR removes the `ToolTip` from each icon on the Iconography page.

## Motivation and Context

The `ToolTip` just shows the icon name, which you can see it on the icon itself or on the left pane.

## How Has This Been Tested?

I ran the app and verified the change.

## Types of changes

### DIFF
--- a/WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
@@ -43,8 +43,7 @@
                 Width="96"
                 Height="96"
                 AutomationProperties.Name="{Binding Name}"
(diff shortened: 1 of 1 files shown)


niels9001:
@AndrewKeepCoding We actually need this for a11y, as some longer titles are cut off otherwise: 
image

Jay-o-Way:
I was just thinking (and about to comment) the exact same thing! 😆

AndrewKeepCoding:
Yeah, I was aware of that, but I just preferred the use experience without seeing the duplicated text. Well, it was worth a shot. 😉

PR #1881, opened 2025-05-21
Include Symbol enum indication in Iconography page

### LINKED ISSUE #1880: Include Symbol enum value in the Iconography page
### Issue type

enhancement

### Which version of the app?

WinUI 3 Gallery

### Description

In the Iconography page, please include the `Microsoft.UI.Xaml.Controls.Symbol` value when the current icon is part of the enum.

This is very helpful so that it enables us to use `SymbolIcon` when possible. `Symbol` is preferred as it's much more self-explanatory and readable over a gibberish font icon unicode value.

Below are a couple of examples, one of the `Delete` symbol which *is* part of the enum, vs. `EraseTool` which is not.

I'm willing to implement that if given green light.

### Screenshots

Image

Image

Image

### Windows version

_No response_

### Additional context

_No response_

weitzhandler:
## Description
- Fixes #1880
- Fixes issues filling up the entire screen with the icon list by removing the max-heigh limit

## Types of changes


WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -72,6 +72,16 @@
                         Text="{Binding Name}"
                         TextTrimming="CharacterEllipsis"
                         TextWrapping="NoWrap" />
+
+                    <InfoBadge
```
niels9001, line 76: Do we actually need this? Just thinking out loud: one would search for an icon, and could then use the `Symbol` if it would be available. What would a visual indicator add?

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml.cs
```diff
@@ -160,4 +176,20 @@ private void TagsItemsView_ItemInvoked(ItemsView sender, ItemsViewItemInvokedEve
...
     }
+
+    ScrollBarVisibility oldOuterScrollingValue;
+    private void OverrideUpperScrollViewer(bool isEnabled)
+    {
+        var scrollViewer = this.FindAscendant<ScrollViewer>();
+
+        if (isEnabled)
+        {
+            scrollViewer.VerticalScrollBarVisibility = oldOuterScrollingValue;
```
niels9001, line 187: image

This crashes when navigating away from the page

niels9001: Also, would it make sense to turn this in a helperclass? That way we would be able to re-use this code to block page scrolling - that'll be helpful for the colors page where we'd like to have the selectorbar to stick and the content below to scroll.

niels9001 (changes requested):

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
AndrewKeepCoding: Also thinking out loud: Adding `SymbolIcon` as tag should help filtering. 🧐

weitzhandler: > Do we actually need this?

I think yes. So that you don't have to open the individual icons to know if they allow you to use a `SymbolIcon`.



> Adding `SymbolIcon` as tag to the corresponding icons.

Could do in a separate PR

Zakariathr22 (commented):
Why don't we just add the SymbolEnum property to `IconData` and JSON data?

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
niels9001: But I think that's my point; why would someone be searching for icons that are part of symbolicons?

Wouldn't they be searching for the right icon, and if SymbolIcon is available, they could use that, and if not, use FontIcon?

weitzhandler:
> Why don't we just add the SymbolEnum property to `IconData` and JSON data?

We anyway bind to its string value so it's pointless.

weitzhandler:
> Why don't we just add the SymbolEnum property to `IconData` and JSON data?

Not a bad idea, but prefer doing this in a separate PR.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
weitzhandler: I've decreased the size of that `InfoBadge` from 8 to 6 and changed the opacity to 0.4 to minimize its footprint.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml.cs
```diff
@@ -49,6 +49,7 @@ public IconographyPage()
     {
         // Fill filtered items
         this.InitializeComponent();
+        this.SuspendOuterScrolling();
```
marcelwgn, line 52: I would really prefer this to be a separate PR, especially since I would like to discuss if we can do this differently.

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
marcelwgn: I agree with @niels9001, I see little value in being able to see this quickly. If this was a criteria, we would also need to allow to hide all non symbol icons via a filter. Since we do not do this, I think its better to not highlight this.

weitzhandler: Removed.

niels9001 (approved):

marcelwgn (approved):

Issue #1858, opened 2025-04-27
Textscaling issues


niels9001:
### Which version of the app?

WinUI 3 Gallery

### Screenshots

When textscaling is set to 200%, a few issues arise:

Titles are cut off
![Image](https://github.com/user-attachments/assets/fda56de2-99b8-4ecd-955c-91be72340d0c)

Icon properties panel is not scrollable

![Image](https://github.com/user-attachments/assets/9b2087b5-dbb9-477d-8851-324cce0b7d11)

Jay-o-Way:
Also check the table in Geometry page

Jay-o-Way:
> [!IMPORTANT]
> Note  that there are already two vertical scrollers: one on the **page** and one on the **IconsItemsView**, which are already tricky together.

niels9001:
> Important
> 
> Note that there are already two vertical scrollers: one on the **page** and one on the **IconsItemsView**, which are already tricky together.

I wonder if we should disable page scrolling for certain pages (colors overview and iconography) so we can scroll parts of those pages?

E.g. for thr colors page we'd want the tab to stick, and only scroll the color tiles. For iconography we'd want the searchbox to stick but the icon grid to stretch and scroll?

Jay-o-Way:
For those pages, it's the text at the top that wants to take up space. If the text had a different place, the content below it would have more space and maybe we could use one scroller for the page. But then, where do we leave the text?
Having the Input/Selector stick to the top is also a cool idea!

niels9001:
> For those pages, it's the text at the top that wants to take up space. If the text had a different place, the content below it would have more space and maybe we could use one scroller for the page. But then, where do we leave the text? Having the Input/Selector stick to the top is also a cool idea!

Right.. Here was my thinking:

<img width="687" alt="Image" src="https://github.com/user-attachments/assets/d461d96c-fe75-4c25-aa2c-e5a8316837dc" />

The red area sticks, so you navigate it at all times. The blue area takes up all available space and is scrollable.

For the icon page it'd be something similiar:

<img width="994" alt="Image" src="https://github.com/user-attachments/assets/847e474a-2a82-4d84-a6b5-318f82bfe54e" />

The searchbox and text sticks, the sidepanel too and takes the available height of the screen. The grid is scrollable and takes all available space vertically.

Zakariathr22:
I think there is no need for the text on the Colors page anymore.
[...]

PR #1884, opened 2025-05-26
Add internal page scrolling functionality

niels9001:
In certain edge cases, we don't want the entire page to scroll but only a specific part of the page (e.g. the Colors and Icons pages).

This PR introduces an attached DP that can be set on a page to enable this.

Closes: https://github.com/microsoft/WinUI-Gallery/issues/1858#issuecomment-2834624805


WinUIGallery/Helpers/ScrollBehaviorHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.UI.Xaml;
+
+namespace WinUIGallery.Helpers;
+public static class PageScrollBehaviorHelper
+{
+    public static readonly DependencyProperty SuppressHostScrollingProperty =
+        DependencyProperty.RegisterAttached(
```
snigdha011997, line 10: Nice use of an attached property to control scroll behavior!. Just a suggestion a brief XML doc comment on the property and methods would help clarify its intended usage for other developers.

Issue #1736, opened 2025-01-29
Create a MainWindow + new titlebar


niels9001:
We currently create a new `Window` upon launch, and settings the `NavigationRootPage` as its content.

Creating a `MainWindow` and moving that code there makes things easier when we want to make use of certain Windowing APIs and the new TitleBar.

Once the new `TitleBar` control is in stable, we can move the NavigationView-button and searchbox to the titlebar area per latest design guidance:

![Image](https://github.com/user-attachments/assets/adc82d36-c85f-4447-97d7-0ff9a62b6c39)

Jay-o-Way:
Must solve #1605

Jay-o-Way:
Also note (for example) #145 which seems to be a work-around that probably should be reverted. If you create the MainWindow with the Navigation from scratch, you should find this. FYI @marcelwgn right?

Jay-o-Way:
Has anybody tested this on Windows 10? Thinking of possible issues like https://github.com/microsoft/microsoft-ui-xaml/issues/10489

PR #1797, opened 2025-03-08
[DRAFT] MainWindow.xaml with new TitleBar + TitleBar samples

niels9001:
DO NOT MERGE UNTIL WASDK 1.7 STABLE

Closes: #1736 #1798

Changes:
- Introducing a new `MainWindow.xaml` in favor of creating a window in code-behind + `NavigationRootPage`.
- Cleanup of unused code
- Using the new `TitleBar` control: moving the pane toggle button and searchbox to the `TitleBar` following design best practices for Windows 11
- Made the Copy link logic dynamic: in debug mode, the link will open the actual dev version of the Gallery for easier debugging.
- Added samples for TitleBar

image

PR #1908, opened 2025-06-30
Replace NavigationRootPage with MainWindow

niels9001:
## Description
This PR creates a XAML-defined `MainWindow` for improved maintaining of the shell UX going forward. It also uses the new `TitleBar` control and moving the searchbox to the top of the window, following UX-guidance and saving some vertical space in the NavView.

Closes: #1736

~~[DO NOT MERGE]: Blocked on `TitleBar` bugs~~

For TitleBar samples, see: #1797


WinUIGallery/MainWindow.xaml.cs
```diff
@@ -8,158 +8,115 @@
...
+        titleBar.Subtitle = "Dev";
+#else
+        this.Title = "WinUI 3 Gallery";
+#endif
+        this.ExtendsContentIntoTitleBar = true;
+        this.SetTitleBar(titleBar);
+        this.AppWindow.SetIcon("Assets/Tiles/GalleryIcon.ico");
+        this.AppWindow.TitleBar.PreferredHeightOption = TitleBarHeightOption.Tall;
+
+        // TO DO: replace this with the new MinWidth/MinHeight AppWindow APIs in 1.7 Stable
```
marcelwgn: Arent we on 1.7?

WinUIGallery/Pages/AllControlsPage.xaml.cs
```diff
@@ -20,10 +20,10 @@ public AllControlsPage()
 
     protected override void OnNavigatedTo(NavigationEventArgs e)
     {
-        if (e.Parameter is NavigationRootPageArgs args && args.NavigationRootPage.NavigationView.MenuItems.OfType<NavigationViewItem>().FirstOrDefault(item => item.Name == "AllControlsItem") is NavigationViewItem item)
-        {
-            item.IsSelected = true;
-        }
+        //if (e.Parameter is NavigationRootPageArgs args && args.NavigationRootPage.NavigationView.MenuItems.OfType<NavigationViewItem>().FirstOrDefault(item => item.Name == "AllControlsItem") is NavigationViewItem item)
```
marcelwgn, line 23: Why is this commented out?

WinUIGallery/Pages/ItemPage.xaml.cs
```diff
@@ -94,16 +90,12 @@ protected override void OnNavigatingFrom(NavigatingCancelEventArgs e)
...
-            pageHeader.ToggleThemeAction = null;
-            pageHeader.CopyLinkAction = null;
-        }
+        App.MainWindow.NavigationViewLoaded = null;
+        pageHeader.ToggleThemeAction = null;
+        pageHeader.CopyLinkAction = null;
 
-        // We use reflection to call the OnNavigatedFrom function the user leaves this page
-        // See this PR for more information: https://github.com/microsoft/WinUI-Gallery/pull/145
+        //// We use reflection to call the OnNavigatedFrom function the user leaves this page
```
marcelwgn: nit: Double comment

niels9001:
I'll update this PR with a custom TitleBar so we are not blocked by the known issues. (Or if someone else can do it if they want to jump in :). The titlebar from Windows Community Toolkit Labs would be a great start!

Jay-o-Way:
What exactly was the problem up til now? And cant we choose to work around it, or just use it with a relatively small imperfection?
Using something else now, and changing it again when the WASDK/WinUi control is perfect, is more work.

marcelwgn:
I think for now it is fine to keep the titlebar implementation we currently have and focus on decluttering the files and their purpose. WinUI 3 titlebar has some bugs including making resizing the window sometimes hide window content so we should not introduce known regressions like that.

Zakariathr22:
As a workaround, could we avoid using the IconSource property of the TitleBar control and instead place an icon in the left header of the TitleBar?

niels9001:
> As a workaround, could we avoid using the IconSource property of the TitleBar control and instead place an icon in the left header of the TitleBar?

Yeah that works. Would you mind create a Pr against this branch so we can merge that in and get this PR ready for review too?

Zakariathr22:
> > As a workaround, could we avoid using the IconSource property of the TitleBar control and instead place an icon in the left header of the TitleBar?
> 
> Yeah that works. Would you mind create a Pr against this branch so we can merge that in and get this PR ready for review too?

- #1915

niels9001:
Thanks @Zakariathr22 ! This PR is now ready for review (cc @marcelwgn).

niels9001:
Thanks @Zakariathr22 ! This PR is now ready for review (cc @marcelwgn).

WinUIGallery/Pages/ItemPage.xaml.cs
niels9001: Fixed!

WinUIGallery/Pages/AllControlsPage.xaml.cs
niels9001: Oversight, I guess! Would you mind taking a look at this to get this PR in?

WinUIGallery/MainWindow.xaml.cs
niels9001: Yeah.. those APIs do not take DPI differences across screens into account, so they are inferior to the implementation we currently have, and that already works. Removed the comment.

marcelwgn (approved):
Great work and cool that we can also use the new TitleBar control (though I certainly need some time to get used to that centered search bar)

niels9001:
Hmm, looks like the UI tests are failing because of some of these changes.

ghost1372:
> Hmm, looks like the UI tests are failing because of some of these changes.

Probably because of the TitlebarPage that I fixed here. 
https://github.com/microsoft/WinUI-Gallery/commit/2ede36e9a283ce562c1e9b3d12005415a4369f35

marcelwgn:
I will look at the test failure, the search box got renamed and the new UIA name of the search field seems to be "TextBox"...

niels9001:
> I will look at the test failure, the search box got renamed and the new UIA name of the search field seems to be "TextBox"...

That seemed to work 🚀🚀

Issue #1627, opened 2024-09-26
AOT Support


ghost1372:
### Which version of the app?

WinUI 3 Gallery

### Description

I am looking to see how AOT is implemented in the gallery. I didn't see anything related to AOT, has it not started yet?
I'm more interested to see how [this line](https://github.com/microsoft/WinUI-Gallery/blob/b8bbce3c9073b9ada0302a88640a5c04e361eb47/WinUIGallery/DataModel/ControlInfoDataSource.cs#L202) will be modified.

https://github.com/microsoft/WinUI-Gallery/blob/b8bbce3c9073b9ada0302a88640a5c04e361eb47/WinUIGallery/DataModel/ControlInfoDataSource.cs#L202

karkarl:
We are working on the AOT samples for WinUI-Gallery. Unfortunately, there are unforeseen issues that is causing the PR validation pipeline to be blocked. Hopefully this can be resolved soon and we can have the samples in!

marcelwgn:
Thank you for your elaborate comment and sorry for the delayed response on my part. I think your approach is good, especially since it makes it easy to keep maintaining that list. If you would like to create a PR for this, I am sure we would accept such a contribution!

ghost1372:
> Thank you for your elaborate comment and sorry for the delayed response on my part. I think your approach is good, especially since it makes it easy to keep maintaining that list. If you would like to create a PR for this, I am sure we would accept such a contribution!

Thank you! @marcelwgn  
I'm planning to start working on a PR.
Should I wait until this one is merged https://github.com/microsoft/WinUI-Gallery/pull/1908 or can I go ahead and begin?

niels9001:
> > Thank you for your elaborate comment and sorry for the delayed response on my part. I think your approach is good, especially since it makes it easy to keep maintaining that list. If you would like to create a PR for this, I am sure we would accept such a contribution!
> 
> Thank you! @marcelwgn  
> I'm planning to start working on a PR.
> Should I wait until this one is merged https://github.com/microsoft/WinUI-Gallery/pull/1908 or can I go ahead and begin?
> 

Maybe branch of #1908 and I'll update that branch with removing the WinUI TitleBar due to blocking bugs, and add a custom TitleBar for now.

marcelwgn:
[...]

PR #1916, opened 2025-07-12
Add NativeAOT Support

ghost1372:
**This PR adds native AOT support #1627 .**

- [x] Using x:Bind instead of Binding
- [x] Using CsWin32 with Disabled Marshaling (NativeMethods)
- [x] Mark Classes as partial
- [x] Remove Reflections
- [x] BindingPage (NonExistentProperty Removed because xBind does not support it)
- [ ] Enforce AOT warnings as errors (Seperate PR)

> [!Note]
> - The app is using reflection and DllImport somewhere, but I didn’t see any trimming or AOT warnings, and I tested it locally without any issues—so I didn’t modify that code.
> - There’s a warning related to `FreeLibrarySafeHandle`, which comes from CsWin32. However, according to this [answer](https://github.com/microsoft/CsWin32/issues/1438#issuecomment-3075989430), it's safe to ignore for now as long as you're not passing these handles across the WinRT ABI.

# Notes for Future Developers / Contributors 
### Please read carefully before modifying the codebase:

> [!Warning]
> - Avoid using reflection whenever possible. It can cause issues with trimming and Native AOT.
> - Enum bindings: Use the recommended pattern.
> Instead of:
> `private IReadOnlyList<TitleBarTheme> titleBarThemes = Enum.GetValues(typeof(TitleBarTheme)).Cast<TitleBarTheme>().ToList();`
> Use:
> `private IReadOnlyList<TitleBarTheme> titleBarThemes { get; set; } = new List<TitleBarTheme>(Enum.GetValues<TitleBarTheme>());`
> - Avoid using P/Invoke directly. Use [CsWin32](https://github.com/microsoft/cswin32) and define your APIs in the NativeMethods.txt file.
> - Mark classes as partial.
> - Always provide a DataTemplate with x:DataType and avoid using `SelectedValuePath`, `DisplayMemberPath`.
> - Avoid binding directly to IEnumerable. Always call `.ToList()` to avoid runtime exceptions.


ghost1372:
Hi @niels9001, @marcelwgn 
The PR is now ready for review. However, there are two options that need to be addressed. I’m waiting for your feedback before deciding how to proceed.

marcelwgn:
Thank you for creating this PR @ghost1372! I think there has been a slight misunderstanding in the scope, I thought you would start with introducing the T4 templates to help address the NAOT issues with our current navigation. 
This PR is very difficult to review since there are often a lot of overlapping changes in a file. Given that you already split some of the changes, it would be great if we can split some of the major things, namely adding partial and moving to CSWin32 into separate PRs. If that is to difficult to do though we can keep the PR as is.

ghost1372:
@marcelwgn Thank you for the feedback! You're right—initially I planned to introduce T4 templates to help address the NAOT issues with navigation. However, during testing I found that the current implementation no longer produces any warnings or issues with `IsAotCompatible/PublishAot=true`, so I decided to keep the existing approach for now and did not proceed with adding T4.
Screenshot 2025-07-14 102635

Previously, this had a warning. But I don't see any warnings anymore and I tried to fix all the relevant warnings in this PR.

So I don't think it's necessary to do it.
I think we can continue with this PR — while there are many commits, each one introduces only small, focused changes, so the overall review should not be too complex. Otherwise, I should close this one and create three separate PRs from scratch if you prefer that approach.

WinUIGallery/Samples/ControlPages/Fundamentals/BindingPage.xaml
```diff
@@ -183,7 +183,7 @@ public string GreetingMessage { get; set; } = "Hello, WinUI 3!";
             <controls:ControlExample.Example>
                 <StackPanel Spacing="8">
                     <TextBlock Text="{Binding NonExistentProperty, FallbackValue='Default Text'}" />
```
Zakariathr22: I suggest removing this and change the example title to **"Binding with TargetNullValue"**

ghost1372: @Zakariathr22 I agree, I’m just waiting for the approval.

niels9001: I agree, I'd remove this to unblock this PR but also because it's a bit of a niche use case

Zakariathr22: Actually, it's possible to use `x:Bind` with `FallbackValue` in some scenarios. Unfortunately, there's a critical bug (https://github.com/microsoft/microsoft-ui-xaml/issues/10429), but the issue was closed as "completed" without any actual fix.

niels9001: I reopened the issue, but looks like there's an ask for a repro

marcelwgn:
We merged the MainWindow PR in, can you update this PR with the latest changes from main and resolve the merge conflicts? @ghost1372? 🙂

ghost1372:
@marcelwgn  
I've resolved the merge conflicts and updated the branch with the latest changes from main.

What should we do with **BindingPage**?
x:Bind does not support Non Existent Property
`                    <TextBlock Text="{Binding NonExistentProperty, FallbackValue='Default Text'}" />
`

marcelwgn:
Thank you @ghost1372 ! I think we can remove the FallbackValue section for now until we figure out what to do with it to unblock this PR.

Also, I agree that we should enforce NAOT warnings in the future but that should be done in a separate PR.

WinUIGallery/Helpers/Language/Language.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace WinUIGallery.Helpers;
+public partial class Language
```
marcelwgn, line 5: Next time, please don't mix general refactoring and changes for the specific scope. It is fine to create refactoring PRs but mixing refactoring into other changes makes it hard to review.

WinUIGallery/Models/ControlInfoData.cs
```diff
@@ -62,7 +62,8 @@ public class ControlInfoDocLink
 /// <summary>
 /// Generic group data model.
 /// </summary>
-public class ControlInfoDataGroup
+[WinRT.GeneratedBindableCustomPropertyAttribute]
```
marcelwgn, line 65: Why do we need this?

WinUIGallery/Samples/ControlPages/BreadcrumbBarPage.xaml
```diff
@@ -21,7 +21,7 @@
     <StackPanel>
         <controls:ControlExample x:Name="Example1" HeaderText="A BreadcrumbBar control">
             <controls:ControlExample.Example>
-                <BreadcrumbBar x:Name="BreadcrumbBar1" />
+                <BreadcrumbBar x:Name="BreadcrumbBar1" ItemsSource="{x:Bind FolderString}"/>
```
marcelwgn: Why are we switching to x:Bind here now?

WinUIGallery/Samples/ControlPages/BreadcrumbBarPage.xaml.cs
```diff
@@ -10,20 +10,29 @@ namespace WinUIGallery.ControlPages;
...
+    {
+        new Folder { Name = "Home" },
+        new Folder { Name = "Folder1" },
+        new Folder { Name = "Folder2" },
+        new Folder { Name = "Folder3" },
     };
 
+    public ObservableCollection<Folder> Folders { get; } = new();
+
+    public readonly string[] FolderString = new string[] { "Home", "Documents", "Design", "Northwind", "Images", "Folder1", "Folder2", "Folder3" };
```
marcelwgn:
```suggestion
    public readonly string[] FoldersString = new string[] { "Home", "Documents", "Design", "Northwind", "Images", "Folder1", "Folder2", "Folder3" };
```

WinUIGallery/Samples/ControlPages/BreadcrumbBarPage.xaml
marcelwgn:
```suggestion
                <BreadcrumbBar x:Name="BreadcrumbBar1" ItemsSource="{x:Bind FoldersString}"/>
```

WinUIGallery/Samples/ControlPages/PersonPicturePage.xaml
```diff
@@ -11,21 +11,12 @@
...
 
             <controls:ControlExample.Example>
-                <PersonPicture
-                    x:Name="personPicture"
-                    Height="300"
-                    VerticalAlignment="Top"
-                    DisplayName="{x:Bind DisplayNameRadio.IsChecked, Mode=OneWay, Converter={StaticResource booleanToValueConverter}, ConverterParameter='Jane Doe'}"
-                    Initials="{x:Bind InitialsRadio.IsChecked, Mode=OneWay, Converter={StaticResource booleanToValueConverter}, ConverterParameter='SB'}"
-                    ProfilePicture="{Binding IsChecked, ElementName=ProfileImageRadio, Converter={StaticResource booleanToValueConverter}, ConverterParameter='https://learn.microsoft.com/windows/uwp/contacts-and-calendar/images/shoulder-tap-static-payload.png'}" />
+                <PersonPicture x:Name="personPicture" Height="300" VerticalAlignment="Top" />
```
marcelwgn, line 14: What happened here? Why not continue using x:Bind?

WinUIGallery/WinUIGallery.csproj
```diff
@@ -12,7 +12,8 @@
         <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
         <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
         <NoWarn>
-            0108;   <!-- 'x' hides inherited member 'y'. Use the new keyword if hiding was intended. -->
+            CA2000;
```
marcelwgn: Why are we suppressing this?

marcelwgn:
I left a few comments on your changes. Regarding this:
> CopyButton: We're using regular Binding here because x:Bind isn’t supported in resource dictionaries. I didn’t encounter any issues during testing, but Visual Studio shows AOT-related warnings. At this point, there doesn’t seem to be a workaround.

AFAIK, you can use x:Bind in a resource dictionary if you back it by a class (which is not difficult to do) so I think we should address this before merging.

WinUIGallery/Helpers/Language/Language.cs
ghost1372: Thanks for the feedback, and sorry about that. I’ll make sure to separate general refactoring from scoped changes in future PRs to keep things easier to review

WinUIGallery/Models/ControlInfoData.cs
ghost1372: I was experiencing some random crashes, and this change resolved those issues.

WinUIGallery/Samples/ControlPages/PersonPicturePage.xaml
ghost1372: Because **ProfilePicture** cannot use **x:Bind** to bind directly to an **object**—it expects an **ImageSource**, and x:Bind does not support automatic type conversion in this case. On the other hand, Binding can perform implicit casting, which makes it work correctly. I initially tried using a converter with x:Bind and return ImageSource, but it caused some random issues, so I moved the logic to code-behind for stability.
As you can see, DisplayName and Initials use x:Bind along with a converter and work correctly. However, ProfilePicture uses Binding instead of x:Bind because it requires an ImageSource and doesn't accept an object directly.

WinUIGallery/WinUIGallery.csproj
ghost1372: Oops, that was a mistake

WinUIGallery/Samples/ControlPages/BreadcrumbBarPage.xaml
ghost1372: I moved the ItemsSource assignment for BreadcrumbBar1 from code-behind to XAML using x:Bind for better readability and maintainability. This way, the data flow is more visible in the markup itself, making the sample easier to understand at a glance without digging into the code-behind.

ghost1372: I didn’t realize your suggestion could be committed directly to the codebase, so I went ahead and applied the change myself. 😅

ghost1372:
> I left a few comments on your changes. Regarding this:
> 
> > CopyButton: We're using regular Binding here because x:Bind isn’t supported in resource dictionaries. I didn’t encounter any issues during testing, but Visual Studio shows AOT-related warnings. At this point, there doesn’t seem to be a workaround.
> 
> AFAIK, you can use x:Bind in a resource dictionary if you back it by a class (which is not difficult to do) so I think we should address this before merging.

I know we can back a ResourceDictionary with a class to use x:Bind, but as you can see, the icon isn’t showing
Screenshot 2025-07-25 154006

ghost1372:
@marcelwgn how about removing Binding and just using a ThemeResource? WinUI-Gallery is not a Class Library project, and 

> > I left a few comments on your changes. Regarding this:
> > > CopyButton: We're using regular Binding here because x:Bind isn’t supported in resource dictionaries. I didn’t encounter any issues during testing, but Visual Studio shows AOT-related warnings. At this point, there doesn’t seem to be a workaround.
> > 
> > 
> > AFAIK, you can use x:Bind in a resource dictionary if you back it by a class (which is not difficult to do) so I think we should address this before merging.
> 
> I know we can back a ResourceDictionary with a class to use x:Bind, but as you can see, the icon isn’t showing 


@niels9001 Can we add 2 more properties for Foreground? (Foreground, HoverForeground, PressedForeground)

niels9001:
> @marcelwgn how about removing Binding and just using a ThemeResource? WinUI-Gallery is not a Class Library project, and
> 
> > > I left a few comments on your changes. Regarding this:
> > > > CopyButton: We're using regular Binding here because x:Bind isn’t supported in resource dictionaries. I didn’t encounter any issues during testing, but Visual Studio shows AOT-related warnings. At this point, there doesn’t seem to be a workaround.
> > > 
> > > 
> > > AFAIK, you can use x:Bind in a resource dictionary if you back it by a class (which is not difficult to do) so I think we should address this before merging.
> > 
> > 
> > I know we can back a ResourceDictionary with a class to use x:Bind, but as you can see, the icon isn’t showing
> 
> @niels9001 Can we add 2 more properties for Foreground? (Foreground, HoverForeground, PressedForeground)

Sure! I'm also fine 'hardcoding' the right themebrushes in the style itself?

marcelwgn:
Sure lets use themeresources for the icon and add the theme brushes in the style. Good idea @ghost1372!

niels9001:
> Sure lets use themeresources for the icon and add the theme brushes in the style. Good idea @ghost1372!

We are using default button brushes, so we could also reference those directly vs defining new ones. @ghost1372 whatever is the easiest for you :)

ghost1372:
Thank you @niels9001  @marcelwgn 
Using ThemeResources is the easiest approach, but it still has some issues 😅. On the BackgroundSection page, there are some ColorTile elements that set the Foreground for the CopyButton. It seems the only solution is to add two new properties: HoverForeground and PressedForeground.

Screenshot 2025-07-25 202845

ghost1372:
Thank you @niels9001 @marcelwgn 
I initially used hardcoded `ThemeResource` values like this:
`Foreground="{ThemeResource ButtonForeground}"
`
However, I later realized that we're using CopyButton inside **TileColor**, where we need to override the **Foreground**. But because it's hardcoded, customizing the Foreground wasn't possible.

Screenshot 2025-07-25 202845


To solve this, I added a new property like **PressedForeground**, but then I discovered that **TemplateBinding** doesn’t work inside **DiscreteObjectKeyFrame**. So I went back to the original approach, but this time I set the Foreground manually in **OnApplyTemplate**, like this:

```
if (Foreground != null && contentPresenter != null && copySuccessPresenter != null)
{
    contentPresenter.Foreground = Foreground;
    copySuccessPresenter.Foreground = Foreground;
}
```
This worked, but after a pointer over or press event, it reverts to using the hardcoded **ThemeResource**, which can conflict with the background color in **TileColor**, making the icon invisible.
aa

So, it seems the best solution is to remove the **PointerOver** and **Pressed** **VisualState** setters for **Foreground** entirely, and handle them in code-behind instead. If you're okay with this approach, I can go ahead and implement it.

niels9001:
Would lightweight styling work in that case? E.g. we define a new brush in the template and on those colortiles we'd override those with lightweight styling?

ghost1372:
Thank you @niels9001 
Excellent Suggestion.
Finally i fixed it.

WinUIGallery/WinUIGallery.csproj
```diff
@@ -32,7 +32,9 @@
         <SingleProject Condition="'$(SolutionName)' == 'WinUIGallery.DesktopWap'">false</SingleProject>
         <WindowsAppSdkIncludeVersionInfo>true</WindowsAppSdkIncludeVersionInfo>
         <PublishProfile>win-$(Platform).pubxml</PublishProfile>
-        <PublishAot>false</PublishAot>
+		<DisableRuntimeMarshalling>true</DisableRuntimeMarshalling>
```
marcelwgn, line 35: Why do we need this?

ghost1372: disables the built-in .NET runtime marshalling infrastructure, which is enabled by default to support [DllImport] and other interop scenarios at runtime.
I've seen this in various projects including the following:
https://github.com/microsoft/XamlBehaviors/blob/a98c24a6a7d9bdd57cf5d406455b6f623e321a72/src/BehaviorsSDKManaged/Microsoft.Xaml.Interactivity.WinUI/Microsoft.Xaml.Interactivity.WinUI.csproj#L12

marcelwgn (approved):

WinUIGallery/WinUIGallery.csproj
```diff
@@ -104,6 +106,10 @@
             just add versionless PackageReferences here.  They will be updated to their actual versions by either the Packages.props file
             in the WinUI repo, or the next ItemGroup below when standalone. -->
     <ItemGroup>
+        <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.183">
+          <PrivateAssets>all</PrivateAssets>
+          <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+        </PackageReference>
```
michael-hawker, line 112: FYI @AjitSurana, not sure why the WinUI submodule isn't being kept in sync often with the main repo here, but this PR broke building in the WinUI repo. `Microsoft.Windows.CsWin32` should have a version property like the other references below in section on L128. 

And then a `MicrosoftCsWin32PackageVersion` build property updated (in the `standalone.props` file in this repo, not sure why that's not included in the sln, and the `packages.props/versions.props` files in the WinUI repo); as specifying a Version here breaks when building in the WinUI repo.

Issue #1557, opened 2024-06-04
Icons don't warn about Win10 not supported


dotMorten:
### Which version of the app?

WinUI 3 Gallery

### Description

On the Iconography page it's too easy to go pick an icon glyph and add it to your app, not realizing it won't work on Windows 10. For instance the following will render a square placeholder on Windows 10:
```xml
<FontIcon Glyph="&#xEBD3;" />
```

However this will render fine:
```xml
<FontIcon Glyph="&#xE896;" />
```
With so many users still on Win10, it needs to be made easier to avoid making such mistakes, since it can get caught too late in the dev process if the developers are using Win11 machines.

### Screenshots

No indication that this isn't supported on Windows 10:
![image](https://github.com/microsoft/WinUI-Gallery/assets/1378165/b3e1647f-bdc4-411a-9ac5-fd5195fbaac1)
This one right next to it though is supported on Windows 10:
![image](https://github.com/microsoft/WinUI-Gallery/assets/1378165/a87f21e2-2f35-48c8-86b1-980e252e31e3)

### Windows version

Windows 10 22H2 (19045)

marcelwgn:
Thats a good point, would you expect there just to be a "Windows 10" disclaimer or also depending on the Windows build when it was available?

niels9001:
<img width="752" alt="image" src="https://github.com/microsoft/WinUI-Gallery/assets/9866362/8af3e5a1-1202-479d-83b2-978b48401fc4">

It's called out in the Expander. 

But I agree, it doesn't jump out.

Maybe we could add an `InfoBar` at the top calling this out and warning folks that some icons might not render at all and testing is required?

dotMorten:
That message isn’t helpful. Many of the icons do work on Win10 but not all

Jay-o-Way:
Interesting. I wasn't aware some icons wouldn't be available in the MDL2 font. I wonder how many of them are missing. In my opinion, I think the code area (for the specific icon)would be the most appropriate place.

![Screenshot_20240607-001253~2.png](https://github.com/microsoft/WinUI-Gallery/assets/65828559/232b6ff0-684e-486f-80db-e5946cdf9e3d)

dotMorten:
If they are on this page they are supported: https://learn.microsoft.com/en-us/windows/apps/design/style/segoe-ui-symbol-font
[...]

PR #1921, opened 2025-07-15
Remove the expander from the header of the `IconographyPage`

### LINKED ISSUE #1903: Icons panel becomes inaccessible when window height is minimized and info card is expanded
**Steps to Reproduce:**

1. Open the *Iconography* page.
2. Resize the window height to its minimum.
3. Expand the info card at the top of the page.

**Expected Result:**
The icons panel remains visible or accessible via scrolling.

**Actual Result:**
The icons panel becomes hidden, and there's no way to scroll to it.

**Screenshots**
Image

Zakariathr22:
## Description
Remove the expander from the header of the `IconographyPage` as suggested in https://github.com/microsoft/WinUI-Gallery/issues/1858#issuecomment-3070528069

## Motivation and Context
- Fixes #1903 

## How Has This Been Tested?
Manually tested

image

## Types of changes


Zakariathr22:
@niels9001 Could you please review this?

niels9001 (changes requested):
Would it make sense to add a description for this page in the json? E.g. something like:

"The icons below are part of Segoe Fluent Icons, the default icon set in Windows 11."

We can then remove the 'Fluent Icons Library' textblock too, above the searchbox.

Zakariathr22:
> The icons below are part of Segoe Fluent Icons, the default icon set in Windows 11.

The problem is that font icons use the `SymbolThemeFontFamily` resource as their font family, which maps to Segoe Fluent Icons on Windows 11 and Segoe MDL2 Icons on Windows 10.

https://github.com/microsoft/WinUI-Gallery/blob/2762188425728b2ce2f8c548dcf09c96c695c2b4/WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml#L62

What about adding this?
"The icons below use Segoe Fluent Icons on Windows 11 and Segoe MDL2 Icons on Windows 10."

marcelwgn:
I agree with @niels9001. I think we can use your proposed sentence @Zakariathr22 and add something like "For more information, visit https://learn.microsoft.com/windows/apps/design/signature-experiences/iconography#system-icons".

Zakariathr22:
> I agree with @niels9001. I think we can use your proposed sentence @Zakariathr22 and add something like "For more information, visit https://learn.microsoft.com/windows/apps/design/signature-experiences/iconography#system-icons".

Done!

niels9001:
> > I agree with @niels9001. I think we can use your proposed sentence @Zakariathr22 and add something like "For more information, visit https://learn.microsoft.com/windows/apps/design/signature-experiences/iconography#system-icons".
> 
> Done!

1) Does all icons work with MDL2? I think a few icons are SFI only? Wouldn't we want devs to use SFI on W10, too?
2) With W10 end of support coming up, how relevant is this still?
3) Can't we set this as the description in the JSON vs. adding a TextBlock in XAML? XAMLStyler doesn't play nice with Hyperlinks, and it's a PITA when working on this page.

Zakariathr22:
> 1. Does all icons work with MDL2? I think a few icons are SFI only? Wouldn't we want devs to use SFI on W10, too?

This leads us to #1557. It might be better for Windows 10 developers to use SFI, but what about Windows 10 users who will actually run the app?

My suggestion is to open another PR to warn developers about icons that won’t be available by default for Windows 10 users.

> 2. With W10 end of support coming up, how relevant is this still?

We don’t know the exact end date for Windows 10 support, but MDL2 is mentioned in the official documentation, which is the main reason we're referencing it: https://learn.microsoft.com/windows/apps/design/signature-experiences/iconography#system-icons

> 3. Can't we set this as the description in the JSON vs. adding a TextBlock in XAML? XAMLStyler doesn't play nice with Hyperlinks, and it's a PITA when working on this page.

We are using `RichTextBlock` only for the hyperlink. I think we can remove it since there's already a Docs button in the page header, and use the JSON description instead.

niels9001:
@Zakariathr22 yeah that makes sense:

~~I guess the cleaneast way would be to make this OS dependent: in the icon config file add a property that shows if the icon is (not) available on Win10, and then just hide these altogether.

But that might be complicated to maintain.~~

Actually, no, we shouldn't hide these but flag to the dev that these icons might not work on W10 when SFI is not installed and using MDL2. We do this for certain WinUI brushes too that are unavailable.

Zakariathr22:
> Actually, no, we shouldn't hide these but flag to the dev that these icons might not work on W10 when SFI is not installed and using MDL2. We do this for certain WinUI brushes too that are unavailable.

Should I include this in the current PR or create a separate one?

niels9001:
> > Actually, no, we shouldn't hide these but flag to the dev that these icons might not work on W10 when SFI is not installed and using MDL2. We do this for certain WinUI brushes too that are unavailable.
> 
> Should I include this in the current PR or create a separate one?

We could do it in a seperate one, maybe? Whatever works for you!

Zakariathr22:
> We could do it in a seperate one, maybe? Whatever works for you!

- #1925

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -89,65 +89,26 @@
...
-            </toolkit:SettingsExpander.Items>
-        </toolkit:SettingsExpander>
-
         <StackPanel
-            Grid.Row="1"
-            Margin="0,24,0,0"
+            Grid.Row="0"
             Spacing="8">
-            <TextBlock Style="{ThemeResource BodyStrongTextBlockStyle}" Text="Fluent Icons Library" />
+            <RichTextBlock TextWrapping="Wrap"
```
niels9001: If we don't use a Hyperlink anymore, we could set this string in the page description in the json, right?

niels9001 (changes requested):

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -89,65 +89,14 @@
...
-                </toolkit:SettingsCard>
-            </toolkit:SettingsExpander.Items>
-        </toolkit:SettingsExpander>
-
-        <StackPanel
-            Grid.Row="1"
-            Margin="0,24,0,0"
-            Spacing="8">
-            <TextBlock Style="{ThemeResource BodyStrongTextBlockStyle}" Text="Fluent Icons Library" />
+        <StackPanel Grid.Row="0"
```
niels9001: StackPanel is no longer needed, right? Just 2 RowDefinitions (auto, *) for the searchbox + icons view?

WinUIGallery/Samples/Data/ControlInfoData.json
```diff
@@ -207,6 +207,7 @@
           "Title": "Iconography",
           "IsUpdated": true,
           "Subtitle": "Icons are a visual design language that can be used to communicate information quickly and effectively.",
+          "Description": "The icons below use Segoe Fluent Icons on Windows 11 and Segoe MDL2 Icons.",
```
niels9001:
```suggestion
          "Description": "The icons below use Segoe Fluent Icons on Windows 11 and Segoe MDL2 Assets on Windows 10.",
```

niels9001 (changes requested):

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -89,65 +89,14 @@
...
+                    Spacing="8"
+                    Margin="0,8,0,0">
             <AutoSuggestBox
```
niels9001:
```suggestion
             <AutoSuggestBox Margin="0,16,0,0"
```

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
Zakariathr22: > StackPanel is no longer needed, right? Just 2 RowDefinitions (auto, *) for the searchbox + icons view?

Yeah, I didn't notice this

WinUIGallery/Samples/Data/ControlInfoData.json
```diff
@@ -207,15 +207,20 @@
...
-              "Title": "Iconography in Windows 11",
+              "Title": "Iconography in Windows",
               "Uri": "https://learn.microsoft.com/windows/apps/design/signature-experiences/iconography#system-icons"
             },
             {
               "Title": "Segoe Fluent Icons font",
               "Uri": "https://learn.microsoft.com/windows/apps/design/style/segoe-fluent-icons-font"
+            },
+            {
+              "Title": "Segoe MDL2 Assets icons",
```
niels9001:
```suggestion
              "Title": "Segoe MDL2 Assets font",
```

niels9001:
@zakariathr22 Looks like some UI tests are failing 🤔

Zakariathr22:
> @zakariathr22 Looks like some UI tests are failing 🤔

Interesting 🤔 I just removed the Expander and a row from the Grid

marcelwgn:
Probably a pipeline hickup, let me rerun the pipeline. Btw, you can also run those tests locally to verify that yourself (albeit running all tests takes some time).

PR #1923, opened 2025-07-17
Automation property added for the info button in fundamentals\resources page - source popup.

sabareesanms:
**Bug Link:** 
[[WinUI 3 Gallery: Fundamentals: Resources]: Name property is not defined for Informational button icon present inside the 'Source code' popup](https://microsoft.visualstudio.com/OS/_workitems/edit/58151866)

**Description:** 
The **info icon** in the source popup in the **fundamentals\resources page** doesn't have an automation property - name which restricts the user who uses narrator to use the button. This is happening across the entire application. 

**Code changes**
Changes made in the file - WinUIGallery/Controls/PageHeader.xaml , `AutomationProperties.Name` is set to `"Info"` for the button for both Sample page source code and Control source code. 

```
<Button
          Padding="6,5,6,6"
          Style="{ThemeResource SubtleButtonStyle}"
          ToolTipService.ToolTip="Source code of this sample page in the WinUI Gallery repository"
          AutomationProperties.Name="Info">
```
**Note**
Since this is a page header custom control and used across all pages, fixing at the control level will fix the issue across the entire application.

**Testing**
Tested the changes with the Accessibility insights.

**Screenshots**
Screenshot 2025-07-17 162520

### DIFF
--- a/WinUIGallery/Controls/PageHeader.xaml
+++ b/WinUIGallery/Controls/PageHeader.xaml
@@ -153,7 +153,8 @@
                                         <Button
                                             Padding="6,5,6,6"
                                             Style="{ThemeResource SubtleButtonStyle}"
-                                            ToolTipService.ToolTip="Source code of this control in the WinUI repository. For some controls only the XAML file is available">
+                                            ToolTipService.ToolTip="Source code of this control in the WinUI repository. For some controls only the XAML file is available"
+                                            AutomationProperties.Name="Info">
                                             <FontIcon
                                                 VerticalAlignment="Center"
                                                 FontSize="14"


WinUIGallery/Controls/PageHeader.xaml
```diff
@@ -185,7 +185,8 @@
                                     <Button
                                         Padding="6,5,6,6"
                                         Style="{ThemeResource SubtleButtonStyle}"
-                                        ToolTipService.ToolTip="Source code of this sample page in the WinUI Gallery repository">
+                                        ToolTipService.ToolTip="Source code of this sample page in the WinUI Gallery repository"
+                                        AutomationProperties.Name="Info">
```
ayushjai30, line 190: Should we add more descriptive Name like "View Source code"? The label ‘Info’ is too generic and doesn’t convey what the button actually does.

ayushjai30, line 190: The bug mentions that the issue exists across the entire application. Does this fix resolve the problem throughout the app, or is it limited to just the Resource page?

Zakariathr22:
The issue isn’t within the Resources page itself, but rather in the PageHeader control

akanpatel2206:
Does this fix everywhere?
image

WinUIGallery/Controls/PageHeader.xaml
Hemantxk: Should it even be a button at the base or just the icon?

akanpatel2206: Unlike the first one, 
2nd and 3rd don't seem to do anything. 
image
They can be icons. 
Also, do we have this issue for 2nd one too? I believe the fix in this PR only fixes the 3rd one.

WinUIGallery/Controls/PageHeader.xaml
Hemantxk, line 1: AutomationProperties.Name needs to be added for Button on line 153 too. You can see that info icon for AutoSuggestBox & try your fix.

WinUIGallery/Controls/PageHeader.xaml
Hemantxk: Yeah, button on line 153 is the one that is showing Icon 2 in your image, I have added a separate comment. Couldn't comment in-line so added a comment on file.

sabareesanms: @ayushjai30  View source code would be misleading. The expected behavior they conveyed is when user navigating to it, screen reader should announce as "<label name>,<control type>". Since it has the icon, we are conveying what icon it is.

WinUIGallery/Controls/PageHeader.xaml
sabareesanms: Addressed the change mentioned.

WinUIGallery/Controls/PageHeader.xaml
sabareesanms: @ayushjai30  It is fixed across the entire application. Updated description accordingly.

sabareesanms:
> The issue isn’t within the Resources page itself, but rather in the PageHeader control

Updated description. It will be fixed across entire application as we fix this in PageHeader control.

sabareesanms:
> Does this fix everywhere? image

Yes. As the fix is at control level, it will fix across all pages.

WinUIGallery/Controls/PageHeader.xaml
niels9001: FYI - they should be icons because we got flagged in the past by the a11y team that these icons should be reached by keyboard to show its tooltip.

PR #1925, opened 2025-07-20
Add warning for Fluent-only icons not available on Windows 10

### LINKED ISSUE #1557: Icons don't warn about Win10 not supported

Zakariathr22:
## Description
Added a warning to indicate that some icons may not be available by default on Windows 10, especially those only included in the **Segoe Fluent Icons** set used by Windows 11. Main changes are:
- Added `IsSegoeFluentOnly` property to both the `IconDataSchema` and `IconData` class.
- Marked Fluent-only icons in the JSON data.
- Displayed a warning icon and message on the Iconography page for Fluent-only icons.

## Motivation and Context
- This change helps clarify icon availability across different Windows versions.
- Closes #1557.

## How Has This Been Tested?
Tested by manually setting the font to **Segoe MDL2 Assets** to simulate the default icon set on Windows 10 and verifying which icons fail to appear.

image

## Types of changes

### DIFF
--- a/WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
@@ -52,6 +52,15 @@
                     BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
                     BorderThickness="1"
                     CornerRadius="{StaticResource ControlCornerRadius}">
+
+                    <FontIcon FontSize="16"
+                              Margin="8"
+                              Foreground="{ThemeResource SystemFillColorCriticalBrush}"
+                              Glyph="&#xE7BA;"
+                              HorizontalAlignment="Right"
+                              VerticalAlignment="Top"
+                              Visibility="{Binding IsSegoeFluentOnly}"/>
+
                     <!--  Icon  -->
                     <Viewbox
                         Width="28"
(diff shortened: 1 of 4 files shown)


WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -52,6 +52,15 @@
                     BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
                     BorderThickness="1"
                     CornerRadius="{StaticResource ControlCornerRadius}">
+
+                    <FontIcon FontSize="16"
```
niels9001: Do we need this in the little cards itself? Or do we only show it in the side panel?

My thinking is that most devs wouldn't run into this, so it might not be crucial information. A dev would always select the card to get the right code / glyphs / symbolname, so showing it there might suffice?

Showing it here might require us to make it selectable for a11y purposes and clutters the UI a bit?

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -192,19 +201,42 @@
...
+                                        CornerRadius="{StaticResource ControlCornerRadius}">
+                                    <FontIcon FontFamily="{StaticResource SymbolThemeFontFamily}"
+                                              FontSize="48"
+                                              Glyph="{Binding SelectedItem.Character, Mode=OneWay}" />
+                                </Border>
+                                <Grid Grid.Column="1"
+                                      HorizontalAlignment="Stretch"
+                                      Visibility="{Binding SelectedItem.IsSegoeFluentOnly}">
+                                    <Grid.ColumnDefinitions>
+                                        <ColumnDefinition Width="auto" />
```
niels9001:
```suggestion
                                        <ColumnDefinition Width="Auto" />
```

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -192,19 +201,42 @@
...
-                                CornerRadius="{StaticResource ControlCornerRadius}">
-                                <FontIcon
-                                    FontFamily="{StaticResource SymbolThemeFontFamily}"
-                                    FontSize="48"
-                                    Glyph="{Binding SelectedItem.Character, Mode=OneWay}" />
-                            </Border>
+                            <Grid Margin="0,0,0,24"
+                                  HorizontalAlignment="Stretch">
+                                <Grid.ColumnDefinitions>
+                                    <ColumnDefinition Width="auto" />
```
niels9001:
```suggestion
                                    <ColumnDefinition Width="Auto" />
```

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -192,19 +201,42 @@
...
+                                        <ColumnDefinition />
+                                    </Grid.ColumnDefinitions>
+                                    <FontIcon FontSize="16"
+                                              Margin="0,6,8,0"
+                                              Foreground="{ThemeResource SystemFillColorCriticalBrush}"
+                                              Glyph="&#xE7BA;"
+                                              VerticalAlignment="Top" />
+                                    <TextBlock Grid.Column="1"
+                                               Margin="0,0,8,0"
+                                               Text="This icon may not be available by default in Windows 10."
```
niels9001: Is this statement correct and specific enough?

Would it make sense to shorten the warning an provide a tooltip? That way we provide all the necessary information and solve this issue once and for all :).

Warning:
"Only supported in Segoe Fluent Icons"..

Tooltip could be something like:
"This icon is only available in Segoe Fluent Icons — the default icon font on Windows 11. On Windows 10, MDL2 is used by default, so the icon may not appear unless Segoe Fluent Icons is installed."

Thoughts?

niels9001 (changes requested):

niels9001:
Amazing work as always! Added a few suggestions

Would love to get @dotmorten 's feedback too

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
Zakariathr22: > Do we need this in the little cards itself? Or do we only show it in the side panel?
> 
> My thinking is that most devs wouldn't run into this, so it might not be crucial information. A dev would always select the card to get the right code / glyphs / symbolname, so showing it there might suffice?
> 
> Showing it here might require us to make it selectable for a11y purposes and clutters the UI a bit?

I agree

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
Zakariathr22: > Is this statement correct and specific enough?
> 
> Would it make sense to shorten the warning an provide a tooltip? That way we provide all the necessary information and solve this issue once and for all :).
> 
> Warning: "Only supported in Segoe Fluent Icons"..
> 
> Tooltip could be something like: "This icon is only available in Segoe Fluent Icons — the default icon font on Windows 11. On Windows 10, MDL2 is used by default, so the icon may not appear unless Segoe Fluent Icons is installed."
> 
> Thoughts?

Updated as suggested!

WinUIGallery/Samples/ControlPages/Design/IconographyPage.xaml
```diff
@@ -192,19 +193,44 @@
...
+                                        BorderThickness="1"
+                                        CornerRadius="{StaticResource ControlCornerRadius}">
+                                    <FontIcon FontFamily="{StaticResource SymbolThemeFontFamily}"
+                                              FontSize="48"
+                                              Glyph="{Binding SelectedItem.Character, Mode=OneWay}" />
+                                </Border>
+                                <Grid Grid.Column="1"
+                                      HorizontalAlignment="Stretch"
+                                      Visibility="{Binding SelectedItem.IsSegoeFluentOnly}"
+                                      ToolTipService.ToolTip="This icon is only available in Segoe Fluent Icons (the default icon font on Windows 11). On Windows 10, MDL2 is used by default, so the icon may not appear unless Segoe Fluent Icons is installed.">
```
niels9001:
```suggestion
                                      ToolTipService.ToolTip="This icon is only available in Segoe Fluent Icons (the default icon font on Windows 11). On Windows 10, Segoe MDL2 Assets is used by default, so the icon may not appear unless Segoe Fluent Icons is installed.">
```

niels9001 (approved):

PR #1930, opened 2025-07-23
Modified custom password box implementation and added automation peer for narrator announcement.

sabareesanms:
**Bug:** [[WinUI 3 Gallery: Fundamentals-> Custom &user controls]: Screen reader fails to announce appeared suggestions to the user.](https://microsoft.visualstudio.com/OS/_workitems/edit/58168275)

**Description:** 
Custom password box in Custom & user controls page failed to announce suggestion via screen reader.

**Fix:** 
The current design of the password suggestion box works with changing the visibility of the text blocks. However, this design doesn't allow the combined text to be provided for screen reader for announcements and also ` AutomationProperties.LiveSetting` doesn't work with visibility change. So, the current design is migrated with `RichEditTextBlock` with ` AutomationProperties.LiveSetting` for screen reader to announce the text changes in the password change event.

**File Changes:**
```
WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.xaml
```
**Code Changes:**

- Switched from multiple visible StackPanel elements to a RichTextBlock
- Added message builder for the RichTextBlock to display the validation messages along with the icons and text colors.
- Added calls to `FrameworkElementAutomationPeer.RaiseAutomationEvent(AutomationEvents.LiveRegionChanged) `after validation updates to notify screen readers of content changes.
- Fixed redundant event triggers causing duplicate narrator announcements by removing unwanted update calls.

**Tests:**
Tested the changes via NarratorBuddy and Narrator.

**Screenshots:**
Screenshot 2025-07-23 164721

### DIFF
--- a/WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
+++ b/WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
@@ -2,7 +2,10 @@
 // Licensed under the MIT License.
 
 using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Automation.Peers;
 using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Documents;
+using Microsoft.UI.Xaml.Media;
 using System.Linq;
 
 namespace WinUIGallery.Samples.ControlPages.Fundamentals.Controls;
@@ -86,7 +89,6 @@ public sealed class ValidatedPasswordBox : Control
             PasswordInput.PasswordChanged += (sender, e) =>
             {
                 Password = PasswordInput.Password;
-                UpdateValidationMessages();
             };
         }
 
@@ -106,26 +108,59 @@ public sealed class ValidatedPasswordBox : Control
         bool hasNumber = Password.Any(char.IsDigit);
 
         IsValid = hasMinLength && hasUppercase && hasNumber;
-
-        if (MissingUppercaseText is not null)
+        var validationRichText = GetTemplateChild("ValidationRichText") as RichTextBlock;
+        if (validationRichText != null)
         {
-            MissingUppercaseText.Visibility = (hasUppercase || string.IsNullOrEmpty(Password)) ? Visibility.Collapsed : Visibility.Visible;
-        }
+            validationRichText.Blocks.Clear();
 
-        if (MissingNumberText is not null)
-        {
-            MissingNumberText.Visibility = (hasNumber || string.IsNullOrEmpty(Password)) ? Visibility.Collapsed : Visibility.Visible;
-        }
-
-        if (TooShortText is not null)
-        {
-            TooShortText.Visibility = (hasMinLength || string.IsNullOrEmpty(Password)) ? Visibility.Collapsed : Visibility.Visible;
-        }
-
-        if (ValidPasswordText is not null)
-        {
-            ValidPasswordText.Visibility = IsValid ? Visibility.Visible : Visibility.Collapsed;
+            Paragraph paragraph = new();
+            if (!hasUppercase && !string.IsNullOrEmpty(Password))
+            {
+                AddValidationLine(paragraph, "\uEA39", "Missing uppercase", "SystemFillColorCriticalBrush");
+            }
+            if (!hasNumber && !string.IsNullOrEmpty(Password))
+            {
+                AddValidationLine(paragraph, "\uEA39", "Missing number", "SystemFillColorCriticalBrush");
+            }
+            if (!hasMinLength && !string.IsNullOrEmpty(Password))
+            {
+                AddValidationLine(paragraph, "\uEA39", "Too short!", "SystemFillColorCriticalBrush");
+            }
+            if (IsValid)
+            {
+                AddValidationLine(paragraph, "\uE930", "Password is valid", "SystemFillColorSuccessBrush");
+            }
+            if (paragraph.Inlines.Count > 0)
+            {
+                validationRichText.Blocks.Add(paragraph);
+                var peer = FrameworkElementAutomationPeer.FromElement(validationRichText)
+                            ?? FrameworkElementAutomationPeer.CreatePeerForElement(validationRichText);
+                peer.RaiseAutomationEvent(AutomationEvents.LiveRegionChanged);
+            }
         }
     }
+    private void AddValidationLine(Paragraph paragraph, string iconGlyph, string message, string resourceBrushKey)
+    {
+        var icon = new FontIcon
+        {
+            Glyph = iconGlyph,
+            FontSize = 14,
+            Foreground = (Brush)Application.Current.Resources[resourceBrushKey]
+        };
+
+        InlineUIContainer iconContainer = new() { Child = icon };
+
+        paragraph.Inlines.Add(iconContainer);
+        paragraph.Inlines.Add(new Run { Text = " " });
+
+        paragraph.Inlines.Add(new Run
+        {
+            Text = message,
+            Foreground = (Brush)Application.Current.Resources[resourceBrushKey]
+        });
+
+        paragraph.Inlines.Add(new LineBreak());
+    }
 }
+
 
--- a/WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.xaml
+++ b/WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.xaml
@@ -10,50 +10,10 @@
                 <ControlTemplate TargetType="local:ValidatedPasswordBox">
                     <StackPanel Spacing="4">
                         <PasswordBox x:Name="PasswordInput" />
-
-                        <!-- Uppercase Validation -->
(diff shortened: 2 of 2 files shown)


WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -106,26 +108,59 @@ private void UpdateValidationMessages()
...
-        if (ValidPasswordText is not null)
+        paragraph.Inlines.Add(iconContainer);
+        paragraph.Inlines.Add(new Run { Text = " " });
+
+        paragraph.Inlines.Add(new Run
         {
-            ValidPasswordText.Visibility = IsValid ? Visibility.Visible : Visibility.Collapsed;
-        }
+            Text = message,
+            Foreground = (Brush)Application.Current.Resources[resourceBrushKey]
```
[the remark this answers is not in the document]
sabareesanms: resourceBrushKey is set as a constant string from the caller function.

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -106,26 +108,59 @@ private void UpdateValidationMessages()
...
-        }
+            validationRichText.Blocks.Clear();
 
-        if (MissingNumberText is not null)
-        {
-            MissingNumberText.Visibility = (hasNumber || string.IsNullOrEmpty(Password)) ? Visibility.Collapsed : Visibility.Visible;
+            Paragraph paragraph = new();
+            if (!hasUppercase && !string.IsNullOrEmpty(Password))
+            {
+                AddValidationLine(paragraph, "\uEA39", "Missing uppercase", "SystemFillColorCriticalBrush");
```
snigdha011997: Can this string be more descriptive something like "Missing uppercase character" ?

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -106,26 +108,59 @@ private void UpdateValidationMessages()
...
+            }
+            if (!hasNumber && !string.IsNullOrEmpty(Password))
+            {
+                AddValidationLine(paragraph, "\uEA39", "Missing number", "SystemFillColorCriticalBrush");
+            }
+            if (!hasMinLength && !string.IsNullOrEmpty(Password))
+            {
+                AddValidationLine(paragraph, "\uEA39", "Too short!", "SystemFillColorCriticalBrush");
+            }
```
snigdha011997: Is there a way to put them as part of resource dictionary?

sabareesanms: It's already part of resource dictionary. We are using it.

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
sabareesanms: I haven't made any changes to the content moving it from XAML.

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -106,26 +108,59 @@ private void UpdateValidationMessages()
...
+            if (IsValid)
+            {
+                AddValidationLine(paragraph, "\uE930", "Password is valid", "SystemFillColorSuccessBrush");
```
godlytalias, line 119: If we keep strings like this, it won'be able to be localized. Better to give sample such a way that people can achieve localization as well easily

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -86,7 +89,6 @@ protected override void OnApplyTemplate()
             PasswordInput.PasswordChanged += (sender, e) =>
             {
                 Password = PasswordInput.Password;
-                UpdateValidationMessages();
```
godlytalias, line 89: Why we are removing updation on password changed event?

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
sabareesanms: Localization is not handled in the gallery app. The app is currently supported only in English.

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
sabareesanms: The update validation message is subscribed to password property. So, adding it again with password changed event triggers the password changed event twice.

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -106,26 +108,59 @@ private void UpdateValidationMessages()
...
         {
-            MissingUppercaseText.Visibility = (hasUppercase || string.IsNullOrEmpty(Password)) ? Visibility.Collapsed : Visibility.Visible;
-        }
+            validationRichText.Blocks.Clear();
 
-        if (MissingNumberText is not null)
-        {
-            MissingNumberText.Visibility = (hasNumber || string.IsNullOrEmpty(Password)) ? Visibility.Collapsed : Visibility.Visible;
+            Paragraph paragraph = new();
+            if (!hasUppercase && !string.IsNullOrEmpty(Password))
```
godlytalias: Isn't it better to keep all these as 'else' part of `if (IsValid)`

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -106,26 +108,59 @@ private void UpdateValidationMessages()
...
+
+        paragraph.Inlines.Add(new Run
         {
-            ValidPasswordText.Visibility = IsValid ? Visibility.Visible : Visibility.Collapsed;
-        }
+            Text = message,
+            Foreground = (Brush)Application.Current.Resources[resourceBrushKey]
+        });
+
+        paragraph.Inlines.Add(new LineBreak());
```
godlytalias, line 165: Should we add linebreak only when adding a new line? Or else will it affect RichEditBox height?

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -106,26 +108,59 @@ private void UpdateValidationMessages()
         bool hasNumber = Password.Any(char.IsDigit);
 
         IsValid = hasMinLength && hasUppercase && hasNumber;
-
-        if (MissingUppercaseText is not null)
+        var validationRichText = GetTemplateChild("ValidationRichText") as RichTextBlock;
```
godlytalias, line 111: Whether these are documented ones? Else if we put in sample and applications start to follow this. Later if in case the name changes, applications will start to break.

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
sabareesanms: Updated

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
sabareesanms: We added newline here, so that the next added line will be added in new line rather than at the same line of the text.

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
sabareesanms: The ValidatedPasswordBox is already in the gallery app. We don't add documentation for samples for custom controls. I followed the naming convention for the RichTextBlock

WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
godlytalias: But the question is why we add newline before a new line coming in. We need to add it only at the point a newline getting added.

sabareesanms: Fixed in this PR https://github.com/microsoft/WinUI-Gallery/pull/1931

PR #2001, opened 2025-07-30
Refactor Application Settings with AppConfig and JSON Storage Support for Packaged/Unpackaged Apps #1924

### LINKED ISSUE #1924: UnPackaged Mode is Broken
### Which version of the app?

WinUI 3 Gallery

### Description

**Unpackaged mode is broken** due to the use of `LocalSettings` for saving app settings, which is not supported in unpackaged apps.

I’m considering two possible solutions:

1. Implement a simple class to serialize/deserialize a model to a JSON file — straightforward, with no extra features.
2. Use [Nucs.JsonSettings](https://github.com/Nucs/JsonSettings), which offers features like:

   * Versioning
   * AutoSave
   * And more...

I’m ready to submit a PR.

### Screenshots

_No response_

### Windows version

_No response_

### Additional context

_No response_

ghost1372:
This PR replaces the legacy ApplicationData-based settings system with a new AppConfig class that serializes settings to JSON files. It introduces support for both packaged and unpackaged app scenarios.

Closes #1924.

Old Settings System:

`appData.LocalSettings.Values[SettingsKeys.IsLeftMode] = true;
`

New Settings System:
```
SettingsHelper.Config.IsLeftMode = true;
SettingsHelper.Save();
```

image

## Types of changes


WinUIGallery/Helpers/SettingsHelper/AppConfig.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// <summary>
+    /// Represents the maximum number of recently visited samples to retain.
+    /// </summary>
+    public int MaxRecentlyVisitedSamples { get; set; } = 7;
+
+    /// <summary>
+    /// Gets or sets the maximum number of favorite samples allowed.
+    /// </summary>
+    public int MaxFavoriteSamples { get; set; } = 0;
```
Zakariathr22, line 52: There is no maximum for the favorite samples. The old logic assumed that if the maximum was set to `0`, it meant there was no limit. Since that logic has been removed, we no longer need this.

WinUIGallery/Helpers/SettingsHelper/AppConfig.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// <summary>
+    /// Gets or sets the list of recently visited items.
+    /// </summary>
+    public List<string> RecentlyVisited { get; set; } = new();
+
+    /// <summary>
+    /// Represents the maximum number of recently visited samples to retain.
+    /// </summary>
+    public int MaxRecentlyVisitedSamples { get; set; } = 7;
```
Zakariathr22, line 47:
```suggestion
    public const int MaxRecentlyVisitedSamples { get; set; } = 7;
```
This is supposed to be a constant

Zakariathr22 (changes requested):
Just a few suggested changes:
Also, I don't think there's any need to save `MaxRecentlyVisitedSamples` and `MaxFavoriteSamples` in the local JSON file.

Image

ghost1372:
> Just a few suggested changes: Also, I don't think there's any need to save `MaxRecentlyVisitedSamples` and `MaxFavoriteSamples` in the local JSON file.
> 
> Image

Thanks, @Zakariathr22 . I don’t have any issues with that, but let’s check with @marcelwgn as well.
Should we remove the `max` properties from the JSON file?

WinUIGallery/Helpers/SettingsHelper/AppConfig.cs
ghost1372: @Zakariathr22 
If we decide to exclude the `max` properties from the JSON file, we should move those properties out of `AppConfig.cs` and into the `SettingsHelper.cs` class.
so this suggestion can not be commited here.

marcelwgn:
After internal discussion, we decided that we want to stick with the ApplicationData as the default mechanism and fall back to JSON if we are running unpackaged, sorry for the change of direction here. For this, I think we should have a common "IPersistenceProvider"/"ISettingsProvider" interface that depending on the packaging mode gets filled by an ApplicationData or JSON file based provider. 

I agree with @Zakariathr22 that MaxRecentlyVisitedSamples and MaxFavoriteSamples should not be persisted, those are policies and not settings.

ghost1372:
> After internal discussion, we decided that we want to stick with the ApplicationData as the default mechanism and fall back to JSON if we are running unpackaged, sorry for the change of direction here. For this, I think we should have a common "IPersistenceProvider"/"ISettingsProvider" interface that depending on the packaging mode gets filled by an ApplicationData or JSON file based provider.
> 
> I agree with @Zakariathr22 that MaxRecentlyVisitedSamples and MaxFavoriteSamples should not be persisted, those are policies and not settings.


Okay, I think it's better to close this one and open a new PR with the requested changes.

ghost1372:
**I'm closing this PR and have opened a new one with the requested changes.**

https://github.com/microsoft/WinUI-Gallery/pull/2014

Issue #2003, opened 2025-07-31
Creating a child window on Windowing page throws System.ExecutionEngineException


AndrewKeepCoding:
### Which version of the app?

WinUI 3 Gallery

### Description

Clicking the `Create new Window` button in the **Windowing** page throws*System.ExecutionEngineException*.

### Screenshots

<img width="1226" height="689" alt="Image" src="https://github.com/user-attachments/assets/212d2f84-86a3-4677-9d51-1262e4fedae6" />

### Windows version

Windows 11 24H2 (26100)

AndrewKeepCoding:
@niels9001 The changes in #1908 is causing this. Before that the child window works.

So far, I found out that we shouldn't be accessing a static `MainWindow` here.

**HomePage.cs**
```cs
protected override void OnNavigatedTo(NavigationEventArgs e)
{
    ((NavigationViewItem)App.MainWindow.NavigationView.MenuItems.First()).IsSelected = true;
```

AndrewKeepCoding:
The issue isn’t with the navigation logic. It turns out the `Win32WindowHelper` class holds two static fields. That’s problematic when the class is used for a second Window, since those statics persist across instances.

PR #2005, opened 2025-07-31
Child window exception fix

AndrewKeepCoding:
## Description

This PR refactors the use of static App.MainWindow and the navigation logic in the entire app.

## Motivation and Context

Fixes: #2003 

## Types of changes

### DIFF
--- a/WinUIGallery/Helpers/NavigationViewExtensions.cs
+++ b/WinUIGallery/Helpers/NavigationViewExtensions.cs
@@ -1,0 +1,145 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.UI.Xaml.Controls;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace WinUIGallery.Helpers;
+
+/// <summary>
+/// Provides extension methods for <see cref="NavigationView"/> and <see cref="NavigationViewItem"/>
+/// to retrieve all navigation view items, including nested items.
+/// </summary>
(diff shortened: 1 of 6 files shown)


AndrewKeepCoding:
@niels9001 Can you take a quick view? Should I keep going?🙂

marcelwgn:
I think the general approach is fine except for some minor details but we can discuss those once its out of draft. Can you explain what caused the crash and how this PR would address the root cause of it?

AndrewKeepCoding:
> Can you explain what caused the crash and how this PR would address the root cause of it?

At first, I thought the issue was caused by the navigation logic, but I'm not so sure now. Just avoiding the call to this method stops the exception from happening.🤔

**Win32WindowHelper.cs**
```cs
public void SetWindowMinMaxSize(POINT? minWindowSize = null, POINT? maxWindowSize = null)
{
    this.minWindowSize = minWindowSize;
    this.maxWindowSize = maxWindowSize;

    var hwnd = GetWindowHandleForCurrentWindow(window);

    newWndProc = new WinProc(WndProc);
    oldWndProc = SetWindowLongPtr(hwnd, WindowLongIndexFlags.GWL_WNDPROC, newWndProc);
}
```

AndrewKeepCoding:
The issue isn’t with the navigation logic—it turns out the `Win32WindowHelper` class holds two static fields. That’s problematic when the class is used for a second Window, since those statics persist across instances.

I’m closing this PR for now. I’ll submit a revised PR that addresses the static field problem, along with another one to refactor the navigation logic.

PR #2012, opened 2025-08-06
[Settings] Tweaking removing recent/favorite samples UX

niels9001:
To avoid potential accessibility issues, switching from a flyout to a dialog when removing favorites or recents. While at it, I simplified the UI a bit.

image
image


marcelwgn (approved):
Excellent idea, thank you!

PR #2014, opened 2025-08-07
Refactor Application Settings and Support for Packaged/Unpackaged Mode #1924

### LINKED ISSUE #1924: UnPackaged Mode is Broken

ghost1372:
This PR replaces the legacy ApplicationData-based settings system with a new SettingsHelper class. It adds support for both packaged and unpackaged app scenarios.

Closes #1924.

**Old Settings System:**

`appData.LocalSettings.Values[SettingsKeys.IsLeftMode] = true; 
`
New Settings System:

`SettingsHelper.Current.IsLeftMode = true;
`
>[!Note]
>Settings are automatically saved when changed.

## Important Note on List Types:
Direct modification of list items won't persist changes:

`SettingsHelper.Current.myList.Add(item); // ❌ This won't work`

```
var list = SettingsHelper.Current.myList;
list.Add(item);
SettingsHelper.Current.myList = list; // ✅ This works
```

## SettingsHelper Architecture:
**Uses an ISettingsProvider abstraction.**
In packaged mode: uses **ApplicationDataContainer**.
In unpackaged mode: uses **JsonSettingsProvider**.

## Adding a New Setting:
```
public bool IsEnabled
{
    get => GetOrCreateDefault<bool>(true);
    set => Set(value);
}

```
## Supported Types:
In unpackaged mode (JSON), you can serialize most types easily.

In packaged mode, ApplicationDataContainer supports only limited types directly:

- Primitives
- string
- DateTimeOffset
- TimeSpan
- Guid
- Windows.Foundation.Point
- Windows.Foundation.Size
- Windows.Foundation.Rect

Other types (e.g., decimal, List<T>) are serialized to JSON before being stored.

## Vector Types:
Types like Vector2, Vector3, and Vector4 do not work in regular mode. Instead, use wrapper classes like Vector2Data, Vector3Data, and Vector4Data:

```
public System.Numerics.Vector2 Position2D
{
    get => GetOrCreateDefault<Vector2Data>(new Vector2Data { X = 0, Y = 0 });
    set => Set((Vector2Data)value);
}
```


WinUIGallery/Helpers/SettingsHelper/Internals/ListExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public static void AddToFirst(this List<string> list, string item, bool isFavorite)
+    {
+        if (string.IsNullOrWhiteSpace(item))
+            return;
+
+        list.Remove(item);
+        list.Insert(0, item);
+
+        if (!isFavorite && SettingsHelper.MaxRecentlyVisitedSamples > 0 && list.Count > SettingsHelper.MaxRecentlyVisitedSamples)
```
Zakariathr22: Can we make these methods more general for better reusability, in case we need to save other lists in the future that are not related to favorites or recently visited?

WinUIGallery/Helpers/SettingsHelper/Internals/Vectors.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace WinUIGallery.Helpers;
+
+internal partial class Vector4Data
```
Zakariathr22: I don't think we need these vectors, nor will we need them in the future, right?

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+        catch {}
+    }
+
+    public void Set<T>(string key, T value)
+    {
+        object storedValue = IsSimpleType(typeof(T))
+            ? value
+            : JsonSerializer.Serialize(value, SettingsJsonContext.Default.GetTypeInfo(typeof(T)));
+
+        container.Values[key] = storedValue;
```
Zakariathr22, line 66: The maximum size of data that can be stored for a key is 8 KB (if we are storing a string, that’s approximately 4,000 characters). So, if the user writes more than 4,000 characters in the scratchpad, it will fail and throw an exception. We need to handle this.

Image

WinUIGallery/Pages/HomePage.xaml.cs
```diff
@@ -31,34 +31,43 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+                    var favs = SettingsHelper.Current.Favorites;
+                    favs.Remove(id);
+                    SettingsHelper.Current.Favorites = favs;
+                }
+                else
+                {
+                    var recentlyVisited = SettingsHelper.Current.RecentlyVisited;
+                    recentlyVisited.Remove(id);
+                    SettingsHelper.Current.RecentlyVisited = recentlyVisited;
+                }
```
Zakariathr22, line 66: This block repeats logic for `favs` and `recentlyVisited`.
Consider extracting a common method to avoid duplication :)

WinUIGallery/Helpers/SettingsHelper/ObservableSettings.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System.ComponentModel;
+using System.Runtime.CompilerServices;
+
+namespace WinUIGallery.Helpers;
+
+public partial class ObservableSettings : INotifyPropertyChanged
```
Zakariathr22, line 6: Just a question:
Do we really need the settings to be observable in our case?
I think this is just adding an extra layer of logic. Any thoughts?

WinUIGallery/Pages/SettingsPage.xaml.cs
```diff
@@ -200,7 +203,9 @@ private async void ClearRecentBtn_Click(object sender, RoutedEventArgs e)
         };
         dialog.PrimaryButtonClick += (s, args) =>
         {
-            SettingsHelper.Delete(SettingsKeys.RecentlyVisited);
+            var recentlyVisited = SettingsHelper.Current.RecentlyVisited;
+            recentlyVisited.Clear();
+            SettingsHelper.Current.RecentlyVisited = recentlyVisited;
```
Zakariathr22: Same here, this repeats logic for favs and recentlyVisited

Zakariathr22 (commented):
Just a few points that I think are important, but of course after hearing @marcelwgn’s opinion about them.

WinUIGallery/Helpers/SettingsHelper/Internals/ListExtensions.cs
ghost1372: Hi @Zakariathr22 
Ahh… Excellent idea! How about something like this
`recentlyVisited.AddToFirst(item, SettingsHelper.MaxRecentlyVisitedSamples);`

WinUIGallery/Helpers/SettingsHelper/Internals/Vectors.cs
ghost1372: Currently, no — we don’t need them. My concern is that in the future, people might run into issues and, without realizing it, end up hacking the code to make it work.

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
ghost1372: Why do we even need to save and restore scratchpad data? We already have default data, and the user can reset it at any time without needing a save/restore mechanism.
I think we should remove the save/restore setting for the scratchpad. Even while writing this PR, I don’t understand why we’re saving and restoring the XAML content.

WinUIGallery/Helpers/SettingsHelper/ObservableSettings.cs
ghost1372: The reason we made them observable is to automatically notify the UI (or other parts of the app) when a setting changes, without having to manually raise events or rebind.

WinUIGallery/Pages/SettingsPage.xaml.cs
ghost1372: mutating a list in place (e.g., .Add() or .Clear()) never triggers Set() and therefore never writes back to storage.
To make SettingsHelper.Current.RecentlyVisited.Add(...) save automatically, we could store an observable collection (or any collection that raises change events) and hook into its events to persist changes.

WinUIGallery/Helpers/SettingsHelper/ObservableSettings.cs
Zakariathr22: Based on what **Michael Hawker** said in [this video (minute 9:25)](https://youtu.be/qGwzQiwGIx8), `INotifyPropertyChanged` is used to **produce events** that **notify the XAML** UI when a property has changed through **one-way binding**. In our case, the settings storing and retrieving logic has nothing to do with the XAML, and no property is being bound, so producing these events is not really necessary.
image

This logic is more suitable when retrieving and storing data from a JSON dataset that is bound directly to the UI, allowing the user to edit it in place.

WinUIGallery/Pages/SettingsPage.xaml.cs
Zakariathr22: I think using `ObservableCollection` will affect performance without any real benefit.
A simpler helper class that uses the provider would be more than sufficient, no `INotifyPropertyChanged` and no `ObservableCollection`, just straightforward logic.

WinUIGallery/Helpers/SettingsHelper/Internals/Vectors.cs
Zakariathr22: I mean how are we supposed to test something that isn’t even used? 🤷‍♂️

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
Zakariathr22: > Why do we even need to save and restore scratchpad data? We already have default data, and the user can reset it at any time without needing a save/restore mechanism. I think we should remove the save/restore setting for the scratchpad. Even while writing this PR, I don’t understand why we’re saving and restoring the XAML content.

IDK, let’s leave the decision to the maintainers on whether to keep it or remove it

WinUIGallery/Pages/HomePage.xaml.cs
ghost1372: Done

WinUIGallery/Pages/SettingsPage.xaml.cs
ghost1372: Done

WinUIGallery/Helpers/SettingsHelper/Internals/Vectors.cs
ghost1372: Done

WinUIGallery/Helpers/SettingsHelper/ObservableSettings.cs
ghost1372: That’s a fair point —In our case, even though we aren’t directly binding settings to XAML right now, I kept the observable pattern, If later we decide to bind certain settings directly to the UI (like theme, language, or toggles), we won’t need to refactor the settings system. The event mechanism is already there.

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
niels9001: Agreed. It's a nice-to-have to save stuff, but if it complicates things I'd be in favor of removing it.

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+        typeof(TimeSpan),
+        typeof(Guid),
+        typeof(Windows.Foundation.Point),
+        typeof(Windows.Foundation.Size),
+        typeof(Windows.Foundation.Rect)
+    };
+
+    private static bool IsSimpleType(Type type)
+    {
+        return type.IsPrimitive || ExtraSimpleTypes.Contains(type);
```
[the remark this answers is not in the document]
ghost1372: This review seems incorrect.
**ExtraSimpleTypes** is already **static readonly**.

WinUIGallery/Helpers/SettingsHelper/Providers/JsonSettingsProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            var json = File.ReadAllText(filePath);
+            values = JsonSerializer.Deserialize(
+                json,
+                SettingsJsonContext.Default.DictionaryStringJsonElement
+            ) ?? new();
+        }
+        catch (Exception)
+        {
+            HandleCorruptedFile();
+        }
```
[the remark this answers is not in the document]
ghost1372: Having multiple specific exceptions can be useful for clarity, but in this case, it may not be necessary. We could handle multiple error cases with a single exception type, which would allow us to catch all errors uniformly instead of only the predefined ones.

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+                return (T)JsonSerializer.Deserialize(str, typeInfo);
+            }
+
+            return (T)Convert.ChangeType(value, typeof(T));
+        }
+        catch (Exception)
+        {
+            HandleCorruptedKey(key);
+            return default;
+        }
```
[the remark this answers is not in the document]
ghost1372: Like above

ghost1372:
> Looks like there’s an issue when running the app unpackaged:
> 
> * Navigate to `App notification` page
> * Click `Show notification` button
> 
> The app crashes with an unhandled exception.

The `AppNotificationManager` throws an "Element Not Found" exception and is not related to this PR!
This crash can be easily fixed but should be addressed in a separate PR.

As described [here](https://learn.microsoft.com/en-us/windows/apps/develop/notifications/app-notifications/app-notifications-quickstart?tabs=cs), we need to register it. I tested it, and it works fine in the unpackaged scenario. So, this requires additional work and should be done in a separate PR (we can keep the current logic for packaged mode and add a registration/scenario for unpackaged mode). Here are the Microsoft docs and samples for unpackaged scenarios.
https://learn.microsoft.com/en-us/windows/apps/develop/notifications/app-notifications/app-notifications-quickstart?tabs=cs
https://github.com/microsoft/WindowsAppSDK-Samples/tree/main/Samples/Notifications/App/CsUnpackagedAppNotifications/CsUnpackagedAppNotifications

WinUIGallery/Helpers/SettingsHelper/Internals/ListExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Collections.Generic;
+
+namespace WinUIGallery.Helpers;
+
+internal static class ListExtensions
+{
+    public static void AddToFirst<T>(this List<T> list, T item, int? maxSize = null)
```
marcelwgn: nit: I think AddAsFirst and AddAsLast are more intuitive names.

WinUIGallery/Helpers/SettingsHelper/Providers/SettingsProviderFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        if (NativeMethods.IsAppPackaged)
+        {
+            return new ApplicationDataSettingsProvider(ApplicationData.GetDefault().LocalSettings);
+        }
+        else
+        {
+            var folder = Path.Combine(
+                Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
+                ProcessInfoHelper.ProductNameAndVersion
```
marcelwgn: I think a new version should not invalidate the settings, productname should be enough here.

WinUIGallery/MainWindow.xaml.cs
```diff
@@ -61,7 +61,7 @@ private void RootGridXamlRoot_Changed(XamlRoot sender, XamlRootChangedEventArgs
 
     private void SetWindowProperties()
     {
-#if DEBUG
+#if DEBUG || DEBUG_UNPACKAGED
```
marcelwgn, line 64: Is Debug_Unpackaged not also ifdeffing DEBUG?

WinUIGallery/Samples/ControlPages/Fundamentals/ScratchPadPage.xaml.cs
```diff
@@ -9,35 +9,17 @@
...
-        var xamlStr = ReadScratchPadXAMLinLocalSettings();
-
-        // If there is no stored XAML, load the default.
-        if (xamlStr == null || xamlStr.Trim().Length == 0)
-        {
-            xamlStr = GetDefaultScratchXAML();
-        }
-
-        m_oldText = xamlStr;
+        m_oldText = GetDefaultScratchXAML();
```
marcelwgn, line 22: Is it intentional we are no longer storing the xaml scratchpad content?

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+    public object Get(string key) => container.Values.TryGetValue(key, out var value) ? value : null;
+
+    public void Set(string key, object value) => container.Values[key] = value;
+
+    public T Get<T>(string key)
+    {
+        if (!container.Values.TryGetValue(key, out var value))
+            return default;
+
+        try
```
marcelwgn: I think this try catch should only be around the json deserializing.

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+            if (value is T t)
+                return t;
+
+            if (value is string str && !IsSimpleType(typeof(T)))
+            {
+                var typeInfo = SettingsJsonContext.Default.GetTypeInfo(typeof(T));
+                return (T)JsonSerializer.Deserialize(str, typeInfo);
+            }
+
+            return (T)Convert.ChangeType(value, typeof(T));
```
marcelwgn: This is a bit suspicious, are we actually hitting this?

WinUIGallery/Helpers/SettingsHelper/Internals/ListExtensions.cs
ghost1372: Done

WinUIGallery/Helpers/SettingsHelper/Providers/SettingsProviderFactory.cs
ghost1372: Done

WinUIGallery/Samples/ControlPages/Fundamentals/ScratchPadPage.xaml.cs
ghost1372: Yes, more info here
https://github.com/microsoft/WinUI-Gallery/pull/2014#discussion_r2278295294

WinUIGallery/MainWindow.xaml.cs
ghost1372: In Unpackaged Mode with the **Debug** profile, I always encounter COM registration exceptions.

image

However, if I use **Debug-Unpackaged**, the app works fine, but the '**Dev**' text does not appear unless I add` || DEBUG_UNPACKAGED`

image

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
ghost1372: Done

WinUIGallery/Helpers/SettingsHelper/Providers/ApplicationDataSettingsProvider.cs
ghost1372: Since it’s been a long time since these lines were added, I don’t remember the exact reason (I am very busy these days.😅), but it was probably to prevent crashes and handle settings stored by the previous system. If you prefer, I can replace it with an `InvalidCastException`.

ghost1372:
Hi @marcelwgn 
Thank you for your review.
I’ve addressed the suggestions you made earlier. 
I’ve added migration methods to preserve the user’s Favorites and Recently Visited items.

WinUIGallery/Helpers/SettingsHelper/SettingsMigration.cs
```diff
@@ -0,0 +1,83 @@
...
+/// Key points:
+/// - Only operates in packaged app mode (NativeMethods.IsAppPackaged).
+/// </summary>
+public partial class SettingsMigration
+{
+    private static ApplicationData appData = ApplicationData.GetDefault();
+
+    private const string RecentlyVisitedKey = "RecentlyVisited";
+    private const string FavoritesKey = "Favorites";
+    private const char delimiter = '\u001f';
```
marcelwgn, line 24: Nice, didnt know that existed!

PR #2017, opened 2025-08-12
Remove unnecessary merged dictionaries from Button.xaml

Zakariathr22:
## Description
Removed unnecessary `<ResourceDictionary.MergedDictionaries>` from Button.xaml.

## Motivation and Context
No longer required since WASDK 1.7.0.

## How Has This Been Tested?
Manually tested.

## Types of changes
- [x] Cleanup (non-breaking change removing obsolete code)


marcelwgn:
Thank you for cleaning that up! Do you happen to remember why we needed this in the first place? 😄

Zakariathr22:
> Thank you for cleaning that up! Do you happen to remember why we needed this in the first place? 😄

In that case, if the new style uses built-in WinUI resources, it won’t be found unless you merge the corresponding WinUI resource dictionary.
It was an ugly workaround, so I’m happy that this issue has been resolved in the newer versions.

PR #2023, opened 2025-08-18
Fix: TitleBar page changing window TitleBar buttons colors

### LINKED ISSUE #2022: TitleBar control in the “TitleBar” page unexpectedly changes the window TitleBar buttons color
### Which version of the app?

WinUI 3 Gallery

### Description

**Steps to Reproduce**
1. Set the App theme to be different from the System theme (e.g. if your system is using Light, set the app to Dark and vice-versa)
2. Navigate to the “TitleBar” page
3. Observe the window TitleBar button colors

### Screenshots

Image

### Windows version

_No response_

### Additional context

**Expected Result**
The TitleBar buttons should use the correct colors defined by the active app theme.

**Actual Result**
As soon as the “TitleBar” page is loaded, the window TitleBar buttons unexpectedly change to a wrong color.

**Note**
If the TitleBar control is removed from the page, this issue does not occur.

Zakariathr22:
## Description
The window TitleBar button colors were being changed when navigating to the TitleBar page while the app theme differed from the system theme.
The exact root cause couldn’t be identified, so the button colors were explicitly set in the control’s `LayoutUpdated` event (the last event triggered in the control’s lifecycle).
Setting the colors in the `Loaded` event didn’t work, whereas setting them during `LayoutUpdated` resolved the issue.

## Motivation and Context
- Closes #2022

## How Has This Been Tested?
Manually tested

## Types of changes

### DIFF
--- a/WinUIGallery/Helpers/TitleBarHelper.cs
+++ b/WinUIGallery/Helpers/TitleBarHelper.cs
@@ -19,7 +19,10 @@ internal partial class TitleBarHelper
 
     public static void SetCaptionButtonColors(Window window, Windows.UI.Color color)
     {
-        window.AppWindow.TitleBar.ButtonForegroundColor = color;
+        if (window.AppWindow != null)
+        {
+            window.AppWindow.TitleBar.ButtonForegroundColor = color;
+        }
     }
 
     public static void SetCaptionButtonBackgroundColors(Window window, Windows.UI.Color? color)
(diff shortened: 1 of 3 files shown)


marcelwgn:
Thank you for working on this. While this technically solves the issue, the flickering is ideal. Seems like that just having another TitleBar control causes this issue so I guess there isnt much we can do.

Zakariathr22:
> Thank you for working on this. While this technically solves the issue, the flickering is ideal. Seems like that just having another TitleBar control causes this issue so I guess there isnt much we can do.

I was thinking about opening an issue in the WinUI repo, but then I realized there’s no scenario where a user would need two TitleBars in a single window (except in this case 😆)

PR #2024, opened 2025-08-18
Fix Custom PasswordBox validation messages not updating with theme

### LINKED ISSUE #2019: Custom PasswordBox validation messages do not use theme resources
### Which version of the app?

WinUI 3 Gallery

### Description

On the Fundamentals->Custom & user controls page, the custom PasswordBox displays validation messages using static resources instead of theme resources. Since theme resources are not used at all, toggling the theme does not update the message colors, resulting in low contrast.

### Screenshots

Image

### Windows version

_No response_

### Additional context

_No response_

Zakariathr22:
## Description
Main Changes
- Validation messages now update according to the active theme
- `ValidationRichText` is collapsed when no password has been entered
- Updated the sample code accordingly

## Motivation and Context
- Closes #2019

## How Has This Been Tested?
Manually tested

image
image
image
image
image

## Types of changes

### DIFF
--- a/WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
+++ b/WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
@@ -7,6 +7,7 @@ using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Documents;
 using Microsoft.UI.Xaml.Media;
 using System.Linq;
+using Windows.UI;
 
 namespace WinUIGallery.Samples.ControlPages.Fundamentals.Controls;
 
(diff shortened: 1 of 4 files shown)


WinUIGallery/Samples/ControlPages/Fundamentals/Controls/ValidatedPasswordBox.cs
```diff
@@ -108,64 +74,55 @@ private void UpdateValidationMessages()
...
+    private void AddValidationLine(Paragraph paragraph, string iconGlyph, string message, bool isCritical)
+    {
+        if (paragraph.Inlines.Any())
+            paragraph.Inlines.Add(new LineBreak());
 
-        paragraph.Inlines.Add(iconContainer);
-        paragraph.Inlines.Add(new Run { Text = " " });
+        var color = new SolidColorBrush(
+            isCritical
+                ? (ActualTheme == ElementTheme.Light ? Color.FromArgb(255, 196, 43, 28) : Color.FromArgb(255, 255, 153, 164))
```
niels9001, line 117: Should we be using `SystemFillColorSuccessBrush` and `SystemFillColorCriticalBrush` to avoid hardcoding brushes that might not be consistent with the rest of the OS or meet a11y criteria?

Zakariathr22: > Should we be using `SystemFillColorSuccessBrush` and `SystemFillColorCriticalBrush` to avoid hardcoding brushes that might not be consistent with the rest of the OS or meet a11y criteria?

There is no direct way to use theme resources in C# (https://github.com/microsoft/microsoft-ui-xaml/issues/9464).
If we try to get theme values, they are retrieved as static resources.
The only known workaround is to create a minimal style and apply it to the UI component, but [`InlineUIContainer`](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.documents.inlineuicontainer?view=windows-app-sdk-1.7) does not have a `Style` property.
Therefore, the only option was to take the exact RGB colors from these resources and update the colors when the theme changes.

niels9001 (approved):

PR #2027, opened 2025-08-20
Adjust spacing on several pages and controls

### LINKED ISSUE #2025: Wrong spacing on section page when the window is resized to minimum width
Image

Zakariathr22:
## Motivation and Context
- Closes #2025

image
image
image
image
image
image

### DIFF
--- a/WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
@@ -51,7 +51,7 @@
             <SelectorBarItem Text="Signal" />
             <SelectorBarItem Text="High Contrast" />
         </SelectorBar>
-        <ScrollViewer Grid.Row="3" Margin="0,0,-12,0">
+        <ScrollViewer Grid.Row="3" Margin="0,0,-12,-36">
             <Frame x:Name="NavigationFrame" VerticalAlignment="Stretch" Padding="0,0,12,36"/>
         </ScrollViewer>
     </Grid>
--- a/WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
@@ -51,8 +51,8 @@
             <SelectorBarItem Text="Signal" />
             <SelectorBarItem Text="High Contrast" />
         </SelectorBar>
-        <ScrollViewer Grid.Row="3">
-            <Frame x:Name="NavigationFrame" VerticalAlignment="Stretch" />
+        <ScrollViewer Grid.Row="3" Margin="0,0,-12,0">
+            <Frame x:Name="NavigationFrame" VerticalAlignment="Stretch" Padding="0,0,12,36"/>
         </ScrollViewer>
     </Grid>
 </Page>
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -2,6 +2,7 @@ name: Bug report
 description: File a bug report
 title: "Bug title"
 labels: ["bug", "needs-triage"]
+type: Bug
 body:
   - type: dropdown
     validations:
--- a/.github/ISSUE_TEMPLATE/feature_request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yaml
@@ -2,6 +2,7 @@ name: Feature request
 description: Suggest a sample or enhancement for the WinUI Gallery
 title: "Sample request title"
 labels: ["needs-triage"]
+type: Feature
 body:
   - type: dropdown
     validations:
--- a/WinUIGallery/Helpers/TitleBarHelper.cs
+++ b/WinUIGallery/Helpers/TitleBarHelper.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.UI;
 using Microsoft.UI.Xaml;
+using Windows.UI;
 
 namespace WinUIGallery.Helpers;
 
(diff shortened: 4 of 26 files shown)


Zakariathr22:
@niels9001 
I found that many other spacing adjustments are needed on several pages, so I’ll try to address all of them in this PR

image
image

niels9001:
> @niels9001 I found that many other spacing adjustments are needed on several pages, so I’ll try to address all of them in this PR
> 
> image image

Great catch! Thanks!!

WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
```diff
@@ -51,8 +51,8 @@
             <SelectorBarItem Text="Signal" />
             <SelectorBarItem Text="High Contrast" />
         </SelectorBar>
-        <ScrollViewer Grid.Row="3">
-            <Frame x:Name="NavigationFrame" VerticalAlignment="Stretch" />
+        <ScrollViewer Grid.Row="3" Margin="0,0,-12,-36">
```
Jay-o-Way, line 54: In general, a ScrollViewer should _never_ have a margin or a padding. It generates bad UI. The screenshot also looks a bit odd: a gutter between the scrollbar and the window border. The problem lies here:

https://github.com/microsoft/WinUI-Gallery/blob/297974e8d766593749504b78cb642c58e4de578c/WinUIGallery/Pages/ItemPage.xaml#L48

It disregards the fact that a page may have a scroll viewer, which must be able to stretch to the sides. This property should be removed here and every page must deal with their content and layout.

CC @niels9001

Jay-o-Way (changes requested):
see comment

WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
Zakariathr22: This was for the wrong spacing.
image
 The page has a margin, and the inside section also has a margin. I thought the easiest fix was to adjust it this way. Do you have any other suggestions?

Jay-o-Way: In lager scope, the **ScrollViewer** should be moved out of _ItemPage.xaml_ to the content pages of the \<frame>. I know this will be a lot of work, but it really is for the best.

https://github.com/microsoft/WinUI-Gallery/blob/297974e8d766593749504b78cb642c58e4de578c/WinUIGallery/Pages/ItemPage.xaml#L43-L62

P.S. this means `<TextBlock x:Name="descriptionText" ...>` must also be moved.

Jay-o-Way: ...effectively this comment, but applied to every page
https://github.com/microsoft/WinUI-Gallery/blob/297974e8d766593749504b78cb642c58e4de578c/WinUIGallery/Pages/ItemPage.xaml.cs#L77-L78

### For example

General styles:

```xaml
    <Thickness x:Key="NavigationViewPageContentMargin">24</Thickness>
    <Style x:Key="PageBaseStackPanelStyle" TargetType="StackPanel">
        <Setter Property="Spacing" Value="{StaticResource SettingsCardSpacing}" />
        <Setter Property="MaxWidth" Value="1000" />
    </Style>
```

And on every Page:
```xaml
  <Page>
    <ScrollViewer>
        <Grid Margin="{StaticResource NavigationViewPageContentMargin}" ...>
            <StackPanel Style="{StaticResource PageBaseStackPanelStyle}">
            (content, starting with "descriptionText")
```

But you'll have to adapt for pages like **Iconography**: the areas that need to scroll, already deal with it as-is.

niels9001:
Some differences in the vertical spacing between the page title and the content when resizing the window

image

image

Zakariathr22:
@Jay-o-Way 
Feel free to commit any changes you find necessary, you still have access :)

Jay-o-Way:
Oh I already forgot about that, haha. I see now that my comments regard a very large issue that needs input from the more-core dev people. I created a discussion for it: #2033

WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
Jay-o-Way: > Do you have any other suggestions?

#2033

niels9001:
Thinking about this a bit more; couldn't we add an additional helper that one could add to the page (something like `MarginsAreHandled` and we check this on `ItemPage`. If it's set to true, we do not invoke the `NarrowLayout` visual states.

The page itself would then need to handle it. Although not ideal, it might be easier because there's only 1 page that runs into this (the Color page) - and it's a bit of an outlier.

niels9001:
> Some differences in the vertical spacing between the page title and the content when resizing the window
> 
> image image

@Zakariathr22 I believe the fix for this is not in yet, right? We'd like to push out a new build sometime next week so if we can get this PR in, that'd be terrific :)!

WinUIGallery/Samples/ControlPages/Design/ColorPage.xaml
Zakariathr22: Marking this as resolved until we reach a final decision on #2033, so we can handle it in a separate PR.

Zakariathr22:
> Thinking about this a bit more; couldn't we add an additional helper that one could add to the page (something like `MarginsAreHandled` and we check this on `ItemPage`. If it's set to true, we do not invoke the `NarrowLayout` visual states.
> 
> The page itself would then need to handle it. Although not ideal, it might be easier because there's only 1 page that runs into this (the Color page) - and it's a bit of an outlier.

https://github.com/microsoft/WinUI-Gallery/pull/2027#discussion_r2325249570

Zakariathr22:
> @Zakariathr22 I believe the fix for this is not in yet, right? We'd like to push out a new build sometime next week so if we can get this PR in, that'd be terrific :)!

@niels9001
Again, sorry for the delay 🙏,
I think everything looks good now

niels9001 (approved):

niels9001:
> > @Zakariathr22 I believe the fix for this is not in yet, right? We'd like to push out a new build sometime next week so if we can get this PR in, that'd be terrific :)!
> 
> @niels9001 Again, sorry for the delay 🙏, I think everything looks good now

No worries! Great work!

PR #2039, opened 2025-09-10
Replace "Controls" header with a separator in app navigation

AndrewKeepCoding:
## Description

This PR replaces the "Controls" header with a separator in app navigation.

## Motivation and Context

Currently, it's not just "controls" under the header.

## How Has This Been Tested?

Ran the app and confirmed the change.

Current:

image

New:

image

## Types of changes

### DIFF
--- a/WinUIGallery/MainWindow.xaml
+++ b/WinUIGallery/MainWindow.xaml
@@ -272,7 +272,7 @@
                             Tag="AccessibilityColorContrast" />
                     </NavigationViewItem.MenuItems>
                 </NavigationViewItem>
-                <NavigationViewItemHeader Content="Controls" />
+                <NavigationViewItemSeparator />
                 <NavigationViewItem x:Name="AllControlsItem" Content="All">
                     <NavigationViewItem.Icon>
                         <FontIcon Glyph="&#xE8A9;" />


niels9001:
Should we rename "Controls" to "Samples"?

AndrewKeepCoding:
The items above (*Fundamentals*, *Design*, *Accessibility**) also includes "samples". That's why I ended up with a separator. But let me know if you prefer "Samples".

niels9001:
> The items above (_Fundamentals_, _Design_, _Accessibility_*) also includes "samples". That's why I ended up with a separator. But let me know if you prefer "Samples".

Yeah, true.. my worry with removing the header completely is that the distinction between the bigger topics (design, a11y, etc.) and the actual control samples is lost

AndrewKeepCoding:
You are right. What about "Components"?

image

Jay-o-Way:
> Currently, it's not just "controls" under the header.

what are the exceptions?

niels9001:
> You are right. What about "Components"?
> 
> image

I'm not sure if that just starts to complicate things 🤔.. most of the samples there are about controls, and even though it's technically not _all_, it pretty much sums it up though IMO. not sure if there's a better title we can come up with, else I'd just leave it as is?

Zakariathr22:
Can we simply call them samples?

Jay-o-Way:
@Zakariathr22 , @AndrewKeepCoding already answered that.

Zakariathr22:
> @Zakariathr22 , @AndrewKeepCoding already answered that.

It may be philosophical 😂
- Taking only a collection of samples (not all of them) doesn’t change the fact that they are still samples.
- And not explicitly calling them samples (like fundamentals, design guidelines, and accessibility) doesn’t mean they aren’t samples.

So let’s just call them what they are.

Jay-o-Way:
@Zakariathr22 saying "all bachelors are unmarried" is true, but that's because that is the definition. Explicitly calling it out doesn't give it more value than it already has.

The reason that this header is there, is because this app originally was called "XAML Controls Gallery". Later, larger topics were added (as Niels pointed out) shifting the scope of the app. Since **by far** most of the menu items here _are_ for controls, I definitely see the value of this header. If anything, if there are samples that are not controls, maybe those need to be moved from out under this header.

AndrewKeepCoding:
I initially aimed for a quick fix since the "Controls" header no longer accurately reflects its contents. I appreciate all your perspectives. I'm going to close this PR and open a separate issue so we can properly discuss whether the current menu structure should be revised or left as-is. 🙂

PR #2040, opened 2025-09-10
Show Microsoft.UI.Colors items

AndrewKeepCoding:
## Description

This PR will add a `GridView` of `Microsoft.UI.Colors` on the Design/Color page.

## Motivation and Context

Makes it easier to check the color of `Microsoft.UI.Colors` items.

## How Has This Been Tested?

Ran the app and confirmed the changes.

## Types of changes


AndrewKeepCoding:
I can go further and use the code from the **Iconography** page with filtering and clipboard buttons if it's required.🙂

marcelwgn:
Thank you for this contribution, however we do not think that this is information we should include in the Gallery right now. While app developers can use them, they do not support theming; developers should use the theme colors provided by the Windows App SDK. 
Again thank you for the work. In the future, it would be best to first create an issue so we can discuss this first and ensure no efforts are wasted!

AndrewKeepCoding:
Yeah. I should've created an issue first. 😅

PR #2051, opened 2025-09-17
Fix app title bar gap beneath the caption buttons

AndrewKeepCoding:
## Description

This PR fixes the app title bar gap beneath the caption buttons.

## Motivation and Context

Provides a cleaner, more professional look.

## How Has This Been Tested?

Ran the app and confirmed the change under different DPI, resolutions, and switches between them on Win 11.

**Current:**

image

**New:**

image

## Types of changes


WinUIGallery/MainWindow.xaml
```diff
@@ -13,6 +13,9 @@
         <MicaBackdrop />
     </Window.SystemBackdrop>
     <Grid x:Name="RootGrid" Loaded="RootGrid_Loaded">
+        <Grid.Resources>
+            <x:Double x:Key="TitleBarExpandedHeight">46</x:Double>
```
Saharsh979: Why was 46 chosen as the value for TitleBarExpandedHeight?
Is this value derived via observation? If yes, then a hardcoded workaround might break under different conditions (e.g., DPI scaling, theme changes, or different Windows versions, resolution settings)

AndrewKeepCoding: Hi @Saharsh979! Thanks for pointing that out. I should’ve mentioned that I tested the fix under different DPI, resolutions, and while switching between them on Win 11.
That said, the current title bar already relies on some hardcoded width/height values, so I followed the same approach here. I couldn’t find a more flexible way to address the gap, but I’m definitely open to better alternatives if you have suggestions.

Zakariathr22 (changes requested):
When the window is maximized, there will be a 1px overlap between the title bar buttons and the NavigationView.
Image
I don’t recommend modifying the `TitleBarExpandedHeight` resource. Instead, adjust the NavigationView’s padding between `"0, -1, 0, 0"` and `"0, -2, 0, 0"` depending on whether the window is maximized or restored.

AndrewKeepCoding:
@Zakariathr22 Great catch! Thanks! I followed your suggestion (I guess you meant `Margin` instead of `Padding`).

Zakariathr22:
> I guess you meant `Margin` instead of `Padding`

Exactly!

WinUIGallery/MainWindow.xaml.cs
```diff
@@ -32,12 +32,35 @@ public NavigationView NavigationView
...
         dispatcherQueue = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
+
+        WindowPresenter = AppWindow.Presenter as OverlappedPresenter;
+        CurrentWindowState = WindowPresenter.State;
+        AdjustNavigationViewMargin(force: true);
+        AppWindow.Changed += (_, _) => AdjustNavigationViewMargin();
+    }
+
+    // Adjusts the NavigationView margin based on the window state to fix the gap between the caption buttons and the NavigationView.
+    private void AdjustNavigationViewMargin(bool? force = null)
```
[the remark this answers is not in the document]
Zakariathr22: > Why do we need the `force` parameter here?

Oh, don't worry I get it now

WinUIGallery/MainWindow.xaml.cs
```diff
@@ -32,12 +32,35 @@ public NavigationView NavigationView
...
+
     public MainWindow()
     {
         this.InitializeComponent();
         SetWindowProperties();
         RootGrid.ActualThemeChanged += (_, _) => TitleBarHelper.ApplySystemThemeToCaptionButtons(this, RootGrid.ActualTheme);
         dispatcherQueue = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
+
+        WindowPresenter = AppWindow.Presenter as OverlappedPresenter;
+        CurrentWindowState = WindowPresenter.State;
```
Saharsh979: What happens if the cast fails? and WindowPresenter is null?

AndrewKeepCoding: A failed cast here would crash the app. I didn’t add a null check since it should never fail unless WinUI changes fundamentally.
For simplicity, I left it as is. Do you think we should add one just in case?

Saharsh979: Doesn't hurt us to be extra cautious, please do add.

AndrewKeepCoding: Added a null check. Please review.🙂

Zakariathr22 (commented):
@AndrewKeepCoding 
I recently found out that this problem comes from the default title bar itself, not from the XAML components. So I suggest marking it as a WinAppSDK issue (https://github.com/microsoft/WindowsAppSDK/issues/5947) and not using this workaround. The issue doesn’t break the app, and we would have to remove the workaround later when it’s fixed anyway, because it would then cause another problem with the spacing between the title bar and the NavigationView.

Image

Image

AndrewKeepCoding:
@Zakariathr22 Yeah, but we don't know "when/if" this will be addressed, so I think we should adjust it for cleaner look.

That said, it's only noticeable when hovering over the caption buttons, so maybe it's not worth the extra effort. 😅
Still, IMHO, this is one of the small details that I'd personally fix. 🙂

WinUIGallery/MainWindow.xaml.cs
```diff
@@ -32,12 +32,42 @@ public NavigationView NavigationView
...
+    private OverlappedPresenterState CurrentWindowState { get; set; }
+
     public MainWindow()
     {
         this.InitializeComponent();
         SetWindowProperties();
         RootGrid.ActualThemeChanged += (_, _) => TitleBarHelper.ApplySystemThemeToCaptionButtons(this, RootGrid.ActualTheme);
         dispatcherQueue = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
+
+        if (AppWindow.Presenter is OverlappedPresenter windowPresenter)
```
marcelwgn, line 51: Can you add comments indicating that this section and the function below are just a workaround for the underlying WinUI issue?

AndrewKeepCoding: Done.🙂

PR #2056, opened 2025-09-22
Fix SplitView sample layout issue when pane is on the right side

### LINKED ISSUE #322: SplitView sample needs work for right-alignment
- [ ] New sample

SplitView

**Describe the sample**
Change placement to Right:
image

It's a bit weird that the content isn't stretched to put the panel on the right-side next to the options.

It get's odder when you change the display mode to `CompactInline` or `CompactOverlay` as the icons don't show up:

image

(Also changing the DisplayMode to `CompactInline` doesn't appear to work the first time, as `IsPaneOpen` is still on, looks like changing to the overlay modes turns it off as well at the same time, so it's inconsistent here.)

AndrewKeepCoding:
## Description

This PR updates the `SplitView` sample to switch layouts dynamically depending on the pane's placement (left or right).

## Motivation and Context

Fixes #322.

## How Has This Been Tested?

Manually tested.

## Types of changes


AndrewKeepCoding:
@marcelwgn Do you think we should horizontally stretch the `SplitView` to the options panel (as suggested in the issue)? I feel it looks too wide for this sample.

marcelwgn:
> @marcelwgn Do you think we should horizontally stretch the `SplitView` to the options panel (as suggested in the issue)? I feel it looks too wide for this sample.

I agree that it would look if we stretch it to the options panel. I think we should have a fixed maximum width for it.

WinUIGallery/Samples/ControlPages/SplitViewPage.xaml.cs
```diff
@@ -45,6 +51,27 @@ private void PanePlacement_Toggled(object sender, RoutedEventArgs e)
         {
             splitView.PanePlacement = SplitViewPanePlacement.Left;
         }
+
+        UpdateNavLinkItemLayout();
+    }
+
+    private void UpdateNavLinkItemLayout()
+    {
+        //var displayMode = (SplitViewDisplayMode)Enum.Parse(typeof(SplitViewDisplayMode), (displayModeCombobox.SelectedItem as ComboBoxItem).Content.ToString());
```
marcelwgn: nit: Remove

WinUIGallery/Samples/ControlPages/SplitViewPage.xaml.cs
```diff
@@ -58,6 +85,16 @@ private void paneBackgroundCombobox_SelectionChanged(object sender, SelectionCha
...
+
+    private void openPaneLengthSlider_ValueChanged(object sender, Microsoft.UI.Xaml.Controls.Primitives.RangeBaseValueChangedEventArgs e)
+    {
+        UpdateNavLinkItemLayout();
+    }
+
+    private void compactPaneLengthSlider_ValueChanged(object sender, Microsoft.UI.Xaml.Controls.Primitives.RangeBaseValueChangedEventArgs e)
+    {
+        UpdateNavLinkItemLayout();
+    }
```
marcelwgn: Are these two required? I think just switching layout when we switch left/right should be enough.

marcelwgn (commented):
Thank you for fixing this! Changes look good, just a little bit of cleanup is necessary.

WinUIGallery/Samples/ControlPages/SplitViewPage.xaml.cs
AndrewKeepCoding: Sorry. Removed unnecessary code.

marcelwgn (approved):

PR #2063, opened 2025-10-03
Replace MainWindow with plain Window in sample projects

### LINKED ISSUE #2062: Child window sample forces parent window navigation
### Which version of the app?

WinUI 3 Gallery

### Description

Parent window navigates back to Home when creating a child window.

1. Go to the **Multiple windows** sample page.
2. Click the **Create new Window** button.
3. Both parent and child window navigates to **Home**.

### Screenshots

_No response_

### Windows version

_No response_

### Additional context

We shouldn't use `MainWindow` for the child window sample anymore.

```cs
protected override void OnNavigatedTo(NavigationEventArgs e)
{
    ((NavigationViewItem)App.MainWindow.NavigationView.MenuItems.First()).IsSelected = true;
    ...
```

AndrewKeepCoding:
## Description

This PR updates `WindowHelper.CreateWindow()` to return a plain `Window` instead of `MainWindow`, which was tightly coupled to several static members.

## Motivation and Context

Fixes #2062.

## How Has This Been Tested?

Manually tested.

image

## Types of changes

### DIFF
--- a/WinUIGallery/Samples/ControlPages/CreateMultipleWindowsPage.xaml.cs
+++ b/WinUIGallery/Samples/ControlPages/CreateMultipleWindowsPage.xaml.cs
@@ -3,8 +3,9 @@
 
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+using Windows.Graphics;
 using WinUIGallery.Helpers;
-using WinUIGallery.Pages;
 
 namespace WinUIGallery.ControlPages;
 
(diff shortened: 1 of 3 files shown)


WinUIGallery/Samples/ControlPages/CreateMultipleWindowsPage.xaml.cs
```diff
@@ -17,12 +18,29 @@ public CreateMultipleWindowsPage()
 
     private void createNewWindow_Click(object sender, RoutedEventArgs e)
     {
-        var newWindow = new MainWindow();
+        var newWindow = new Window
```
Zakariathr22: Reading this was painful 😅, can't we just create a `NewWindow` or `SecondaryWindow`?

AndrewKeepCoding: My intension was to avoid adding custom code to the sample as much as possible, but sure I can do that.🙂

marcelwgn: I agree with the both of you, its difficult to read but for the sake of the sample we should not rely on helpers. Do we need to need to have ThemeHelper provide the theme?
Also can we call the constructor instead of omitting the "()"? 😄

WinUIGallery/Samples/SampleCode/Window/CreateWindowSample1.txt
```diff
@@ -1,11 +1,24 @@
...
-// C# code to navigate in the new window
-var targetPageType = typeof(HomePage);
-string targetPageArguments = string.Empty;
-rootPage.Navigate(targetPageType, targetPageArguments);
\ No newline at end of file
+childWindow.AppWindow.ResizeClient(new SizeInt32(500, 500));
+
+// The WindowHelper is a helper class in the WinUIGallery project
+// that helps us close child windows when the main window closes.
+WindowHelper.TrackWindow(childWindow);
```
marcelwgn: Do we need to include this?

marcelwgn (changes requested):

WinUIGallery/Samples/ControlPages/CreateMultipleWindowsPage.xaml.cs
AndrewKeepCoding: >  for the sake of the sample we should not rely on helpers

I completely agree.

> Do we need to need to have ThemeHelper provide the theme?

~If we want the new window use the app theme, which might differ from the system theme, we should set it here using `ThemeHelper`. We can remove it from the sample code (CreateWindowSample1.txt).~
We can get the theme from the parent.

> Also can we call the constructor instead of omitting the "()"? 😄

Yes, we [can](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers?WT.mc_id=%3Fwt.mc_id%3DMVP_356303#object-initializers), but I'll add them back since I'm not sure is preferred or not. 🙂

WinUIGallery/Samples/SampleCode/Window/CreateWindowSample1.txt
AndrewKeepCoding: We can remove the helpers from the sample code text file.

PR #2076, opened 2025-11-09
Fix CornerRadius Mismatch Between GridViewItem and Its ItemTemplate

ghost1372:
ItemTemplates.xaml uses an 8px corner radius, while the GridViewItem itself uses 4px. This difference is visible when hovering over the item.

## Types of changes

Before:
image

After:
image

### DIFF
--- a/WinUIGallery/Pages/HomePage.xaml
+++ b/WinUIGallery/Pages/HomePage.xaml
@@ -20,6 +20,8 @@
     HighContrastAdjustment="None">
     <Grid>
         <Grid.Resources>
+            <CornerRadius x:Key="GridViewItemCornerRadius">8</CornerRadius>
+
             <animations:ImplicitAnimationSet x:Name="ShowTransitions">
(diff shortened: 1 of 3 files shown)


Zakariathr22 (changes requested):
Hello @ghost1372, this is actually a really good idea! We should consider applying it to items in:
- Recently visited
- Favorites
- SectionPage
- SearchResultsPage

ghost1372:
Thank you @Zakariathr22 
Good catch! I missed applying it earlier, it’s all fixed now.

Jay-o-Way:
Why not use the CornerRadius property directly? With Static resource for Overlay.

ghost1372:
> Why not use the CornerRadius property directly? With Static resource for Overlay.

Since it's been 2 weeks, if I remember correctly I tried to set CornerRadius in GridViewItem, but it didn't work and I used lightweight styling.

niels9001 (approved):

PR #2100, opened 2026-01-29
Enable Built-in COM interop for ContentIsland sample

### LINKED ISSUE #2074: Built-in COM disabled via feature switch when loading `ContentIsland` sample
### Which version of the app?

WinUI 3 Gallery

### Description

When running the `ContentIsland` sample and clicking **Load model**, the app throws a `System.NotSupportedException `with the message **Built-in COM has been disabled via a feature switch**.

### Screenshots

Image

### Windows version

_No response_

### Additional context

_No response_

Zakariathr22:
## Description
Explicitly enables built-in COM interop in the project by setting <BuiltInComInteropSupport>true</BuiltInComInteropSupport> in the .csproj file.
This ensures that the `ContentIsland` sample can load models without throwing a `System.NotSupportedException` when trimming is enabled.

## Motivation and Context
- Fixes #2074 

## How Has This Been Tested?
Tested by running the `ContentIsland` sample locally in both `Release` and `Debug` mode. 

None

## Types of changes

### DIFF
--- a/WinUIGallery/Controls/SampleCodePresenter.xaml
+++ b/WinUIGallery/Controls/SampleCodePresenter.xaml
@@ -26,6 +26,7 @@
             <ContentPresenter
                 x:Name="CodePresenter"
                 Padding="{x:Bind Padding, Mode=OneWay}"
+                MinHeight="30"
                 Loaded="CodePresenter_Loaded" />
         </ScrollViewer>
         <Border
--- a/WinUIGallery/Samples/ControlPages/BadgeNotificationManagerPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/BadgeNotificationManagerPage.xaml
@@ -35,7 +35,7 @@
                 <x:String xml:space="preserve">
 private void SetBadgeCountButton_Click(object sender, RoutedEventArgs e)
 {
-    BadgeNotificationManager.Current.SetBadgeAsCount((uint)BadgeCountBox.Value);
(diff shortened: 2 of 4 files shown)


Zakariathr22 (commented):
> [!IMPORTANT]
> I’m not sure if this will impact the NAOT adoption in PR #2095, so it may need verification to ensure there are no runtime issues.

ghost1372:
@niels9001  I should mention that this will increase the size of  the program, even in AOT mode it will add 10 ~ 30 MB.
So if the size is not important, it is not important, but if the size is to be mentioned in the shows, I think it is better to think about it, such as removing this sample from the gallery! (Consider a situation where, for example, NAoT reduces the application size by about 30 MB, and then you want to highlight this improvement in conferences or Shows. However, the built-in COM cause the same 30 MB to be added back, so in the end there is no difference in the application size!)

this is my DevWinUI gallery app with and without Built-in COM interop support:

Enabled:
image


Disabled:
image

Zakariathr22:
> @niels9001 I should mention that this will increase the size of the program, even in AOT mode it will add 10 ~ 20 MB. So if the size is not important, it is not important, but if the size is to be mentioned in the shows, I think it is better to think about it, such as removing this sample from the gallery!

@marcelwgn I’m very interested in knowing your thoughts on this.

marcelwgn:
That size increase is quite bad, a non working sample is worse though in my opinion. I think for now, we should enable it since the only other option really is only removing the samples requiring such interactivity. The latter is something we might revisit in the future.

PR #2102, opened 2026-01-30
Remove duplicated `RadioButtons` page

### LINKED ISSUE #2068: Duplicate RadioButtons sample in Basic Input and Layout sections
The RadioButtons sample appears in both the Basic Input and Layout sections.
The RadioButtons sample should appear only once, under Basic Input.
It should be removed from the Layout section.

Zakariathr22:
## Description
Main changes are;
- Moved the example from the RadioButtons page to the RadioButton page
- Removed the redundant RadioButtons page
- Updated ControlInfoData.json accordingly
- Removed RadioButtons.png from assets

## Motivation and Context
- closes #2068 

## How Has This Been Tested?
Manually tested

## Types of changes

### DIFF
--- a/WinUIGallery/Pages/ItemPage.xaml
+++ b/WinUIGallery/Pages/ItemPage.xaml
@@ -53,6 +53,8 @@
                 <TextBlock
                     x:Name="descriptionText"
                     Grid.ColumnSpan="2"
+                    MaxWidth="1064"
+                    HorizontalAlignment="Left"
                     Margin="0,4,24,0"
                     Style="{ThemeResource BodyTextBlockStyle}"
                     Text="{x:Bind Item.Description}"


WinUIGallery/Samples/Data/ControlInfoData.json
Jay-o-Way:
```suggestion
          "Description": "Use RadioButton controls to let a user choose between mutually exclusive, related options. Generally contained within a RadioButtons group control.",
```
To differentiate between the two very different items

Jay-o-Way:
Did we lose this link?
image

Zakariathr22:
> I think we should converge into the RadioButtons page instead of RadioButton. There are very little use cases for the RadioButton control (without s), you would probably almost always use a RadioButtons control. Before RadioButtons, RadioButton was the only option and it did some grouping via key logic and walking the tree (iirc), but now thats no longer necessary. RadioButtons is the easier and better option for most scenarios.

@marcelwgn, should it be in the controls section or the layouts section?

marcelwgn:
> > I think we should converge into the RadioButtons page instead of RadioButton. There are very little use cases for the RadioButton control (without s), you would probably almost always use a RadioButtons control. Before RadioButtons, RadioButton was the only option and it did some grouping via key logic and walking the tree (iirc), but now thats no longer necessary. RadioButtons is the easier and better option for most scenarios.
> 
> @marcelwgn, should it be in the controls section or the layouts section?

We should put it into controls since it is a control for user input, I would say layouts are more "how do organize UI elements" which RadioButtons is not really for.

marcelwgn:
After internal discussion, I think its fine to stick with RadioButton instead of RadioButtons.

marcelwgn (approved):

Zakariathr22:
> After internal discussion, I think its fine to stick with RadioButton instead of RadioButtons.

I made the necessary changes but I forgot to commit them 😅

PR #2106, opened 2026-02-05
Add window state save/restore feature

### LINKED ISSUE #1606: Resize the window on startup
### Issue type

enhancement

### Which version of the app?

WinUI 3 Gallery

### Description

On small monitors/laptops etc, the launch window default size is probably acceptable. On larger monitors (ultra wide for example) it is massive, 3824 x 1002.

Request that a max size is set 1366 x 768, or that it remembers the size it was last set to (like UWP did).

### Screenshots

_No response_

### Windows version

Windows 11 23H2 (22631)

### Additional context

_No response_

Zakariathr22:
## Description
- Adds support for saving and restoring main window size, position, scale, and maximized state between sessions.
- Introduces new settings properties and UI toggle for enabling/disabling this feature.
- Implements `WindowStateHelper` and `WindowState` classes for state management.
- Window state is restored on load and saved on close (only if enabled in `Settings` page).

## Motivation and Context
- Closes #1606 

## How Has This Been Tested?
Manually tested

## Types of changes


niels9001:
@zakariathr22 I haven't tested this yet, will do soon! How will this work when moving between monitors? E.g. opening the app on multi-monitor setup, and then unplugging and only using the laptop's screen?

Zakariathr22:
> @Zakariathr22 I haven't tested this yet, will do soon! How will this work when moving between monitors? E.g. opening the app on multi-monitor setup, and then unplugging and only using the laptop's screen?

I tested it, and the window consistently reopens on the main screen after being closed and launched again.

How the code works:

* Retrieves the `AppWindow` instance.
* Uses `XamlRoot.RasterizationScale` to account for DPI scaling, ensuring the window maintains the same physical size across different monitors.
* Before saving the window state, it temporarily restores the window if it is maximized. This ensures the actual window size is saved rather than the maximized dimensions.
* When restoring the window state, it validates the saved position against the display area bounds to keep the window fully visible and prevent it from appearing partially off-screen or on a disconnected monitor.

WinUIGallery/Helpers/WindowStateHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            // adjust for DPI/scaling change
+            double scaleFactor = currentScale / state.Scale;
+            int width = (int)(state.Width * scaleFactor);
+            int height = (int)(state.Height * scaleFactor);
+            int x = state.PositionX;
+            int y = state.PositionY;
+
+            // ensure window fits display area
+            var displayArea = DisplayArea.GetFromWindowId(window.AppWindow.Id, DisplayAreaFallback.Primary).WorkArea;
```
marcelwgn: That is super suspicious, why do we query the initial display area window places our window which is just the default screen? I tried out this change and if I move the window to my secondary monitor, this line will always move it back to the main display.

WinUIGallery/Pages/SettingsPage.xaml
```diff
@@ -103,6 +103,15 @@
                         </StackPanel>
                     </toolkit:SettingsCard>
 
+                    <toolkit:SettingsCard Header="Save window state"
```
marcelwgn, line 106: This shouldnt be an option, either we support it and it works or we dont support it.

marcelwgn (changes requested):

WinUIGallery/Helpers/WindowStateHelper.cs
Zakariathr22: I’m not sure yet how to make it function like the default Windows applications. Let me take some time to think it through and research the best approach.

marcelwgn: I think easiest is to use DisplayAreas.FindAll (or whatever its called) and just iterate through them to figure out if we can move to our original target.

Zakariathr22: Resolved in https://github.com/microsoft/WinUI-Gallery/pull/2106/changes/c2996aa22ca34f1ce81a248bee4a4088ed087b35 

The solution was much easier than I expected 😆. We retrieve the display area from the last saved position using `DisplayArea.GetFromPoint` method and use the primary display area as a fallback.

WinUIGallery/Pages/SettingsPage.xaml
Zakariathr22: Can we negotiate this?

Sometimes, when I use applications that have this functionality, I actually wish it wasn’t enabled. I may resize the window for a specific task, but that doesn’t mean I’ll need it at that size again later.

What if we enable the feature by default (make it `true`), but also keep providing users with the option to disable it if they prefer not to use it?

ghost1372:
@Zakariathr22 We can use WindowsAppSdk Built-in API for this. Spec is ready and is working fine.
https://github.com/microsoft/WindowsAppSDK/blob/afd4ac42a32a329f4fdfc76d7d443a0200774135/specs/Windowing/AppWindowPlacement.md

@marcelwgn  Can you ask your internal friends whether this API will be included in the stable v2 release or not? It is currently available in experimental 5.

WinUIGallery/Pages/SettingsPage.xaml
niels9001: Do we have use cases why it should (not) be optional? E.g. when I look at most (inbox) Windows apps they have logic to do this, without the option to turn it off?

Zakariathr22:
> @Zakariathr22 We can use WindowsAppSdk Built-in API for this. Spec is ready and is working fine. https://github.com/microsoft/WindowsAppSDK/blob/afd4ac42a32a329f4fdfc76d7d443a0200774135/specs/Windowing/AppWindowPlacement.md
> 
> @marcelwgn Can you ask your internal friends whether this API will be included in the stable v2 release or not? It is currently available in experimental 5.

We can proceed with closing this PR once we have confirmation that the API will be part of the upcoming release.

niels9001:
@zakariathr22 @ghost1372 I'm following up internally to understand what to expect from these APIs. Will circle back here

ghost1372:
@niels9001 any news?

PR #2115, opened 2026-02-14
Switch icons tags layout from `WrapLayout` to `FlowLayout`

Zakariathr22:
## Description
Replaced `WrapLayout` with `FlowLayout` for tag display in `IconographyPage`, updating both XAML and code-behind to use `LineSpacing` and `MinItemSpacing` instead of `HorizontalSpacing` and `VerticalSpacing`.

This change may improve the visual arrangement and spacing of tag items.


Zakariathr22:
> Tested, works well.. thanks! Nit: we can remove the `CommunityToolkit.WinUI.Controls` namespaces from both the `.xaml` and `.xaml.cs` pages?

Done!

PR #2118, opened 2026-02-16
Fix keyboard accessibility for SplitButton color picker flyouts (MAS 2.1.1)

pranav-gupta-msft:
## Description
Fixed keyboard accessibility for SplitButton color picker flyouts on the SplitButton sample page. Both SplitButton controls ("Font color" and "Choose color") now open their color picker flyouts when activated via keyboard (Enter/Space). Previously, keyboard and assistive technology users could not expand the flyouts to access the color options.

### Changes:
- Added `Click="myColorButtonReveal_Click"` to the "Choose color" SplitButton in XAML
- Added `myColorButtonReveal_Click` handler that programmatically opens the flyout via `sender.Flyout.ShowAt(sender)`
- Updated `myColorButton_Click` to also open the flyout after applying the selected color
- Updated `SplitButtonSample2.txt` sample code to include the `Click` handler for consistency

## Motivation and Context
Keyboard and assistive technology users were unable to expand the color picker flyouts on the SplitButton page. When navigating to either SplitButton and pressing Enter/Space, nothing happened — the flyout would not open, preventing users from selecting a color. This is a violation of **MAS 2.1.1 – Keyboard** (Trap ID: 2.1 - Physical Challenge), which requires all functionality to be operable via keyboard.

## How Has This Been Tested?
- Navigated to the SplitButton page using keyboard only
- Verified that pressing Enter/Space on both SplitButtons opens the color picker flyout
- Verified that color selection still works correctly via both mouse and keyboard
- Confirmed no regressions in existing flyout dismiss behavior

image

## Types of changes

### DIFF
--- a/WinUIGallery/Samples/ControlPages/SplitButtonPage.xaml
+++ b/WinUIGallery/Samples/ControlPages/SplitButtonPage.xaml
@@ -84,7 +84,8 @@
                     MinHeight="0"
                     Padding="5"
                     VerticalAlignment="Top"
-                    AutomationProperties.Name="Font color with text">
+                    AutomationProperties.Name="Font color with text"
+                    Click="myColorButtonReveal_Click">
                     Choose color
                     <SplitButton.Flyout>
                         <Flyout Placement="Bottom">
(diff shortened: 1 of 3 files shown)


dipeshmsft:
I think this behavior should be fixed in the control, rather than in the sample here. I will look through the issue more comprehensively and get back. But the fix looks good if the requirement is to fix it here

pranav-gupta-msft:
Agreed. Closing the PR, since, it is not a correct solution.

PR #2132, opened 2026-02-27
[Clipboard] Porting UWP clipboard samples to WinUI Gallery

niels9001:
Source: https://github.com/microsoft/Windows-universal-samples/tree/main/Samples/Clipboard/cs

### DIFF
--- a/WinUIGallery/Samples/ControlPages/ClipboardPage.xaml.cs
+++ b/WinUIGallery/Samples/ControlPages/ClipboardPage.xaml.cs
@@ -5,10 +5,11 @@ using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media.Imaging;
 using System;
+using System.Collections.Generic;
 using System.Text;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.DataTransfer;
-using Windows.Storage.Pickers;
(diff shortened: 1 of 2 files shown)


WinUIGallery/Samples/Data/ControlInfoData.json
```diff
@@ -3231,6 +3231,69 @@
             }
           ]
         },
+        {
+          "UniqueId": "Clipboard",
```
marcelwgn: Dont we already have a clipborad page? I think this causes the app crash on startup

WinUIGallery/Samples/ControlPages/ClipboardPage.xaml.cs
```diff
@@ -3,9 +3,13 @@
 
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media.Imaging;
 using System;
+using System.Text;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.DataTransfer;
+using Windows.Storage.Pickers;
```
Zakariathr22: Can't we use `Microsoft.Windows.Storage.Pickers`

marcelwgn: Good point, switching to these APIs now. Thanks for pointing that out!

PR #2140, opened 2026-03-10
Fix the remaining warnings

marcelwgn:
Fixing the remaining issues for the WinUIGaller.slnx


WinUIGallery/Samples/ControlPages/AppWindowTitleBarPage.xaml
```diff
@@ -179,7 +179,7 @@ public sealed partial class AppWindowTitleBarWindow : Window
                         Width="200"
                         Header="TitleBarHeightOption"
                         ItemsSource="{x:Bind titleBarHeightOptions}"
-                        SelectedItem="{x:Bind selectedHeight}"
+                        SelectedIndex="0"
```
niels9001, line 182: Why is this change needed?

marcelwgn: The field referenced caused an unused warning and admittedly, its a bit strange to have a field that gets only read once for the default value.

PR #2167, opened 2026-05-14
Restructure samples

marcelwgn:
> This is a draft to garner feedback before we fully convert one example to so AI can convert the rest

We are introducing a new SampleCodePath property on ControlExample that loads both XAML and C# code snippets from a single file using a simple section-based format:
```
 --- xaml
 <Button Content="Click me" />
 --- c#
 private void OnClick() { }
```
Sections are identified by ---  followed by the type (xaml or c#). They can appear in any order, and either may be omitted.

Folder restructuring:

 - Moved ButtonPage + its sample code into Samples/Button/ to have a per-control folder pattern
 - Moved Data, SampleHelpers, and SamplePages from Samples/ into a new SampleSupport/ root folder to separate support files from actual sample pages

Namespaces are unchanged - only physical file locations and MSBuild/runtime path references were updated.

Future direction: The --- type marker format is designed to be extensible. In a future change, metadata like title and description could also be inlined in the same file (e.g. --- title, --- description), allowing a single file to fully describe a control example (except the code needed for setup of course)


WinUIGallery/Controls/ControlExample.xaml.cs
```diff
@@ -127,6 +129,13 @@ public string CSharpSource
         set { SetValue(CSharpSourceProperty, value); }
     }
 
+    public static readonly DependencyProperty SampleCodePathProperty = DependencyProperty.Register("SampleCodePath", typeof(string), typeof(ControlExample), new PropertyMetadata(null, OnSampleCodePathChanged));
```
niels9001: Nit: nameof() instead of string.

niels9001 (approved):

PR #2171, opened 2026-05-17
Remove the wapproj

0x5bfa:
## Description

Remove the WAP project


WinUIGallery.slnx
```diff
@@ -13,9 +13,64 @@
     <Platform Name="x64" />
     <Platform Name="x86" />
   </Configurations>
-  <Folder Name="/Solution Items/">
+  <Folder Name="/global/">
```
marcelwgn: Keep this as Solution Items.

WinUIGallery.slnx
```diff
@@ -13,9 +13,64 @@
...
     <File Path="../Directory.Build.props" />
   </Folder>
+  <Folder Name="/src/">
```
marcelwgn: WinUI Gallery and the source generator should remain top level in my opinion.

marcelwgn (changes requested):
Thanks for doing this but next time please file an issue before doing this, especially if its about highly opinionated things like project (infra)structure.

0x5bfa:
Since I have a few questions, I've opened an issue.

marcelwgn (approved):

PR #2206, opened 2026-07-28
Fix keyboard navigation for ConnectedAnimation ItemsRepeater sample

protikbiswas100:
The 'Connected animation with ItemsRepeater' sample rendered each item as a non-focusable Grid that only handled Tapped, so keyboard users could not focus, navigate between, or invoke the list items (accessibility issue, MAS 2.1.1).

Host each item in a chrome-less Button so items are keyboard-focusable and can be invoked with Enter/Space (and expose an Invoke automation pattern for Narrator), enable XYFocusKeyboardNavigation on the ItemsRepeater for arrow-key navigation, and return focus to the activated item on back-navigation. The displayed sample source (ConnectedAnimationItemsrepeater.txt) is updated to match.

Work item: https://dev.azure.com/microsoft/OS/_workitems/edit/62821652

## How Has This Been Tested?
Build and Launched the solution file

Before

https://github.com/user-attachments/assets/39a85bef-d476-4962-b580-a88d421668d8

After

https://github.com/user-attachments/assets/e5ef1b40-ac7d-4a02-a26f-78d2c290f058

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/WinUIGallery/SampleSupport/SamplePages/ItemsRepeaterCollectionPage.xaml.cs
+++ b/WinUIGallery/SampleSupport/SamplePages/ItemsRepeaterCollectionPage.xaml.cs
@@ -28,8 +28,6 @@ public sealed partial class ItemsRepeaterCollectionPage : Page
 
     private void Repeater_ElementPrepared(ItemsRepeater sender, ItemsRepeaterElementPreparedEventArgs args)
     {
-        // The item's Grid is itself the focusable element (IsTabStop is set in the template), so no
-        // Button wrapper is needed. Wire pointer (Tapped) and keyboard (Enter/Space) activation on it.
         args.Element.Tapped -= Item_Tapped;
         args.Element.Tapped += Item_Tapped;
         args.Element.KeyDown -= Item_KeyDown;
--- a/WinUIGallery/SampleSupport/SamplePages/ItemsRepeaterCollectionPage.xaml.cs
+++ b/WinUIGallery/SampleSupport/SamplePages/ItemsRepeaterCollectionPage.xaml.cs
@@ -3,7 +3,6 @@
 
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
-using Microsoft.UI.Xaml.Input;
 using Microsoft.UI.Xaml.Media;
 using Microsoft.UI.Xaml.Media.Animation;
 using Microsoft.UI.Xaml.Navigation;
@@ -27,12 +26,16 @@ public sealed partial class ItemsRepeaterCollectionPage : Page
 
     private void Repeater_ElementPrepared(ItemsRepeater sender, ItemsRepeaterElementPreparedEventArgs args)
     {
-        // Attach a Tapped handler to each item container so we can navigate on click.
-        args.Element.Tapped -= Item_Tapped;
-        args.Element.Tapped += Item_Tapped;
+        // Each item is hosted in a Button (see the DataTemplate) so it is keyboard-focusable and can be
+        // invoked with Enter/Space as well as the pointer. Wire the Click handler on the realized container.
+        if (args.Element is Button button)
+        {
+            button.Click -= Item_Click;
+            button.Click += Item_Click;
+        }
     }
 
-    private void Item_Tapped(object sender, TappedRoutedEventArgs e)
+    private void Item_Click(object sender, RoutedEventArgs e)
     {
         if (sender is not FrameworkElement element)
         {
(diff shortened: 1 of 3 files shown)


WinUIGallery/Samples/ConnectedAnimation/ConnectedAnimationItemsrepeater.txt
```diff
@@ -17,18 +18,37 @@ Connected animation with ItemsRepeater
...
-                           Stretch="UniformToFill" />
-                    <Border VerticalAlignment="Bottom"
-                            Background="{ThemeResource AcrylicBackgroundFillColorBaseBrush}"
-                            Padding="8,4">
-                        <TextBlock Text="{x:Bind Title}"
-                                   Style="{ThemeResource CaptionTextBlockStyle}" />
-                    </Border>
-                </Grid>
+                <!-- ItemsRepeater has no built-in focus/selection, so host each item in a
+                     Button to make it keyboard-focusable and invokable with Enter/Space.
```
godlytalias, line 22: Whether this means each tab navigates to next item? Ideally navigation through list is supposed to work with arrow keys and tab navigation have to take to the next control I guess. Please confirm on the experience here

protikbiswas100: @godlytalias
The behavior is that I can navigate through arrow key also, not just using tabs

godlytalias: With the new changes, whether tab key moves to next item or navigates out of ItemsRepeater, ideally the experience would be to use only arrow key for list navigations

protikbiswas100: dropped the Button. I was using it only to get a focusable, invokable item, but since IsTabStop and the key events are actually on UIElement, the item Grid can just be the focus target itself (IsTabStop + UseSystemFocusVisuals). Wired Tapped + Enter/Space on it for activation.

On the navigation experience, it now works exactly as you described:
- Tab: enters the repeater once, next Tab moves out to the next control (TabFocusNavigation="Once" on the ScrollViewer), so it does not step through every item.
- Arrow keys: move between items inside the repeater (XYFocusKeyboardNavigation="Enabled").
- Enter/Space: opens the focused item.

Pushed in dcae72f.

WinUIGallery/Samples/ConnectedAnimation/ConnectedAnimationItemsrepeater.txt
```diff
@@ -17,7 +23,14 @@ Connected animation with ItemsRepeater
         </ItemsRepeater.Layout>
         <ItemsRepeater.ItemTemplate>
             <DataTemplate x:DataType="local:CustomDataObject">
-                <Grid CornerRadius="4">
+                <!-- ItemsRepeater has no built-in focus/selection, but IsTabStop and the keyboard
```
godlytalias: Such comments may not be interesting for a future reader, remove them

WinUIGallery/Samples/ConnectedAnimation/ConnectedAnimationItemsrepeater.txt
```diff
@@ -43,14 +56,33 @@ private double _persistedScrollPosition;
 
 private void Repeater_ElementPrepared(ItemsRepeater sender, ItemsRepeaterElementPreparedEventArgs args)
 {
-    // Attach a Tapped handler to each item container.
+    // The item's Grid is itself focusable (IsTabStop in the template), so no Button is needed.
```
godlytalias: Such comments may not be interesting for a future reader, remove them

WinUIGallery/SampleSupport/SamplePages/ItemsRepeaterCollectionPage.xaml.cs
```diff
@@ -27,14 +28,32 @@ public ItemsRepeaterCollectionPage()
 
     private void Repeater_ElementPrepared(ItemsRepeater sender, ItemsRepeaterElementPreparedEventArgs args)
     {
-        // Attach a Tapped handler to each item container so we can navigate on click.
+        // The item's Grid is itself the focusable element (IsTabStop is set in the template), so no
```
godlytalias: Such comments may not be interesting for a future reader, remove them

WinUIGallery/SampleSupport/SamplePages/ItemsRepeaterCollectionPage.xaml.cs
```diff
@@ -68,21 +87,34 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
         scrollViewer.ChangeView(null, _persistedScrollPosition, null, disableAnimation: true);
         UpdateLayout();
 
+        // Find the element for the stored item so we can run the back animation and restore focus to it.
+        int index = repeater.ItemsSourceView.IndexOf(_storedItem);
+        var container = repeater.TryGetElement(index) as FrameworkElement;
```
godlytalias, line 90: Why these logic are required? Whether we are adding some new container into which focus to be given?

protikbiswas100: This block isn't creating a new container or a new focus target - it retrieves the item element that ItemsRepeater has **already realized** for `_storedItem`:

- `ItemsSourceView.IndexOf(_storedItem)` -> the data item's index.
- `repeater.TryGetElement(index)` -> the existing realized `UIElement` for that index (returns null if it isn't realized).

It's required for the **back connected animation**: `ConnectedAnimationService.GetAnimation("BackConnectedAnimation")` needs a concrete target element to animate into, so we locate the `connectedElement` image inside that realized container and call `animation.TryStart(...)`. Unlike ListView/GridView, ItemsRepeater has no `TryStartConnectedAnimationAsync()`, so the element must be looked up manually.

The same already-realized container is then reused to return keyboard focus to the tile the user activated, so keyboard/Narrator users resume where they left off after back-navigation. No new container is added.

WinUIGallery/Samples/ConnectedAnimation/ConnectedAnimationItemsrepeater.txt
protikbiswas100: Done - removed in 2c4d5ca.

WinUIGallery/Samples/ConnectedAnimation/ConnectedAnimationItemsrepeater.txt
protikbiswas100: Done - removed in 2c4d5ca.

WinUIGallery/SampleSupport/SamplePages/ItemsRepeaterCollectionPage.xaml.cs
protikbiswas100: Done - removed in 2c4d5ca.

godlytalias (approved):

PR #2209, opened 2026-08-03
Fix keyboard focus trap in StoragePickers sample after picker closes

HariniMalothu17:
## Summary
On the **Storage pickers** sample page, keyboard focus becomes trapped after a picker dialog closes. Each picker button handler disables the button while the picker is shown and re-enables it afterward:

```csharp
button.IsEnabled = false;      // focused button is disabled -> keyboard focus is dropped
var result = await picker...;  // show the Open/Save/Folder dialog
button.IsEnabled = true;       // re-enabled, but focus is NOT restored
```

Disabling the focused button drops keyboard focus, and re-enabling it does not bring focus back, so a keyboard-only user is left with no focused element and cannot Tab away (**MAS 2.1.2 - No Keyboard Trap**).

## Fix
Restore focus to the button with `Focus(FocusState.Programmatic)` immediately after re-enabling it. This matches the focus-restoration pattern already used elsewhere in the repo (e.g. `ThemeTransitionPage`, `DetailedInfoPage`). No visual or behavioral change other than focus correctly returning to the button.

Applied to all six handlers in `StoragePickersPage.xaml.cs` (Pick single file, Pick multiple files, Save a file, Pick folder, Pick file for thumbnail, Select suggested folder). The five displayed sample code snippets (`.txt`) are updated to match.

## Repro (before fix)
1. Open **Storage pickers**, scroll to the **Save a file** section.
2. Using the keyboard, Tab to **Save a file** and press Enter.
3. In the Save dialog, type a name, Tab to **Save**, press Enter.
4. Press Tab / Shift+Tab -> focus is lost / trapped.

After the fix, focus returns to the **Save a file** button and Tab navigation works normally.

## Testing
- Verified manually: after saving, keyboard focus returns to the invoking button.
- Aligns with the repo's accessibility conventions (enforced by the Axe.Windows UI test scans).

[

https://github.com/user-attachments/assets/b054c543-1226-445a-bcc8-cfffb5309d5c

](url)

### DIFF
--- a/WinUIGallery/Samples/StoragePickers/StoragePickersFileThumbnail.txt
+++ b/WinUIGallery/Samples/StoragePickers/StoragePickersFileThumbnail.txt
@@ -28,6 +28,9 @@ private async void PickFileForThumbnailButton_Click(object sender, RoutedEventAr
             ThumbnailDetailsTextBlock.Text = "No file selected.";
             ThumbnailImage.Source = null;
             button.IsEnabled = true;
+
+            // Return keyboard focus to the button, which lost focus when it was disabled, to avoid a keyboard trap.
+            button.Focus(FocusState.Programmatic);
             return;
         }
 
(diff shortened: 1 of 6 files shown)


iamAbhi-916 (approved):
lgtm

PR #2221, opened 2026-09-03
Add a DP for ColorSelector for enabling alpha channel

HO-COOH:
## Description
I see the `ColorSelector` controls is currently only used for demo `AppWindowTitleBar` customization. And the alpha channel is ignored, [see  this doc](https://learn.microsoft.com/en-us/windows/apps/develop/title-bar?tabs=wasdk#colors). So it's better to disable the alpha channel to avoid confusion. And it can be enabled using the DP if the control is used for other places in the future.

## How Has This Been Tested?
Manual.

image

## Types of changes


marcelwgn (approved):
Good catch, thank you for fixing this in a flexible way for future usages of the control!

PR #2105, opened 2026-02-01
Add AppWindow Title Bar Customization sample

Zakariathr22:
## Description
Main changes are:
- Introduced `AppWindowTitleBarPage` showcasing customization examples.
- Updated `ControlInfo.json` accordingly.
- Implemented a lightweight `ColorSelector`control.
- Added new Sample Windows to demonstrate usage scenarios.
- Updated both `AppWindow` and `TitleBar` pages with the new enhancements.

## Motivation and Context
- Closes #2037 

## How Has This Been Tested?
Manually tested

## Screenshots (if appropriate):
image
image
image
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #2227, opened 2026-09-05
Fix the AppWindow example where PreferredHeightOption was misspelled as HeightOption

YexuanXiao:
The property was misspelled when this example was added in #2105; it should be [PreferredHeightOption](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.appwindowtitlebar?view=windows-app-sdk-2.0), not HeightOption.

This issue was originally reported at https://github.com/microsoft/microsoft-ui-xaml/issues/11779.


marcelwgn (approved):
Good catch, thank you for fixing this!

